PR #189, opened 2013-09-13
allow parameters in extension

docteurklein:
This is a **BC Break**.

You must change `phpspec.yml` accordingly: 

    # before
    extensions:
        - Some\Extension

    # after:
    extensions:
        Some\Extension: ~
        # or, and that's why
        Some\Other\Extension:
            option1: ~

I started to spec this until I realised this class is not speccable with current impl.

We should extract this extension loading mechanism to its own class.

### DIFF
--- a/features/configuration/developer_adds_extensions.feature
+++ b/features/configuration/developer_adds_extensions.feature
@@ -1,0 +1,129 @@
+Feature: Developer enables extensions
+  As a Developer
+  I want to enable and configure extensions
+  In order to customize or add features to phpspec behavior
+
+  Scenario: Adding non parametrized extensions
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
@docteurklein I don't think we'll be updating the PHP version soon

stof:
@ciaranmcnulty I suggest putting this on hold until we decide what we do with Testwork (if we go the way of using Testwork extensions, this proposal will be obsolete)

docteurklein:
@stof @ciaranmcnulty good idea.

ciaranmcnulty:
I'm going to close the PR for now, it doesn't mean it's not a good idea!

PR #215, opened 2013-11-15
Added generating of not found interfaces and implementing

karolsojko:
I was thinking that it would be cool if `PHPSpec` helped you with generating and implementing interfaces :)
### Generating interfaces

so for example if I had a spec like this: 

``` php
<?php

namespace spec\KingsLanding;

use PhpSpec\ObjectBehavior;

class KingSpec extends ObjectBehavior
{
    function it_should_be_a_royalty()
    {
        $this->shouldImplement('Royal\RoyaltyInterface');
    }
}
```

Let's say the interface `Royal\RoyaltyInterface` didn't exist because I haven't done it yet. So when we distinguish the `shouldImplement` matcher, `PHPSpec` could ask me if I want to create the `Royal\RoyaltyInterface` - and if I said yes it could generate the file for me.
### Implementing interfaces

Ok now when the interface is created we're not quite there yet. The interface is generated but the class still doesn't implement it. So I thought it could ask me if `KingsLanding\King` should implemented `Royal\RoyaltyInterface` and if I said yes it could add the implementation header to the class :)

I think this could automate nicely a few things at least. The whole process in a console is illustrated in this screenshot: http://d.pr/i/Uefo
### Result

So in the end I would end up with something like this:

``` php
<?php
//Royal/RoyalInterface.php
namespace Royal;

interface RoyaltyInterface
{
}

//KingsLanding/King.php
<?php

namespace KingsLanding;

class King implements \Royal\RoyaltyInterface
{
}

```

### DIFF
--- a/spec/PhpSpec/CodeGenerator/Generator/InterfaceGeneratorSpec.php
+++ b/spec/PhpSpec/CodeGenerator/Generator/InterfaceGeneratorSpec.php
@@ -1,0 +1,149 @@
+<?php
+
+namespace spec\PhpSpec\CodeGenerator\Generator;
+
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+class InterfaceGeneratorSpec extends ObjectBehavior
+{
+    /**
+     * @param PhpSpec\Console\IO $io
+     * @param PhpSpec\CodeGenerator\TemplateRenderer $tpl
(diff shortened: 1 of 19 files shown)


jakzal:
Related to the discussion of a special matcher for interfaces: http://blog.cleancoder.com/uncle-bob/2015/01/08/InterfaceConsideredHarmful.html

PR #335, opened 2014-03-22
Allow instantiation via a static factory method

damonjones:
It's not currently possible to use a factory method to instantiate a subject (useful when you have a private constructor).

This PR addresses that, by introducing the new `beConstructedThrough()` method, which requires a callable and optional constructor arguments.

Example class:

``` php
final class Example
{
    private $name;

    private function __construct($name)
    {
        $this->name = $name;
    }

    public static function createWithName($name)
    {
        return new self($name);
    }

    public function getName()
    {
        return $this->name;
    }
}
```

``` php
function it_should_be_constructed_through_a_factory_method()
{
    $this->beConstructedThrough('Example::createWithName', ['Damon']);
    $this->shouldHaveType('Example');
    $this->getName()->shouldEqual('Damon');
}
```

This implements the WIP feature/pull request that was originally made to the phpspec2 repository by @mathiasverraes: https://github.com/phpspec/phpspec2/pull/95

### DIFF
--- a/src/PhpSpec/Wrapper/Subject.php
+++ b/src/PhpSpec/Wrapper/Subject.php
@@ -90,10 +90,10 @@ class Subject implements ArrayAccess, WrapperInterface
     }
 
     /**
-     * @param callable $factoryMethod
-     * @param array    $arguments
+     * @param array|string $factoryMethod
+     * @param array        $arguments
      */
-    public function beConstructedThrough(callable $factoryMethod, array $arguments = array())
+    public function beConstructedThrough($factoryMethod, array $arguments = array())
     {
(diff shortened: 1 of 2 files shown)


bweston92:
How can I check an exception gets thrown in the constructor? Use case, value objects that have assertions.

mdavis1982:
@bweston92 You can do:

```php
$this
    ->shouldThrow(YourException::class)
    ->during('__construct', ['list', 'of', 'parameters']);
```

bweston92:
But the constructor is private with instantiation via static method?

ciaranmcnulty:
With named constructors there currently isn't a good way to do this. See #699

PR #412, opened 2014-08-04
Trigger an error when expected spec class is not found in file

### LINKED ISSUE #354: PhpSpec should warn when a spec class does not match its expected name
In case there is a mismatch between the spec class name and the expected name based on the path, PhpSpec ignores the file silently. This makes it really hard to spot such mistakes.

Reproducing it: create a file `spec/Acme/Demo/GreeterSpec.php`, and put the class `spec\Acme\Hello\GreeterSpec` in it (notice the different namespace). Running PhpSpec will pass, but your GreeterSpec will not run and you won't know it.

This happened to me (and it is broken since a few weeks), and I detected it only because I ran the PHP-CS-Fixer on the spec folder and it warned me about a PSR-0 mismatch.

tomphp:
fixes #354

First phpspec PR, please be gentle!

I have often come across this issue when renaming/moving a spec file and forgetting to update the classname/namespace inside.

I'm not sure throwing an exception is the correct thing to do here, if there is a more appropriate way to handle this a nudge in the right direction would be muchly appreciated.

### DIFF
--- a/features/bootstrap/PhpSpecContext.php
+++ b/features/bootstrap/PhpSpecContext.php
@@ -112,18 +112,21 @@ class PhpSpecContext implements Context
     }
 
     /**
-     * @Given /^(?:|the )(?:spec |class )file "(?P<file>[^"]+)" contains:$/
+     * @Given /^(?:|the )class file "(?P<file>[^"]+)" contains:$/
      */
     public function theFileContains($file, PyStringNode $string)
     {
-        $dirname = dirname($file);
-        if (!file_exists($dirname)) {
-            mkdir($dirname, 0777, true);
-        }
-
-        file_put_contents($file, $string->getRaw());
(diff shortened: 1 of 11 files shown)


respinoza:
@tomphp I wonder if you are still around. We have had this problem, see #354 
At this point I don't know if you are going to implement this or not but I would like to resurrect this fix to avoid the dangerous situation where the CI could execute phpspec and detect no problems but in reality no specs were run.

tomphp:
@respinoza I would still like to see this happen but the implementation was a bit tricky without some BC breaks. I've forgotten where we got to with it so I'll have to read back and try and work it out - plus I'll have a look at the current code and see if it's changed. Maybe if it isn't possible without a sensible, non-BC breaking, solution, we can try for in to be added in a major release?

PR #541, opened 2014-10-23
Allow StringPresenter::presentCallback() to work with static classes

alreece45:
When given a callable array, StringFormatter assumes that the first part of the array is an object, even though it may also be a string. 

The StringFormatter is updated to check to see if the first parameter is an object. To allow distinguishing between static and non-static callables, the presentValue() method uses the existing presentValue() method to present the object.

### DIFF
--- a/spec/PhpSpec/Formatter/Presenter/StringPresenterSpec.php
+++ b/spec/PhpSpec/Formatter/Presenter/StringPresenterSpec.php
@@ -67,6 +67,14 @@ class StringPresenterSpec extends ObjectBehavior
             ->shouldReturn('[date()]');
     }
 
+
+    function it_presents_invokable_object_as_string(WithMagicInvoke $object)
+    {
+        $className = get_class($object->getWrappedObject());
+        $this->presentValue($object)
+            ->shouldReturn(sprintf('[obj:%s]', $className));
+    }
+
     function it_presents_string_as_string()
     {
         $this->presentString('some string')->shouldReturn('some string');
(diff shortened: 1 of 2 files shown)


src/PhpSpec/Formatter/Presenter/StringPresenter.php
```diff
@@ -345,7 +345,8 @@ private function shouldSkipTracePresentation(array $call)
     private function presentCallable($value)
     {
         if (is_array($value)) {
```
stof: @ciaranmcnulty presentCallable is called only with a valid callable. Callable arrays have always a length of 2

spec/PhpSpec/Formatter/Presenter/StringPresenterSpec.php
```diff
@@ -61,6 +61,47 @@ function it_presents_exception_as_class_with_constructor()
...
+    }
+
+    function it_presents_magic_method_as_string(WithMagicCall $object)
+    {
+        $className = get_class($object->getWrappedObject());
+        $this->presentValue(array($object, 'undefinedMethod'))
+            ->shouldReturn(sprintf('[obj:%s]::undefinedMethod()', $className));
+    }
+
+    function it_presents_static_method_as_string(WithMethod $object)
```
stof: @alreece45 can you fix this ?

alreece45: @stof Its been updated to not need the collaborator

stof: you updated a different example, not this one

PR #575, opened 2014-12-02
[WIP] Introduce symfony service container and testwork extensions

richardmiller:
Starts to replace the custom container with the Symfony one. Its wrapped inside it at the moment with a view to just having the symfony container in a later major version. As a result of trying to maintain b/c its quite hacky at the moment and would require quite a bit of clean up at a later stage.

Supports both current phpspec extensions and testwork extensions. Started moving some of the core services to testwork extensions.


src/PhpSpec/ServiceContainer.php
```diff
@@ -21,26 +26,32 @@
...
 
+    public function __construct($container)
+    {
+        $container = new ContainerBuilder();
+        $container->setParameter('paths.base', __DIR__);
+
+        $this->containerBuilder = $container;
+        $definition = new Definition('PhpSpec\\ServiceFactory');
+        $this->containerBuilder->setDefinition(self::COMPAT_SERVICE_FACTORY_ID, $definition);
+        $this->serviceFactory = $this->containerBuilder->get(self::COMPAT_SERVICE_FACTORY_ID);
```
ciaranmcnulty: For me the question is whether it causes any issues for existing extensions, and whether it would encourage any bad practice going forwards in future when people write extensions only using the new interface.

ciaranmcnulty:
@richardmiller I'd like to see this come in, if you can come back to it?

To split out TestWork I suggest we fork the subtree into `phpspec/testwork` then PR back anything we need to change to make it work better. @everzet and @stof what are your feelings about that?

The test for this change is:

* Does it make life easier for future Extension writers?
* Are we sure existing extensions won't break?
* Is supporting both containers maintainable?
* Does it bring us closer to a shared platform (this is not a primary concern but it's a valid one)

stof:
@ciaranmcnulty please don't create your own subtree for Testwork. this is something we will do in Behat itself as soon as we are ready (I think it could happen after https://github.com/Behat/Behat/pull/656 is completed).
Introducing 2 different (conflicting) source code for Behat Testwork is a bad idea, as people might start depending on ``phpspec/testwork`` even though it is not the canonical code (and we would have to mark ``behat/behat`` as replacing ``phpspec/testwork`` to avoid a mess in composer, which seems bad.

Supporting both containers would be sustainable if we really build the PhpSpec ServiceContainer as a simple BC layer on top of the DI component, keeping everything in the DI component. This BC layer would be there only for extensions written against the current API. (note that the current code is the wrong way to implement ServiceContainer on top of the DI component)
Using Testwork extensions has a benefit that it makes extensions much powerful (possibility to configure them with a config being validated) and cleaner (in case the extensions care about being robust). So I think it is a good move. It would also help moving toward using Testwork as the PhpSpec basis (which will become easier after the tester refactoring is merged in Testwork).

stof:
However, to keep this refactoring easier to achieve, I think we need to split it in several steps: introducing the DI component inside the ServiceContainer first, and then refactoring the extension system to allow using Testwork extensions as a second step.

ciaranmcnulty:
Splitting it into two steps might be clearer

ciaranmcnulty:
@richardmiller as this has desynced so much (and is WIP) I'm going to close for now, we can hopefully collaborate on making it work still

PR #579, opened 2014-12-03
Generate collaborators when they don't yet exist

ciaranmcnulty:
collaborators

* Offers to generate an Interface when a specified collaborator does not exist
* Generates methods in collaborators as long as they are interfaces


ciaranmcnulty:
Is there anything more I should do to this?

ciaranmcnulty:
@stof re your last comment, I might take a look at doing method generation for interfaces only. I'm 99% sure it's a good idea for interfaces and on the fence for class collaborators.

clarkeash:
+1

peterjmit:
@ciaranmcnulty I thought PhpSpec used to do this when you used annotation based mocks (rather than type hinting for them).

Either way :+1:

ciaranmcnulty:
Having used this a bit, it's very easy to miss a `use` and get prompted to generate `spec\Foo`. I'll look at adding some filtering against the current `spec_prefix`es to trim this.

ciaranmcnulty:
@MarcelloDuarte suggested when a user subsequently tries to `describe` an Interface we offer to convert it to a class for them.

gnugat:
+1 for this feature. What about the spec for those collaborators? Should it also generate a spec skeleton?

ciaranmcnulty:
@gnugat we're only offering to generate interfaces, to avoid that complexity. The idea is the author can stay focused on the object they're speccing. 

Then when they focus on the collaborators they can either convert the interface to a class (we could automate it), or they simply describe a new class that implements the interface.

I like the idea that they're guided towards that 'should this stay an interface, or should it be a class?' question.

ciaranmcnulty:
This is now 'feature complete' as far as this PR goes.

I had to make some Prophecy changes to enable the functionality, so it can't be merged until we see if they get accepted + prophecy does a stable release.

This change has highlighted the problems with code generators, so we'll need to do a refactor in there.

gnugat:
What kind of problems? I'm interested, because I've always felt that it had several limitations (e.g. templates not flexible enough, usage of regex+replace to insert new code).

ciaranmcnulty:
@gnugat see #619

ciaranmcnulty:
I'd appreciate some review on this as I don't anticipate changing much. 

Wondering how to handle the Prophecy version issue - I could detect presence of the relevant APIs and bail out, but the testing environment would have to switch between versions.

Far simpler would be to bump the Prophecy version requirement. In the past I'd not have hesitated by Prophecy is now a 'shared' lib.

Any opinions @stof ? @everzet ?

ubick:
:+1:

composer.json
```diff
@@ -19,7 +19,7 @@
 
     "require": {
         "php":                      ">=5.3.3",
-        "phpspec/prophecy":         "~1.1",
+        "phpspec/prophecy":         "dev-master",
```
MarcelloDuarte: @everzet we need a proper release tag to depend on. We are waiting for your next release.

features/bootstrap/Fake/Prompter.php
```diff
@@ -17,11 +18,17 @@ public function setAnswer($answer)
     public function askConfirmation($question, $default = true)
     {
         $this->hasBeenAsked = true;
+        $this->question = $question;
         return (bool)$this->answer;
     }
 
-    public function hasBeenAsked()
+    public function hasBeenAsked($question=null)
```
MarcelloDuarte: missing space before and after operator (=)

features/bootstrap/IsolatedProcessContext.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    private $lastOutput;
+
+    /**
+     * @beforeSuite
+     */
+    public static function checkDependencies()
+    {
+        chdir(sys_get_temp_dir());
+        if (!@`which expect`) {
```
MarcelloDuarte: this isn't going to work under windows...

MarcelloDuarte: I guess that's why you have the error suppression operator...

ciaranmcnulty: This method is specifically checking whether expect is present on the system and showing a sensible message if it's not.

IsolatedProcessContext is only used in the 'smoke' suite (not in the `default` suite) so someone checking the project out on Windows and running `bin/behat` won't get any issues

rtuin:
Great feature, and +1 for doing this for interfaces only! :+1:

rtuin:
The method generation for collaborating interfaces only, that is.

stof:
@ciaranmcnulty I will make a Prophecy release this weekend. Then, you will be able to bump the requirement

features/bootstrap/IsolatedProcessContext.php
stof: @ciaranmcnulty where is it used ? I don't see any change to the behat config in this PR to introduce the registration of this context class

ciaranmcnulty: I am not sure what's going on with the diff here - This class and the DotFormatterSpec both look the same as `master` - any ideas?

stof: try rebasing your feature branch on top of master to fix this diff

ciaranmcnulty:
@stof :+1: thanks

ciaranmcnulty:
Rebasing seems to have fixed the bad diff

src/PhpSpec/CodeGenerator/Generator/templates/method_signature.template
```diff
@@ -0,0 +1,3 @@
+
+    public function %name%(%arguments%);
```
stof: this template generates a method signature for an interface only, not for a class. So it may need a better name.

src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @param ExampleEvent $event
+     */
+    public function afterExample(ExampleEvent $event)
+    {
+        if (!$this->io->isCodeGenerationEnabled()
+        || !($exception = $event->getException())
+        || !$exception instanceof MethodNotFoundException) {
```
stof: wrong indentation

src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * @param mixed $prophecyArguments
+     * @return array
+     */
+    private function getRealArguments($prophecyArguments)
+    {
+        if ($prophecyArguments instanceof ArgumentsWildcard)
+        {
```
stof: should be on the previous line

src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php
```diff
@@ -0,0 +1,158 @@
...
+     * @param SuiteEvent $event
+     */
+    public function afterSuite(SuiteEvent $event)
+    {
+        foreach ($this->interfaces as $interface => $methods) {
+
+            try {
+                $resource = $this->resources->createResource($interface);
+            }
+            catch (\RuntimeException $e) {
```
stof: should be on the previous line. And shouldn't we use some named exception here rather than catching all RuntimeException ?

stof:
Does it work for collaborators defining their type through the phpdoc rather than through a typehint ? I doubt it

ciaranmcnulty:
I'll add a feature about type hints in params, I hadn't considered that case.

src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php
ciaranmcnulty: Runtime exception is what resource generator throws. Perhaps that should indeed be more specific

src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php
ciaranmcnulty: According to which standard? What would correct look like?

ciaranmcnulty:
@stof Docblock params seem to just work

ciaranmcnulty:
@stof I addressed your comments, do you have any plans for a prophecy tag (beta is ok, I'll probably do a 2.2.0-beta before a real release)

ciaranmcnulty:
@stof @everzet Can either of you make a Prophecy tag?

PR #597, opened 2015-01-04
Fixing division by zero error when no specs are found

ciaranmcnulty:
Fixes 596

### DIFF
--- a/src/PhpSpec/Formatter/ProgressFormatter.php
+++ b/src/PhpSpec/Formatter/ProgressFormatter.php
@@ -87,7 +87,7 @@ class ProgressFormatter extends ConsoleFormatter
     private function getBarLengths($percents)
     {
         $stats = $this->getStatisticsCollector();
-        $specProgress = $stats->getTotalSpecs() ? ($stats->getTotalSpecs()+1)/$stats->getTotalSpecsCount() : 0;
+        $specProgress = $stats->getTotalSpecsCount() ? ($stats->getTotalSpecs() + 1) / $stats->getTotalSpecsCount() : 0;
 
         $barLengths = array_map(
             function ($percent) use ($specProgress) {
--- a/src/PhpSpec/Formatter/ProgressFormatter.php
+++ b/src/PhpSpec/Formatter/ProgressFormatter.php
@@ -16,6 +16,7 @@ namespace PhpSpec\Formatter;
 use PhpSpec\Console\IO;
 use PhpSpec\Event\SuiteEvent;
 use PhpSpec\Event\ExampleEvent;
+use PhpSpec\Listener\StatisticsCollector;
 
 class ProgressFormatter extends ConsoleFormatter
 {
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
I would suggest we release a 2.1.1 with this included

cakper:
:+1: looks good

src/PhpSpec/Formatter/ProgressFormatter.php
```diff
@@ -87,7 +87,7 @@ function ($count) use ($total) {
     private function getBarLengths($percents)
     {
         $stats = $this->getStatisticsCollector();
-        $specProgress = ($stats->getTotalSpecs()+1)/$stats->getTotalSpecsCount();
+        $specProgress = $stats->getTotalSpecs() ? ($stats->getTotalSpecs()+1)/$stats->getTotalSpecsCount() : 0;
```
jakzal: Technically, we're dividing by `totalSpecsCount` here, not `totalSpecs`. I'd rather check if `getTotalSpecsCount()` returns something that's not a zero. If there was an error in the statistics collector, this could still fail. By using `totalSpecs` instead of `totalSpecsCount`, `ProgressFormatter` makes assumptions on how statistics collector works.

Also, formatting is not consistent with the rest of the code. More spaces please ;)

```php
$specProgress = $stats->getTotalSpecsCount() ? ($stats->getTotalSpecs() + 1) / $stats->getTotalSpecsCount() : 0;
```

ciaranmcnulty: That's a good catch! Maybe these two methods have bad names?

ciaranmcnulty:
I'll write a spec or feature that covers this case before merge.

src/PhpSpec/Formatter/ProgressFormatter.php
stof: @ciaranmcnulty the division by 0 happens when ``getTotalSpecsCount`` returns 0. So the safeguard should check this method too. The name of the method is not involved here. Your logic is

ciaranmcnulty: I made the mistake because of the similarities of the names

everzet: @ciaranmcnulty make sure it doesn't happen again

src/PhpSpec/Formatter/ProgressFormatter.php
```diff
@@ -87,7 +87,7 @@ function ($count) use ($total) {
     private function getBarLengths($percents)
     {
         $stats = $this->getStatisticsCollector();
-        $specProgress = ($stats->getTotalSpecs()+1)/$stats->getTotalSpecsCount();
+        $specProgress = $stats->getTotalSpecsCount() ? ($stats->getTotalSpecs() + 1) / $stats->getTotalSpecsCount() : 0;
```
everzet: I'd extract a method for clarity here :)

jakzal: :+1: :)

ciaranmcnulty: http://en.wikipedia.org/wiki/Parkinson's_law_of_triviality

If nothing else, my error has brought @everzet back into the project ;-P

PR #598, opened 2015-01-07
Fix composer.json

nexik:
Currently composer return following error:

Could not load package phpspec/phpspec in http://packagist.org: [UnexpectedValueException] Could not parse version constraint ^1.0.1: Invalid version string "^1.0.1"

This change fix this problem


ciaranmcnulty:
Can you update your composer and try installing again?

ciaranmcnulty:
The `^` syntax is a recent addition.

nexik:
Yes this fix the problem, so I'll close this PR

PR #603, opened 2015-01-09
Show which spec class is not the correct base spec namespace

adamelso:
I received this exception during execution

```
  [RuntimeException]                                                                    
  Spec class must be in the base spec namespace `spec\Acme\DemoBundle`. 
```

However, I had absolutely no idea which spec caused the exception, even with verbose output, until passed in the classname to the exception.

Now, exception message should read for example:

```
  [RuntimeException]                                                                    
  Spec class `spec\Acme\Demo\Controller\ControllerSpec` must be in the base spec namespace `spec\Acme\DemoBundle`. 
```


ciaranmcnulty:
This looks great, thanks

Issue #580, opened 2014-12-09
Remove deprecated function calls


WouterJ:
Using Symfony 2.7@dev components, these deprecated errors are triggered when simply running `phpspec desc`:

* Deprecated: "Symfony\Component\Console\Helper\DialogHelper" is deprecated since version 2.5 and will be removed in 3.0. Use "Symfony\Component\Console
\Helper\QuestionHelper" instead.
* Deprecated: The ability to pass file names to Yaml::parse() was deprecated in 2.7 and will be removed in 3.0. Please, pass the contents of the file in
stead.

You should lower the error reporting level or increase dependencies to newer 2.5.*.\

ciaranmcnulty:
Looking inside Console\IO, there are three methods that use `DialogHelper`: `ask`. `askAndValidate`, and `askQuestion`.  Of these only the last is actually used inside the codebase, and has a decent API. The first two are unused and are tightly coupled to the `DialogHelper` API.

It makes sense to me to deprecate the first two unused methods and not implement a `QuestionHelper`version of them.

ciaranmcnulty:
The deprecated calls are now removed, and there is a travis build to warn us if any more turn up in unstable symfony versions.

PR #604, opened 2015-01-11
Support QuestionHelper in newer symfony versions

ciaranmcnulty:
Removes problems with `DialogHelper` deprecation (see #580)

Direct access to `DialogHelper` is now deprecated and a new `Prompter`interface is introduced, which will be using either `DialogHelper` or `QuestionHelper` as appropriate based on availability.


src/PhpSpec/Console/Application.php
```diff
@@ -57,9 +58,17 @@ public function getContainer()
      */
     public function doRun(InputInterface $input, OutputInterface $output)
     {
+        $helperSet = $this->getHelperSet();
         $this->container->set('console.input', $input);
         $this->container->set('console.output', $output);
-        $this->container->set('console.helper.dialog', $this->getHelperSet()->get('dialog'));
+        $this->container->set('console.helper.dialog', $helperSet->get('dialog'));
```
stof: Getting the dialog helper here will still trigger the deprecation warning, so it will not solve the issue. You should set the service in a lazy way, so that the retrieval is only done if an extension actually gets it (in which case the deprecation warning will be legitimate):

```php
$this->container->set('console.helper.dialog', function () use ($helperSet) {
    return $helperSet->get('dialog');
});
```

src/PhpSpec/Console/Application.php
```diff
@@ -57,9 +58,17 @@ public function getContainer()
...
+        $this->container->set('console.prompter.factory', function($c) use ($helperSet) {
```
stof: shouldn't it be shared ?

src/PhpSpec/Console/IO.php
```diff
@@ -285,12 +298,14 @@ public function askConfirmation($question, $default = true)
...
-        return $this->dialogHelper->askConfirmation(
-            $this->output, implode("\n", $lines). "\n", $default
-        );
+        $formattedQuestion = implode("\n", $lines) . "\n";
+
+        return $this->prompter->askConfirmation($formattedQuestion, $default);
     }
 
     /**
+     * @deprecated
```
stof: is this method used somewhere ?

src/PhpSpec/Console/Application.php
ciaranmcnulty: Currently the Application will always get the dialogHelper. I had not realised that it was deprecated that strongly. Would it be too much of a BC break to remove it from the constructor of the Application? Otherwise I could look at injecting a Proxy of some sort.

src/PhpSpec/Console/IO.php
ciaranmcnulty: Not in the codebase, but potentially in an Extension.

src/PhpSpec/Console/Application.php
ciaranmcnulty: :+1:

src/PhpSpec/Console/Application.php
ciaranmcnulty: I can't see a way of doing this without a BC break to be honest, it's always going to be instantiated when IO is, and in absence of a clean extension point it's not clear where we can make the change.

If I get time I'll do a survey of extensions and see if any are actually using this directly.

stof: @ciaranmcnulty I suggest adopting the same policy than Behat: considering that people are not meant to instantiate the PhpSpec classes themselves (this is done in the DI container) at least for these levels of the codebase. This means we can change the constructor arguments as we want for refactorings (and actually, you are already relying on this as you added a mandatory constructor argument in the IO).
This means you could move ``askAndValidate`` to the prompter as well

ciaranmcnulty: I was thinking of the case where someone extends or decorates the DialogHelper and tries to inject their modified version via the container. Maybe we don't care about that.

stof: but you already broke BC for this case, as he needs to provide the Prompter to the parent.

For composition, this could be handled by providing a nice way to handle extending a service by composition in the container instead. Both Pimple (since 1.0) and Symfony DIC (since 2.5) have a nice way for this (the Symfony DIC allows to do it since 2.0, but in a less simple way). This would avoid the decorator to know anything about the instantiation of the inner service. This is an issue with the container only if they have to redo all the work.

for inheritance, this would indeed forbid to consider the constructor as an internal API. We don't face this issue in Behat though, because most our classes are final.

ciaranmcnulty: OK. In that case I'm going to break BC and remove dialogHelper here.

It annoys me supporting `ask` and `askAndValidate` - part of me thinks they should be removed and Extension writers can catch up, but that might be too big a step until the versioning policy is more explicitly stated.

ciaranmcnulty:
Todo:

- [ ] Move Integration tests out of phpspec into something else (probably PHPUnit)
- [ ] Remove dependency from `IO` to `dialogHelper`
- [ ] Make a decision on BC of `ask` and `askAndValidate`

stof:
+1 for using PHPUnit for integration spec

regarding ask and askAndValidate, are they used anywhere in PhpSpec itself ? If no, are they used in known extensions ?

ciaranmcnulty:
They're used nowhere in PhpSpec, I'll have to do research re: extensions.

So it's a BC break, but might be a BC break that impacts very few people.

ciaranmcnulty:
@stof According to this script https://gist.github.com/ciaranmcnulty/7706c36bbc62ee858317 there are no uses in the extensions we link to.

stof:
@ciaranmcnulty then I guess it is fine with removing these methods (we just need to make it clear in the changelog for people which might have written their own extensions). Can you finish this PR ?

ciaranmcnulty:
I've got the PHPUnit tests in place, just need to do the refactoring.

I'm leaning back towards leaving those methods in, seeing as it's not hard to implement them. Should be able to finish off today/tomorrow.

ciaranmcnulty:
If I take the methods out, I'll add a section to CHANGES.md for any internal BC breaks.

ciaranmcnulty:
@stof removing the dependency was pretty straightforward.

There is still a possibility to support the `ask` and `askandvalidate` methods but it would require cluttering the `Prompter` interface. It feels bad to make the interface messy, but also bad to break the BC.

integration/PhpSpec/Console/Prompter/DialogTest.php
```diff
@@ -0,0 +1,59 @@
...
+     * @var \Symfony\Component\Console\Helper\DialogHelper
+     */
+    private $dialogHelper;
+
+    /**
+     * @var \PhpSpec\Console\Prompter
+     */
+    private $prompter;
+
+    function setUp()
```
stof: this should be a protected method

integration/PhpSpec/Console/Prompter/DialogTest.php
```diff
@@ -0,0 +1,59 @@
...
+    {
+        $this->output = $this->getMock('Symfony\Component\Console\Output\OutputInterface');
+        $this->dialogHelper = $this->getMockBuilder('Symfony\Component\Console\Helper\DialogHelper')
+                                   ->disableOriginalConstructor()->getMock();
```
stof: I suggest using Prophecy doubles in the PHPUnit tests rather than the ugly PHPUnit mock objects (they will be in core in PHPUnit 4.5, and there is the https://github.com/phpspec/prophecy-phpunit package for older PHPUnit versions)

integration/PhpSpec/Console/Prompter/DialogTest.php
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * @test
+     */
+    function it_can_ask_a_question_and_return_the_result()
+    {
+        $this->dialogHelper->expects($this->once())
+                           ->method('askConfirmation')
+                           ->with($this->equalTo($this->output), 'Are you sure?', true)
```
stof: this should use ``$this->identicalTo()`` instead, to do strict comparison for the object

integration/PhpSpec/Console/Prompter/FactoryTest.php
```diff
@@ -0,0 +1,108 @@
...
+     * @var \Symfony\Component\Console\Output\OutputInterface
+     */
+    private $output;
+
+    /**
+     * @var int
+     */
+    private $reportingLevel;
+
+    function setUp()
```
stof: should be protected

src/PhpSpec/Console/Application.php
```diff
@@ -57,9 +58,21 @@ public function getContainer()
...
-        $this->container->set('console.helper.dialog', $this->getHelperSet()->get('dialog'));
+        $this->container->setShared('console.prompter.factory', function($c) use ($helperSet) {
+            return new Factory(
+                $c->get('console.input'),
+                $c->get('console.output'),
+                $helperSet
+            );
+        });
+        if (!$this->container->isDefined('console.prompter')) {
+            $this->container->setShared('console.prompter', function($c) {
```
stof: this should be done in the ContainerAssembler IMO

integration/PhpSpec/Console/Prompter/DialogTest.php
ciaranmcnulty: I didn't want to use the extension when 4.5 is just around the corner. I'll convert to prophecy when 4.5 is stable.

integration/PhpSpec/Console/Prompter/DialogTest.php
ciaranmcnulty: Good point

src/PhpSpec/Console/Application.php
ciaranmcnulty: The Factory or the Prompter?

stof: Well, the factory currently relies on the helperset. Defining this one in the ContainerAssembler requires putting the helperset as a service on its own (which might be cleaner).

But the prompter itself should be defined in the assembler for sure.

ciaranmcnulty: I'll move the prompter, it's not a symfony concept. I think factory definition should stay here for now.

ciaranmcnulty:
@stof I think I addressed all your points, did I miss any?

stof:
:+1:

ciaranmcnulty:
Cool, thank you!

PR #606, opened 2015-01-12
Added Scenario skipping extension link

gquemener:


docs/cookbook/extensions.rst
```diff
@@ -69,3 +69,4 @@ Miscellaneous
  * `Data provider <https://github.com/coduo/phpspec-data-provider-extension>`_
  * `Matcher extension <https://github.com/coduo/phpspec-matcher-extension>`_
  * `Behat Integration <https://github.com/richardmiller/BehatSpec>`_
+ * `Scenario skipping through annotation <https://github.com/akeneo/PhpSpecSkipExampleExtension>`_
```
stof: the extension is skipping examples, not scenarios. There is no concept of scenario in PhpSpec (this is a Gherkin concept, which is then used in Behat)

Issue #608, opened 2015-01-15
Smarter example name matching and splitting to end the wars


Dragonrun1:
As many of you know there are three main coding naming conventions:

snake_case
camelCase
PascalCase

and everyone has an opinion on which is best which has been know to turn into flame wars on forums, around water coolers, etc when brought up. I'm not here to say that phpspec was right or wrong in going with snake_case for examples in specs but to offer another way of looking at it and possible end any battles at least as far as phpspec is concerned. So why not allow the developers using phpspec to decide which they prefer in their own project? This can be done without adding any settings or anything but simply by making phpspec a little bit smarter when it is loading the examples. Let's do a little digging into how phpspec currently finds, loads, and enforce the naming convention in the spec examples before getting on to my _better_ way.

__Code BackGround__
Phpspec uses ```ResourceLoader::load()``` to find, load, split up the spec files into Specifications and Examples. On line 69 is where it determines of it thinks a method is an example as should here:
```php
if (!preg_match('/^(it|its)[^a-zA-Z]/', $method->getName())) {
```
Just so everyone understands it the RegEx will match any string starting with 'its' or 'it' followed by a non-alphabetical character. So for some examples to show things it would _not_ match:

```
itsShould...
itShould...
ItsShould...
IDoNotMatch...
```

and ones that do:

```
its_should_...
it_should_...
its_Should...
its0001...
```

Some of you probably are surprised by the last couple of allowed names but they would be allowed given the current RegEx that phpspec uses. I think this was not intended and they simply were trying to insure that camel case or pascal case names didn't match. For those of you wondering the following RegEx is better at allowing only what was intended ``'/^(it_|its_)/'``. There is only one other place where the use of snake case vs something else comes into play in phpspec and it is in the same ``load()`` method on line 78 as seen here:
```php
$example = new Node\ExampleNode(str_replace('_', ' ', $method->getName()), $method);
```
It uses a simple ``str_replace()`` to make the ``_`` into a space instead. This is done to make it into a more human readable form used by the output formatters. These couple of lines are the only thing that causes phpspec to currently require the use of underscore case vs anything else. So now that we understand some background on how little phpspec really cares one way of the other it let me show you what I think is a _better_ way.
 
__A Better Way__

[...]

PR #609, opened 2015-01-15
Better way pull for issue #608

Dragonrun1:
Here the code to let phpspec allow the user to decide which coding style they want to use for spec examples.


ciaranmcnulty:
See discussion on #608

PR #611, opened 2015-01-16
Implement a Test Anything Protocol (TAP) formatter

padraic:
Who doesn't like nice machine friendly output that isn't XML? ;)

This is a basic implementation of TAP 13: https://testanything.org/tap-version-13-specification.html Intention is to ease integration with Humbug which is a TAP consumer and save me from the horror that is JUnit.


spec/PhpSpec/Formatter/TapFormatterSpec.php
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    function it_outputs_version_on_beforesuite_event(SuiteEvent $event, IO $io)
+    {
+        $this->beforeSuite($event);
+        $expected = 'TAP version 13';
+        $io->writeln($expected)->shouldHaveBeenCalled();
+    }
+
+    function it_outputs_plan_on_aftersuite_event(SuiteEvent $s_event, ExampleEvent $e_event, ExampleNode $example, IO $io, StatisticsCollector $stats)
```
ciaranmcnulty: In most places we use camelCase for collaborators, it might be better for consistency if you do that here.

ciaranmcnulty:
Form a first look, this looks like a good addition. It's not a format I've used but I'm sure some people would find it useful.

Is it possible to add some Behat coverage?

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * @param ExampleEvent $event
+     */
+    public function afterExample(ExampleEvent $event)
+    {
+        $this->examplesCount++;
+        $desc = sprintf(
+            self::DESC,
+            $this->currentSpecificationTitle,
+            preg_replace('/^it /', '', $event->getExample()->getTitle())
```
ciaranmcnulty: Example titles can also start with `its`

spec/PhpSpec/Formatter/TapFormatterSpec.php
```diff
@@ -0,0 +1,116 @@
...
+    {
+        $stats->getEventsCount()->willReturn(3);
+        $e_event->getExample()->willReturn($example);
+        $example->getTitle()->willReturn('foobar');
```
ciaranmcnulty: It doesn't look like this stub is really used in this specific example, and it's common to each example, you could move some of this stubbing into `let()` if it won't impact understandability.

ciaranmcnulty:
You could also add a note to the docs: https://github.com/phpspec/phpspec/blob/master/docs/cookbook/configuration.rst#formatter

ciaranmcnulty:
Similarly the formatters are listed in the run command's usage information: https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Console/Command/RunCommand.php#L75-L81

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,124 @@
...
+            self::PLAN,
+            $this->getStatisticsCollector()->getEventsCount()
+        ));
+    }
+
+    /**
+     * @param ExampleEvent $event
+     * @return string
+     */
+    protected function getExceptionMessage(ExampleEvent $event, $depth = null)
```
stof: I suggest making this private. and ``$depth`` is unused

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,124 @@
...
+    {
+        if (null === $exception = $event->getException()) {
```
ciaranmcnulty: Currently Pending and Skipped always have an exception, so I'm not sure this guard is needed

spec/PhpSpec/Formatter/TapFormatterSpec.php
```diff
@@ -0,0 +1,116 @@
...
+        $e_event->getResult()->willReturn(0);
+
+        $this->afterExample($e_event);
+        $this->afterSuite($s_event);
+
+        $expected = '1..3';
```
stof: you don't really need to use a variable here IMO

spec/PhpSpec/Formatter/TapFormatterSpec.php
padraic: It's there to prime the pump, i.e. the 1 in 1..3. Would print zero otherwise which should be invalid TAP ergo technically correct and technically me being pedantic ;).

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * @param ExampleEvent $event
+     */
+    public function afterExample(ExampleEvent $event)
+    {
+        $this->examplesCount++;
+        $desc = sprintf(
+            self::DESC,
+            $this->currentSpecificationTitle,
+            preg_replace('/^it[s]* /', '', $event->getExample()->getTitle())
```
ciaranmcnulty: `itssssss` shouldn't match so you might be better with `/^its? /`

spec/PhpSpec/Formatter/TapFormatterSpec.php
```diff
@@ -0,0 +1,116 @@
...
+        $this->beforeSuite($event);
+        $expected = 'TAP version 13';
+        $io->writeln($expected)->shouldHaveBeenCalled();
+    }
+
+    function it_outputs_plan_on_aftersuite_event(SuiteEvent $suiteEvent, ExampleEvent $exampleEvent, ExampleNode $example, IO $io, StatisticsCollector $stats)
+    {
+        $stats->getEventsCount()->willReturn(3);
+        $exampleEvent->getExample()->willReturn($example);
+        $example->getTitle()->willReturn('foobar');
```
ciaranmcnulty: What I meant here was that `getTitle` is only used in string concat so doesn't affect anything you're doing in this example.

These lines seem common to all examples, so might be clearer in `let()`:

```php
$specEvent->getSpecification()->willReturn($spec);
$exampleEvent->getExample()->willReturn($example);
$example->getTitle()->willReturn('it foobar');
```

src/PhpSpec/Formatter/TapFormatter.php
padraic: @ciaranmcnulty Done!

padraic:
A question, should I count pending results as ok or not ok?

ciaranmcnulty:
@padraic We treat `Pending` results as failures when we work out whether the suite passed (see https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Console/ResultConverter.php#L30-L39) so if there's no 'pending' analogue in TAP you should probably treat it as failure.

padraic:
@ciaranmcnulty I was thinking that. I'll switch it to a "not ok" result - consumers can check the directive or severity to differentiate from other "not ok" results which are fails/errors. Only really thought of it because PHPUnit basically passes through incompletes which never seemed right but I'd gotten used to it!

ciaranmcnulty:
@padraic Yes, Pending is different from Incomplete I think... Pending means 'I haven't done this yet and I need to do this for the suite to be right', whereas I've never quite been sure what Incomplete means for PHPUnit. 'I haven't made it work but have committed it anyway' perhaps? ;-)

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,87 @@
...
+
+      use PhpSpec\Exception\Example\SkippingException;
+      use PhpSpec\Exception\Example\ErrorException;
+
+      class Tap
+      {
+          public function fire($stuff)
+          {
+              switch ($stuff) {
+                case 'pass':
```
ciaranmcnulty: Some indentation here would help legibility

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -29,6 +29,7 @@
 use PhpSpec\Config\OptionsConfig;
 use RuntimeException;
 use Symfony\Component\Process\PhpExecutableFinder;
+use Symfony\Component\Yaml\Dumper as YamlDumper;
```
ciaranmcnulty: Is this used? It's not showing in the diff

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,136 @@
...
+    private function getExceptionMessage(ExampleEvent $event, $yaml = false)
+    {
+        if (false === $yaml) {
+            return str_replace(
+                array("\r\n", "\n", "\r"),
+                ' \\ ',
+                $event->getException()->getMessage()
+            );
+        }
+        $message = $event->getException()->getMessage();
```
ciaranmcnulty: You could move this earlier and use the var on line 125

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,87 @@
+Feature: Use the TAP formatter
+  So that I can get non-XML parseable results
+  As a Developer
+  I need to be able to use a TAP formatter
+
+  Scenario: Empty spec causes pending result
```
ciaranmcnulty: Looks like your scenario has suffered from scope creep - rather than rename it the feature could probably benefit from being split into several scenarios rather than one big one.

src/PhpSpec/Console/ContainerAssembler.php
padraic: It's not...removed

src/PhpSpec/Formatter/TapFormatter.php
padraic: This has shifted around a bit in last commit. Probably a few other pieces of cleanup, but this is close to done unless I start digging into data blocks. We'll see!

features/formatter/use_the_tap_formatter.feature
padraic: Sort of - I actually renamed the scenario after!

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,154 @@
...
+
+    private function indent($string)
+    {
+        return preg_replace(
+            array("%(^[^\n\r])%", "%([\n\r]{1,2})%", "%\s+...[\n\r]{1,2}\s+---%"),
+            array("  \\1", "\\1  ", ""),
+            $string
+        );
+    }
+
```
GrahamCampbell: there shouldn't be an extra line before the closing brace

features/formatter/use_the_tap_formatter.feature
ciaranmcnulty: That often happens initially with new features, then as the code grows they diverge more.

stof:
> Only really thought of it because PHPUnit basically passes through incompletes which never seemed right but I'd gotten used to it!

@padraic  This actually depends whether you run PHPUnit in strict mode or no.

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,92 @@
...
+                case 'pass':
+                return 'pass';
+                break;
+                case 'fail':
+                return 'fail';
+                break;
+                case 'broken':
+                throw new ErrorException('error','something terrible occurred','foo.php',99);
+                case 'skip':
+                throw new SkippingException('php is not installed');
```
stof: the SkippingException is meant to be used in spec examples, not in the code. Please make the scenario reflect the real usage of the exception

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,92 @@
...
+      use PhpSpec\Exception\Example\ErrorException;
+
+      class Tap
+      {
+          public function fire($stuff)
+          {
+              switch ($stuff) {
+                case 'pass':
+                return 'pass';
+                break;
```
stof: ``break;`` is dead code here

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,92 @@
...
+        severity: fail
+        ...
+      not ok 4 - Formatter\TapExample1\Tap: is most definitely broken
+        ---
+        message: 'error: something terrible occurred in foo.php line 99'
+        severity: fail
+        ...
+      ok 5 - Formatter\TapExample1\Tap: is most definitely skipping # SKIP skipped: php is not installed
+      1..5
+      """
```
stof: please add the final EOL

features/formatter/use_the_tap_formatter.feature
padraic: Noted - fixed these up to the correct 4 spaces.

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,161 @@
...
+     * @param int $result
+     * @return string
+     */
+    private function getResultData(ExampleEvent $event, $result = null)
+    {
+        if (null === $result) {
+            return $this->stripNewlines($event->getException()->getMessage());
+        }
+        $message = $event->getException()->getMessage();
+        switch ($result) {
```
ciaranmcnulty: I'm not sure it's worth using a switch for one case, YMMV

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,88 @@
+Feature: Use the TAP formatter
+  So that I can get non-XML parseable results
+  As a Developer
+  I need to be able to use a TAP formatter
+
+  Scenario: Empty spec causes various result states
```
ciaranmcnulty: 'Empty spec' doesn't seem to relate to the scenario.

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,88 @@
...
+                      throw new ErrorException('error','something terrible occurred','foo.php',99);
+              }
+          }
+      }
+      """
+    When I run phpspec using the "tap" format
+    Then I should see:
+      """
+      TAP version 13
+      not ok 1 - Formatter\TapExample1\Tap: is most definitely pending # TODO todo: write pending example
```
ciaranmcnulty: If this is meant to be human readable the last bit is maybe too confusing

features/formatter/use_the_tap_formatter.feature
```diff
@@ -0,0 +1,88 @@
...
+        ...
+      not ok 4 - Formatter\TapExample1\Tap: is most definitely broken
+        ---
+        message: 'error: something terrible occurred in foo.php line 99'
+        severity: fail
+        ...
+      ok 5 - Formatter\TapExample1\Tap: is most definitely skipping # SKIP skipped: php is not installed
+      1..5
+
+      """
```
ciaranmcnulty: Newline here please

features/formatter/use_the_tap_formatter.feature
padraic: Can you clarify? The first TODO is a directive, everything after that is the actual message (if any) straight from phpspec.

ciaranmcnulty: Appreciate that, I mean '# TODO todo: ' isn't clear as a result

padraic: The result per se is at the start of each primary line "ok" or "not ok". Anything following that will be description, directive, short message and/or YAML data (in that order with directive/data being optional). TAP assume a consumer can parse those out if needed.

ciaranmcnulty: I'll bow to your TAP knowledge

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,161 @@
...
+                $result = sprintf(self::OK, $this->examplesCount) . $desc . $message;
+                break;
+            case ExampleEvent::BROKEN:
+            case ExampleEvent::FAILED:
+                $message = $this->getResultData($event, $event->getResult());
+                $result = sprintf(self::NOT_OK, $this->examplesCount) . $desc . "\n" . $message;
+                break;
+        }
+
+        $this->getIO()->writeln($result);
```
ciaranmcnulty: I'm not really happy using `$result` here as it may not be defined. Can you add a `default:` with an appropriate message, or set a starting value for $result before the switch?

src/PhpSpec/Formatter/TapFormatter.php
```diff
@@ -0,0 +1,161 @@
...
+    private function getResultData(ExampleEvent $event, $result = null)
+    {
+        if (null === $result) {
+            return $this->stripNewlines($event->getException()->getMessage());
+        }
+        $message = $event->getException()->getMessage();
+        switch ($result) {
+            case ExampleEvent::PENDING:
+                $message = sprintf(self::TODO, $this->stripNewlines($message))
+                    . $this->indent(sprintf(self::SEVERITY, Yaml::dump('todo')));
```
ciaranmcnulty: Scrutinizer points out that `Yaml::dump` is typehinted to accept `array` - is the use of a string here safe?

ciaranmcnulty:
The complexity of the examples for `TapFormatter`, along with your observation that those examples mirror the Behat suite, both point to SRP violations in the Formatter.

However this is probably not your fault, it points to the fact PhpSpec requires a formatter to do too many things, perhaps. It might be an area we can focus on in future.

There are a couple of minor comments to address, otherwise this PR is in good shape.

features/formatter/use_the_tap_formatter.feature
padraic: You may depart our presence ;)

src/PhpSpec/Formatter/TapFormatter.php
padraic: Sure thing. What would most appropriate default be if an example had no result? It does seem unlikely for phpspec to propagate resultless examples.

ciaranmcnulty: I'm thinking more like future result types (Skipped was fairly recent for instance). In 2020 when we have quantum computing and someone adds 'ExampleEvent::UNCOLLAPSED' I don't want your formatter to break ;-)

 I think it should be 'not OK' so that someone comes and shouts at me.

padraic:
@ciaranmcnulty The problem is that TAP is not one format, but three. You have the regular expected lines, then directives, followed by optional YAML data blocks. For all that it's a relatively short class thankfully. The specs are complex - and may be overkill since the behat feature now hits all the main points at once. Not sure whether the possible duplication is worth retaining.

ciaranmcnulty:
It's worth keeping in IMO. As soon as you add collaborators to split up the class responsibilities the specs and features start looking quite different but it's very common for the initial set to look similar.

Having both right now helps clarity when we want to know how a class behaves (Features describe the overall system, if I want to find out what `TapFormatter` is for I'll naturally look for `TapFormatterSpec`).

ciaranmcnulty:
For instance, the switch in `TapFormatter` could in future be replaced with something like...

```php
$lineFormatterFactory->getLineFormatterFor($event)->outputLine();
```

+ you'd introduce a concept of a`LineFormatter` interface (`FailedLineFormatter`, `SkippedLineFormatter`...) and a factory with the switch in. At that stage your specs will be competely different to your feature's scenarios.

src/PhpSpec/Formatter/TapFormatter.php
stof: this is actually a wrong phpdoc in the Yaml class

padraic: It a correct usage. The Yaml class must have a bad phpdoc ref.

src/PhpSpec/Formatter/TapFormatter.php
padraic: Will do

src/PhpSpec/Formatter/TapFormatter.php
ciaranmcnulty: Ah yes. `Yaml\Dumper::dump` is correctly hinted as `mixed` but the facade in `Yaml::dump` is `array` (and the parameter is called `$array`). My main concern was whether this might change but I guess not.

ciaranmcnulty:
:+1: looks good to me

@stof Any more comments?

stof:
:+1:

ciaranmcnulty:
Great, thanks!

padraic:
Thanks all!

PR #615, opened 2015-01-18
Adding smoke test running command at CLI

ciaranmcnulty:
Covers two critical scenarios by running them through `sh` and interacting as a `tty`.

### DIFF
--- a/features/bootstrap/IsolatedProcessContext.php
+++ b/features/bootstrap/IsolatedProcessContext.php
@@ -5,7 +5,6 @@ use Behat\Behat\Context\Context;
 use Behat\Behat\Context\SnippetAcceptingContext;
 use Behat\Gherkin\Node\PyStringNode;
 use Behat\Gherkin\Node\TableNode;
-use PhpSpec\Exception\Example\SkippingException;
 use Symfony\Component\Process\Process;
 
 /**
@@ -31,10 +30,6 @@ class IsolatedProcessContext implements Context, SnippetAcceptingContext
      */
     public function iRunPhpspecAndAnswerWhenAskedIfIWantToGenerateTheCode($answer)
     {
-        if (!`which expect`) {
-            throw new SkippingException('Cannot test interactive scripts on this platform');
-        }
-
         $process = new Process(
             "exec expect -c '\n" .
             "set timeout 10\n" .
@@ -42,13 +37,12 @@ class IsolatedProcessContext implements Context, SnippetAcceptingContext
             "expect \"Y/n\"\n" .
             "sleep 0.1\n" .
             "send \"$answer\n\"\n" .
-            "sleep 0.5\n" .
+            "sleep 1\n" .
             "interact\n" .
             "'"
         );
 
         $process->run();
-
         $this->lastOutput = $process->getOutput();
 
         expect($process->getErrorOutput())->toBe(null);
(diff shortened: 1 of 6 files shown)


ciaranmcnulty:
Travis doesn't love me.

features/bootstrap/IsolatedProcessContext.php
```diff
@@ -0,0 +1,72 @@
...
+        expect($process->getExitCode())->toBe(0);
+    }
+
+    /**
+     * @When I run phpspec and answer :answer when asked if I want to generate the code
+     */
+    public function iRunPhpspecAndAnswerWhenAskedIfIWantToGenerateTheCode($answer)
+    {
+        if (!`which expect`) {
+            throw new SkippingException('Cannot test interactive scripts on this platform');
```
stof: a PhpSpec SkippingException in Behat ? You expect too much :)

ciaranmcnulty: PhpStorm hides it from me ;-)

I want to skip when `expect` is not available, but it seems to not be possible I think?

ciaranmcnulty:
Got this passing on travis but:

* Had to use sudo, so we are not on the container architecture :-(
* Users who check out the code and run the specs may be missing the command so will get failures

GrahamCampbell:
This will mean we can't run the test suite on windows?

GrahamCampbell:
We'll need to skip if we're windows and if expect is not available?

ciaranmcnulty:
That's one issue yes. Looks like it fails on HHVM too.

Suggestions about how to do this without any UNIX one-liners welcomed. Symfony\Console is quick to assume it's not executing in a TTY so a lot of strategies won't allow prompting. Symfony\Process's `isTty()` did not do what I expected.

features/bootstrap/IsolatedProcessContext.php
```diff
@@ -0,0 +1,66 @@
...
+            "send \"$answer\n\"\n" .
+            "sleep 1\n" .
+            "interact\n" .
+            "'"
+        );
+
+        $process->run();
+        $this->lastOutput = $process->getOutput();
+
+        expect($process->getErrorOutput())->toBe(null);
```
stof: this is failing on some Symfony Process versions (getting an empty string instead of null)

ciaranmcnulty: I'll change it to a Boolean check

ciaranmcnulty:
I don't have HHVM so I'm unclear whether the failing HHVM build is actually indicating a real problem with rerunning.

ciaranmcnulty:
Moved the smoke tests into a separate profile, and only run them on some builds (this also means people without expects can run behat suite without a surprise)

The HHVM failure is a worry, if anyone has HHVM and can debug that would be appreciated.

ciaranmcnulty:
Looks like the HHVM fail is a genuine bug! Let's hear it for smoke tests :-)

Will leave this open until I can repro+fix the bug on master.

ciaranmcnulty:
This is in a pretty good place now - it's passing on all platforms and the fact it helped me find an issue is a :+1: for having it.

The only downside is that it takes us off the container-based architecture :-/

stof:
Taking us off the new infrastructure also means we cannot have the composer cache anymore. Is there a better way to get ``expect`` on Travis without using apt-get (maybe donwloading a prebuilt binary and putting it in the path ?)

ciaranmcnulty:
Prebuilt binary is a good idea, I'll investigate.

It would be nice to be able to do what `expect` is doing in PHP but I've been unsuccessful in my attempts so far.

ciaranmcnulty:
It looks like expect is pretty messy to build, unfortunately.

ciaranmcnulty:
I spent some time investigating different tools to do this but came up blank. My inclination at the moment is to take the performance penalty and merge this. The smoke tests are more valuable than 30% less execution time, as we're not often waiting for builds to finish.

PR #621, opened 2015-01-23
replace ternaries with array_merge

hanneskaeufler:
My try to improve on what scrutinizer thinks is a critical offender of code quality. Tbh I don´t even know if I like this better.


ciaranmcnulty:
Looks good - I don't want to chase the Scrutiniser score blindly butI think this version is slightly more readable (fewer variables in local scope).

Thanks!

PR #622, opened 2015-01-23
Hk test dot formatter

hanneskaeufler:
I did not find any specs or other coverage for the `DotFormatter` so for practice I whipped some up. This then allowed me to move some things around. This is not 100% coverage for sure, but better than none :)


ciaranmcnulty:
Thanks!

Issue #632, opened 2015-02-10
Make classes final by default


ciaranmcnulty:
This would be a fairly simple template change.

Would it kill our user base? :-)

/cc @everzet

ciaranmcnulty:
Pro: Promotes something that I'd consider a best practice

Con: Might freak out some newcomers (this isn't that important)

stof:
should we make this configurable (without having to overwrite the whole template, which is kind of hard as it forces to update it for any improvement done in PhpSpec later) ? This way, we could satisfy both people wanting final classes and people not liking them

ciaranmcnulty:
We could; we could have two templates and switch between them based on config.

@mdavis1982 had a PR that implemented this at one stage.

I'm not sure about having two ways to configure it though, and if we do it it should default to 'ON'

mdavis1982:
I still have the code for this PR lying around... At the time, I didn't know how to add the tests needed for the feature to be included. Now that I do, I'll try to resurrect it ASAP

MarcelloDuarte:
+1 for configurable, not default behaviour

ciaranmcnulty:
default on or default off, though?

I mean, it's configurable right now by making a personal class template in `~`

norzechowicz:
Maybe it could be a question before generating class with default value "yes"?
Or maybe class could be generated as finall when there will be something like $this->shouldBeAFinal() when class does not exists yet?

MarcelloDuarte:
I suggest that by default it is off, but a simple flag in phpspec.yml would turn the final template on.

ciaranmcnulty:
If the default is off, I don't see any benefit in having the option - if a user is savvy enough to switch it on in phpspec.yml they're savvy enough to edit the template.

mdavis1982:
@ciaranmcnulty I've just open the above PR to address this.

stof:
@ciaranmcnulty overwriting the template means that they need to sync their custom template with changes done in phpspec all the time. It is less maintainable than a flag in phpspec.yml

MarcelloDuarte:
Here's an idea, if `shouldHaveType` or any [type matcher](http://www.phpspec.net/en/latest/cookbook/matchers.html#type-matcher) is used, and the interface does not exist, then offer to create the interface AND ask if you want the class to be final. (Inspired by @Ocramius [blog post](http://ocramius.github.io/blog/when-to-declare-classes-final/))

ciaranmcnulty:
[...]

PR #633, opened 2015-02-11
Add option to generate final classes

mdavis1982:
This adds the option to generate final classes by specifying:

```yaml
generator.final_classes: true
```

In the `phpspec.yml` file. By default the option is switched off.

This addresses #632.

### DIFF
--- a/src/PhpSpec/Console/ContainerAssembler.php
+++ b/src/PhpSpec/Console/ContainerAssembler.php
@@ -201,9 +201,15 @@ class ContainerAssembler
             );
         });
         $container->set('code_generator.generators.class', function (ServiceContainer $c) {
+            $generateFinalClasses = (bool) $c->getParam('generator.final_classes', false);
+
             return new CodeGenerator\Generator\ClassGenerator(
                 $c->get('console.io'),
-                $c->get('code_generator.templates')
+                $c->get('code_generator.templates'),
+                null,
+                array(
+                    'generator.final_classes' => $generateFinalClasses
+                )
             );
         });
         $container->set('code_generator.generators.method', function (ServiceContainer $c) {
--- a/features/code_generation/developer_generates_class.feature
+++ b/features/code_generation/developer_generates_class.feature
@@ -19,6 +19,26 @@ Feature: Developer generates a class
 
       """
 
+  @issue632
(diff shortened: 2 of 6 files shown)


ciaranmcnulty:
You're using 5.4 array syntax I think...

Could you also write a feature for this so we can get an acceptance test?

mdavis1982:
There are already instances of the 5.4 array syntax in the same files that I modified - which is why I used the 5.4 syntax! :wink:

mdavis1982:
I'll try to add a feature for this... I'm not 100% sure what I'm doing, but I'll give it a go anyway

mdavis1982:
I've added the acceptance test and it passes :smile:

ciaranmcnulty:
There are no instances of 5.4 syntax in phpspec...

ciaranmcnulty:
I like the feature - it's cool you didn't have to write any code ;-)

mdavis1982:
I've removed the 5.4 array syntax... The awkward moment you realise you have PHPStorm set to force the syntax on save!

It's very cool I didn't have to write any code! Behat is cool! :smile:

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -201,9 +201,15 @@ private function setupGenerators(ServiceContainer $container)
...
         });
         $container->set('code_generator.generators.class', function (ServiceContainer $c) {
+            $generateFinalClasses = (bool) $c->getParam('generator.final_classes', false);
+
             return new CodeGenerator\Generator\ClassGenerator(
                 $c->get('console.io'),
-                $c->get('code_generator.templates')
+                $c->get('code_generator.templates'),
+                null,
+                [
```
ciaranmcnulty: brackets here

mdavis1982:
Gah! Fixed!

ciaranmcnulty:
This looks good, cheers.

Now to decide whether it's a good feature :-)

rtuin:
Nice PR. If it is decided that this will be a config option I propose a more natural language approach: "generate.final_classes" for example.

Also an entry in the documentation would be good when the feature gets introduced.

ciaranmcnulty:
Closing in favour of #637

PR #636, opened 2015-02-12
add Windows support

strebl:
With that file you're able to call phpspec with `vendor/bin/phpspec` like you are on a unix based os.

This is relevant if you use something like laravel elixir.

### DIFF
--- a/bin/phpspec.bat
+++ b/bin/phpspec.bat
@@ -1,0 +1,3 @@
+@ECHO OFF
+SET BIN_TARGET=%~dp0/../phpspec/phpspec/bin/phpspec
+php "%BIN_TARGET%" %*


bin/phpspec.bat
```diff
@@ -0,0 +1,3 @@
+@ECHO OFF
+SET BIN_TARGET=%~dp0/../phpspec/phpspec/bin/phpspec
```
stof: this path is wrong AFAICT

stof:
Composer already generates a bat binary when running it on Windows (and it generates a working one).

And your PR does not allow to run ``vendor/bin/phpspec``, because you don't ask Composer to put the bat file in ``vendor/bin`` anyway

ciaranmcnulty:
This is an attempt to avoid having to prefix with `php`? Like @stof says I was under the impression this already worked.

strebl:
My bad... Composer not created the .bat file. I deleted composer.lock, the vendor directory and added my fork. the installation added the .bat file to the bin directory. I wasn't my change...

strebl:
But why does, for example behat, add the file to their github repo and phpspec not, what do I miss?

ciaranmcnulty:
It looks like Behat supports a PEAR install, so has to build that in to the `.bat`

strebl:
Thanks for your help.
I'll test better next time :disappointed:

ciaranmcnulty:
No, thank you!

It's great that you try and help the tool improve.

stof:
@ciaranmcnulty Behat used to support a PEAR install (the PEAR channel has been discontinued, but we indeed haven't removed the bat file)

PR #640, opened 2015-02-17
New matcher asserting $array[$key] === $expectedValue

echochamber:
New matcher that allows for asserting a specific value for a specific key of methods that return an array and or an implementor of ArrayAccess.

Example Spec:

```php
<?php

SomeClassSpec extends ObjectBehavior
{

    function it_returns_an_array_with_myValue_for_myKey()
    {
        $this->someMethodThatReturnsAnArray()->shouldHaveKeyWithValue('myKey', 'myValue');
    }
}
```

Example method under test
```php
<?php

SomeClass
{

    public function someMethodThatReturnsAnArray()
    {
        return ['myKey' => 'myValue'];
    }
}

```

### DIFF
--- a/src/PhpSpec/Matcher/ArrayKeyValueMatcher.php
+++ b/src/PhpSpec/Matcher/ArrayKeyValueMatcher.php
@@ -1,0 +1,118 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Matcher;
+
+use ArrayAccess;
+use PhpSpec\Exception\Example\FailureException;
+use PhpSpec\Formatter\Presenter\PresenterInterface;
+
+class ArrayKeyValueMatcher extends BasicMatcher
+{
+    /**
+     * @var \PhpSpec\Formatter\Presenter\PresenterInterface
+     */
+    private $presenter;
+
+    /**
+     * @param PresenterInterface $presenter
+     */
+    public function __construct(PresenterInterface $presenter)
+    {
+        $this->presenter = $presenter;
+    }
+
+    /**
+     * @param string $name
+     * @param mixed  $subject
+     * @param array  $arguments
+     *
+     * @return bool
+     */
+    public function supports($name, $subject, array $arguments)
+    {
+        return 'haveKeyWithValue' === $name
+        && 2 == count($arguments)
+        && (is_array($subject) || $subject instanceof ArrayAccess)
+            ;
+    }
+
+    /**
+     * @param ArrayAccess|array $subject
+     * @param array $arguments
+     *
+     * @return bool
+     */
+    protected function matches($subject, array $arguments)
+    {
+        $key = $arguments[0];
+        $value  = $arguments[1];
+
+        if ($subject instanceof ArrayAccess) {
+            return $subject->offsetExists($key) && $subject->offsetGet($key) === $value;
+        }
+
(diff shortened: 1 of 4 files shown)


ciaranmcnulty:
Can you add some feature coverage for this please?

src/PhpSpec/Matcher/ArrayKeyValueMatcher.php
```diff
@@ -0,0 +1,104 @@
+<?php
+
```
ciaranmcnulty: There needs to be a docblock here with copyright info

src/PhpSpec/Matcher/ArrayKeyValueMatcher.php
```diff
@@ -0,0 +1,104 @@
...
+        // Match failed because the key wasn't set
+        if (($subject instanceof ArrayAccess && $subject->offsetExists($key)) || array_key_exists($key, $subject)) {
+            return new FailureException(sprintf('Expected %s to have value %s for %s key, but no key was set.',
+                $this->presenter->presentValue($subject),
+                $this->presenter->presentValue($expectedValue),
+                $this->presenter->presentString($key)
+            ));
+        }
+
+        // Match failed because key was set but the value wasn't what was expected
```
ciaranmcnulty: don't think you need these comments.

src/PhpSpec/Matcher/ArrayKeyValueMatcher.php
```diff
@@ -0,0 +1,104 @@
...
+     * @return FailureException
+     */
+    protected function getFailureException($name, $subject, array $arguments)
+    {
+        $key = $arguments[0];
+        $expectedValue = $arguments[1];
+        $actualValue = $subject[$key];
+
+        // Match failed because the key wasn't set
+        if (($subject instanceof ArrayAccess && $subject->offsetExists($key)) || array_key_exists($key, $subject)) {
```
ciaranmcnulty: conditional would be clearer if extracted to a private method

ciaranmcnulty:
@echochamber are you going to get a chance to look at this again?

echochamber:
@ciaranmcnulty Hey, thanks for the ping. I totally forgot I had opened this PR. I'm busy tonight but I'll make this my priority tomorrow.

ciaranmcnulty:
Great thanks, if you need help with the feature coverage let me know

src/PhpSpec/Matcher/ArrayKeyValueMatcher.php
```diff
@@ -0,0 +1,104 @@
...
+     *
+     * @return FailureException
+     */
+    protected function getFailureException($name, $subject, array $arguments)
+    {
+        $key = $arguments[0];
+        $expectedValue = $arguments[1];
+        $actualValue = $subject[$key];
+
+        // Match failed because the key wasn't set
```
ciaranmcnulty: please remove the comment

It's better to pull bits of this out into privates so that the overall shape is clear

ciaranmcnulty:
We're intending to release 2.2.0-beta after the next Prophecy release so if you can get this done in the next week or so I can bundle it in - it's a good addition.

We'll also need documentation but we can raise that separately to this PR

cakper:
@ciaranmcnulty @echochamber by asserting key in the array you are asserting the state of the object and not it's behaviour. I don't consider this as a good object oriented design practice and I think we should not encourage people to do so by enabling such features by default.

ciaranmcnulty:
We assert state of return values all the time (e.g. `shouldMatch`)

ciaranmcnulty:
Unless you're arguing against methods that return arrays.

echochamber:
@ciaranmcnulty Ok, I have fixed all the issues addressed in your comments and added a feature. If there are any other changes you'd like please let me know :smile:

echochamber:
@ciaranmcnulty Also regarding documentation, I can write some and open a PR once this one is merged.

ciaranmcnulty:
The feature doesn't appear to pass

echochamber:
@ciaranmcnulty Fixed now.

ciaranmcnulty:
Thanks!

PR #643, opened 2015-02-25
- Fixed broken link & Added clarification that by default phpspec creates mock classes

jeroenvdgulik:
- Fixed broken link
- Added clarification that by default phpspec creates mock classes

Pretty salty that I spend 30 min figuring out why my tests kept failing only to find out I was working with a mock the entire time instead of the actual object. This is mentioned nowhere in the docs, so I added this for future beginners.

### DIFF
--- a/docs/cookbook/construction.rst
+++ b/docs/cookbook/construction.rst
@@ -73,6 +73,29 @@ You can tell **phpspec** to pass values to the constructor when it constructs th
             }
         }
 
+Keep in mind that by default, **phpspec** creates a mock object. If you want
+to inject an actual object you could do it like this:
+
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
We can certainly make it less ambiguous - I'd prefer linking back to an example in the Doubles chapter than having a code block in every instance where we use a type hint.

e.g. your 'note this is a double' is fine, but it could just do 'see here for more detail' and we can make the Doubles chapter more clear perhaps?

ciaranmcnulty:
The broken link was fixed in #646 and I'm not convinced we need that notice, sorry.

PR #644, opened 2015-02-26
Patch MethodAnalyser to find code of the method in traits from different files

thomasruiz:
When using traits from different file, the analyzer fails to find the right code.
This commit fixes it by looking for traits recursively based on [this post](http://mouf-php.com/blog/php_reflection_api_traits).

I am not sure about the flat require in the spec, maybe it would be better to [define autoload-dev in composer.json](https://getcomposer.org/doc/04-schema.md#autoload-dev).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,10 @@ install:
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
+before_script:
+  - echo "<?php if (PHP_VERSION_ID >= 50400) echo ',@php5.4';" > php_version_tags.php
+
 script:
    - bin/phpspec run --format=pretty
    - ./vendor/bin/phpunit --testdox
-   - ./vendor/bin/behat --format=pretty `if [ "$SMOKE" == "true" ]; then echo '--profile=smoke'; fi;`
-
+   - ./vendor/bin/behat --format=pretty `if [ "$SMOKE" == "true" ]; then echo '--profile=smoke'; fi;` --tags '~@php-version'`php php_version_tags.php`
--- a/features/bootstrap/FilesystemContext.php
+++ b/features/bootstrap/FilesystemContext.php
@@ -59,8 +59,9 @@ class FilesystemContext implements Context, MatchersProviderInterface
     /**
      * @Given the class file :file contains:
      * @Given the spec file :file contains:
+     * @Given the trait file :file contains:
      */
-    public function theClassOrSpecFileContains($file, PyStringNode $contents)
+    public function theClassOrTraitOrSpecFileContains($file, PyStringNode $contents)
     {
         $this->theFileContains($file, $contents);
         require_once($file);
--- a/features/code_generation/developer_generates_return_constant.feature
+++ b/features/code_generation/developer_generates_return_constant.feature
@@ -259,3 +259,66 @@ Feature: Developer generates a method returning a constant
       """
     When I run phpspec interactively
     Then I should be prompted for code generation
+
+  @php-version @php5.4
+  Scenario: Generating a scalar return type when method is in trait
+    Given the spec file "spec/CodeGeneration/ConstantExample7/MarkdownSpec.php" contains:
+      """
+      <?php
+
+      namespace spec\CodeGeneration\ConstantExample7;
+
+      use PhpSpec\ObjectBehavior;
+      use Prophecy\Argument;
+
+      class MarkdownSpec extends ObjectBehavior
+      {
+          function it_converts_plain_text_to_html_paragraphs()
+          {
+              $this->toHtml('Hi, there')->shouldReturn('<p>Hi, there</p>');
+          }
+      }
+
+      """
+    And the trait file "src/CodeGeneration/ConstantExample7/MarkdownTrait.php" contains:
+      """
+      <?php
+
+      namespace CodeGeneration\ConstantExample7;
+
+      trait MarkdownTrait
+      {
+          public function toHtml($argument1)
+          {
+          }
+      }
+
+      """
+    And the class file "src/CodeGeneration/ConstantExample7/Markdown.php" contains:
+      """
+      <?php
+
+      namespace CodeGeneration\ConstantExample7;
+
+      class Markdown
+      {
+          use MarkdownTrait;
+      }
+
+      """
+    When I run phpspec with the option "fake" and answer "y" when asked if I want to generate the code
+    Then the class in "src/CodeGeneration/ConstantExample7/MarkdownTrait.php" should contain:
+      """
+      <?php
+
+      namespace CodeGeneration\ConstantExample7;
+
+      trait MarkdownTrait
+      {
+          public function toHtml($argument1)
+          {
+              return '<p>Hi, there</p>';
+          }
+      }
+
+      """
--- a/spec/PhpSpec/Util/ExampleObjectUsingTrait.php
+++ b/spec/PhpSpec/Util/ExampleObjectUsingTrait.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace spec\PhpSpec\Util;
+
+class ExampleObjectUsingTrait
+{
+    use ExampleTrait, AnotherExampleTrait;
+}
--- a/spec/PhpSpec/Util/MethodAnalyserSpec.php
+++ b/spec/PhpSpec/Util/MethodAnalyserSpec.php
@@ -2,6 +2,7 @@
 
 namespace spec\PhpSpec\Util;
 
+use PhpSpec\Exception\Example\SkippingException;
 use PhpSpec\ObjectBehavior;
 use Prophecy\Argument;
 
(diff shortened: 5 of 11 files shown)


spec/PhpSpec/Util/MethodAnalyserSpec.php
```diff
@@ -35,10 +35,20 @@ function it_identifies_internal_classes_as_non_empty()
...
     }
+
+    function it_identifies_methods_from_traits()
+    {
+        $this->methodIsEmpty('spec\PhpSpec\Util\ExampleObject', 'emptyMethodInTrait')->shouldReturn(true);
+        $this->methodIsEmpty('spec\PhpSpec\Util\ExampleObject', 'nonEmptyMethodInTrait')->shouldReturn(false);
+    }
 }
 
+require 'ExampleTrait.php';
```
stof: please use autoload-dev instead indeed

spec/PhpSpec/Util/MethodAnalyserSpec.php
```diff
@@ -35,10 +35,20 @@ function it_identifies_internal_classes_as_non_empty()
     {
         $this->methodIsEmpty('DateTimeZone', 'getOffset')->shouldReturn(false);
     }
+
+    function it_identifies_methods_from_traits()
```
stof: this should be marked as skipped when running the example on PHP 5.3

spec/PhpSpec/Util/MethodAnalyserSpec.php
```diff
@@ -35,10 +35,20 @@ function it_identifies_internal_classes_as_non_empty()
...
+        $this->methodIsEmpty('spec\PhpSpec\Util\ExampleObject', 'emptyMethodInTrait')->shouldReturn(true);
+        $this->methodIsEmpty('spec\PhpSpec\Util\ExampleObject', 'nonEmptyMethodInTrait')->shouldReturn(false);
+    }
 }
 
+require 'ExampleTrait.php';
+
 class ExampleObject
 {
+    use ExampleTrait;
```
stof: please use a different class defined in its own file, so that PHP 5.3 can still run all other examples

src/PhpSpec/Util/MethodAnalyser.php
```diff
@@ -50,15 +50,47 @@ public function reflectionMethodIsEmpty(\ReflectionMethod $method)
...
 
         return preg_replace('/.*function[^{]+{/s', '', $code);
     }
 
+    private function getDeclaringClass(\ReflectionMethod $reflectionMethod, $methodStartLine, $methodEndLine)
+    {
+        $reflectionClass = $reflectionMethod->getDeclaringClass();
+
+        $fileName = $reflectionMethod->getFileName();
+        $trait = $this->getDeclaringTrait($reflectionClass->getTraits(), $fileName, $methodStartLine, $methodEndLine);
```
stof: this is broken on PHP 5.3 (which does not have the ``getTraits`` method)

src/PhpSpec/Util/MethodAnalyser.php
```diff
@@ -50,15 +50,47 @@ public function reflectionMethodIsEmpty(\ReflectionMethod $method)
...
+     *
+     * @return null|\ReflectionClass
+     */
+    private function getDeclaringTrait(array $traits, $file, $start, $end)
+    {
+        foreach ($traits as $trait) {
+            if ($trait->getFileName() == $file && $trait->getStartLine() <= $start && $trait->getEndLine() >= $end) {
+                return $trait;
+            }
+            return $this->getDeclaringTrait($trait->getTraits(), $file, $start, $end);
```
stof: this is wrong. You are always returning something for the first trait in ``$traits``, meaning that multiple traits on the class are not supported

ciaranmcnulty:
This is used to detect whether there is any body in the method, so that we can use the `--fake` option to write methods.

This is a good start but to properly make this work with Traits we will probably also need to modify the `ReturnConstantGenerator`

spec/PhpSpec/Util/ExampleObjectUsingTrait.php
```diff
@@ -0,0 +1,6 @@
+<?php namespace spec\PhpSpec\Util;
```
ciaranmcnulty: Is there a reason you've not inlined this into the spec? It's what we've done in other similar situations.

ciaranmcnulty:
Is there some feature coverage you can add that shows the motivation driving this change?

spec/PhpSpec/Util/ExampleObjectUsingTrait.php
thomasruiz: Yes: if the trait is in the same file, there is no issue (due to `ReflectionMethod::getStartLine()` and `ReflectionMethod::getEndLine()` refering to the method defined in the trait). The bug is here only when the trait is in a different file.

ciaranmcnulty: Aha, right

thomasruiz:
I'll work on the feature coverage during the afternoon.

If I find any problem with ReturnConstantGenerator I will try to fix it :)

ciaranmcnulty:
The scenario will be something like (not real steps):

Given an empty method is defined in a trait
And there is a spec expecting that method to return a scalar value
When I run phpspec with `--fake`
Then a return value is generated in the trait

ciaranmcnulty:
Actually... that might just work anyhow due to how reflection works, but it'll be good to have the feature in there explicitly saying how it should work

thomasruiz:
Indeed, there was an error related to `MethodReturnedNullListener`.

Should I clean the commit history with git rebase?

ciaranmcnulty:
That's optional - you can rebase if you want to

src/PhpSpec/Util/MethodAnalyser.php
```diff
@@ -44,28 +44,90 @@ public function reflectionMethodIsEmpty(\ReflectionMethod $method)
...
     }
 
     /**
+     * @param  \ReflectionMethod $reflectionMethod
+     * @param  int $methodStartLine
+     * @param  int $methodEndLine
+     *
+     * @return \ReflectionClass
+     */
+    private function getDeclaringClass(\ReflectionMethod $reflectionMethod, $methodStartLine, $methodEndLine)
```
ciaranmcnulty: This isn't really returning a class any more, is it?

thomasruiz: It is returning a class. Maybe there could be improvements in the methods' names... In summary, there is:

(public) getRealDeclaringClass: get the name of the real class (or trait) declaring the method
(private) getDeclaringClass: get the ReflectionClass on the real class (or trait) declaring the ReflectionMethod
(private) getDeclaringTrait: either return the trait defining a method or null if the method is defined in the class

ciaranmcnulty: I clicked on the wrong line ;-)

getRealDeclaringClass is the one that seems wrong, as it's not really (always) returning a class.

ciaranmcnulty: Not that I can think of a name, frankly.

features/bootstrap/FilesystemContext.php
```diff
@@ -67,6 +67,19 @@ public function theClassOrSpecFileContains($file, PyStringNode $contents)
     }
 
     /**
+     * @Given the trait file :file contains:
+     */
+    public function theTraitFileContains($file, PyStringNode $contents)
+    {
+        if (version_compare(PHP_VERSION, '5.4.0', '<')) {
+            throw new PendingException('Traits implemented since PHP 5.4');
```
ciaranmcnulty: Pending isn't appropriate in this context

src/PhpSpec/Util/MethodAnalyser.php
thomasruiz: Naming things is always a problem :)
How about getRealMethodOwner?

features/bootstrap/FilesystemContext.php
thomasruiz: How would you mark it as skipped? I didn't find a way in Behat to do so

ciaranmcnulty: You can't. Behat's own suite uses tags (e.g. @php-5.4) on the scenario and a selectively filters the tags when running

ciaranmcnulty: https://github.com/Behat/Behat/blob/master/.travis.yml#L32-L36

We could use something like this.

spec/PhpSpec/Util/ExampleObjectUsingTrait.php
ciaranmcnulty: namespace declaration should be on a different line to the opening `<?php`

.travis.yml
```diff
@@ -30,9 +30,12 @@ install:
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
+before_script:
+  - echo "<?php echo '@php5' . implode(',@php5.', range(3, PHP_MINOR_VERSION));" > php_version_tags.php
```
ciaranmcnulty: As we're only detecting 5.4 right now this seems overkill

src/PhpSpec/Util/MethodAnalyser.php
ciaranmcnulty: I like `getMethodOwner`

ciaranmcnulty:
There are only a couple of minor things to be addressed here, if you can get it tidied up I'll merge for 2.2.0

thomasruiz:
All done :)

features/code_generation/developer_generates_return_constant.feature
```diff
@@ -258,4 +258,67 @@ Feature: Developer generates a method returning a constant
       fake: true
       """
     When I run phpspec interactively
-    Then I should be prompted for code generation
\ No newline at end of file
+    Then I should be prompted for code generation
+
+  @php-version @php5.4
```
ciaranmcnulty: what is `@php-version` doing?

.travis.yml
```diff
@@ -32,8 +32,10 @@ install:
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
+before_script:
+  - echo "<?php if (PHP_VERSION_ID >= 50400) echo ',@php5.4';" > php_version_tags.php
```
ciaranmcnulty: Writing to the temp file seems wrong, you can evaluate this into the statement similar to how the smoke tests are executed

ciaranmcnulty:
I'll fiddle with the suite config myself

Thanks for the contribution!

.travis.yml
stof: There was an issue on HHVM when inlining it at the time @everzet did this for Behat

ciaranmcnulty: Ah, thanks. I'll see if it's still current

Issue #585, opened 2014-12-17
Class does not exist loop


ciaranmcnulty:
This is a FAQ: If an autoloader is not configured, users get trapped in a loop where they are prompted to create the class, suite reruns, class does not exist,users are re-prompted.

This can be solved by checking class_exists directly after writing the file, although currently there's not a neat place to do this.

codereviewvideos:
To add a little extra helper here, make sure your composer.json is as per the docs: 

http://www.phpspec.net/en/latest/

And also, run `composer dumpautoload` after making any adjustments. Solved it for me :)

PR #651, opened 2015-03-08
Check for successful autoloading after class is generated

ciaranmcnulty:
This is an FAQ when users have not correctly configured their autoloader and can get stuck in a 'Do you want to generate the class?' loop (see #585)

Now after a class is generated a quick `class_exists` checks whether the class has been written out to an autoloadable path and tells the user / cancels suite re-run if it is not the case.

autoloader

### DIFF
--- a/src/PhpSpec/Listener/ClassNotFoundListener.php
+++ b/src/PhpSpec/Listener/ClassNotFoundListener.php
@@ -76,13 +76,11 @@ class ClassNotFoundListener implements EventSubscriberInterface
 
             if ($this->io->askConfirmation($message)) {
                 $this->generator->generate($resource, 'class');
-                if (class_exists($classname)) {
-                    $event->markAsWorthRerunning();
-                }
-                else {
+                if (!class_exists($classname)) {
                     $message = 'File was written but the class %s was not autoloadable - do you have an autoloader configured?';
                     throw new ClassGenerationFailedException(sprintf($message, $classname));
                 }
+                $event->markAsWorthRerunning();
             }
         }
     }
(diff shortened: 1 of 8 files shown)


src/PhpSpec/Listener/ClassNotFoundListener.php
```diff
@@ -75,7 +76,13 @@ public function afterSuite(SuiteEvent $event)
 
             if ($this->io->askConfirmation($message)) {
                 $this->generator->generate($resource, 'class');
-                $event->markAsWorthRerunning();
+                if (class_exists($classname)) {
+                    $event->markAsWorthRerunning();
+                }
+                else {
```
richardmiller: You could lose this else by throwing the exception if the class doesn't exist first.

src/PhpSpec/Listener/ClassNotFoundListener.php
```diff
@@ -75,7 +76,13 @@ public function afterSuite(SuiteEvent $event)
...
+                    $message = 'File was written but the class %s was not autoloadable - do you have an autoloader configured?';
```
richardmiller: Is there more detail in the docs on this? If so, is it worth having the url for that in the message as well?

src/PhpSpec/Listener/ClassNotFoundListener.php
ciaranmcnulty: nice, thanks

src/PhpSpec/Listener/ClassNotFoundListener.php
ciaranmcnulty: There's not; I'll raise a documentation issue

PR #654, opened 2015-03-10
Test on PHP 7.0, allow it to fail.

frankdejonge:


ciaranmcnulty:
Nice, thanks.

How stable is the PHP7 feature set? I'm behind on internals.

What I mean is: at what point is it worth investing time trying to make it work?

frankdejonge:
@ciaranmcnulty judging by the PHP 7 release timeline (https://wiki.php.net/rfc/php7timeline) I'd say investing time into "getting it to work" would be somewhere around September-ish. Though I think it's good to see what the project does on this version regardless. In my project, the phpspec tests fail on php 7 (only), which has to do with the `let()` double resolving. I figured this would put that on the radar, if it still exists by that time.

ciaranmcnulty:
Makes sense to me, and we 'fail early' so it won't delay build results.

ciaranmcnulty:
Would it make sense for the next release to specify a composer dependency on PHP <7.0 ?

frankdejonge:
@ciaranmcnulty you could specify the php version as `^5.3.3` which is the equivalent of `>=5.3.3, <6.0.0`.

ciaranmcnulty:
I'm a bit nervous about ^ in the next tagged release, and whether it'll break people using stable composer releases

stof:
@ciaranmcnulty I would not change the requirement to forbid PHP 7 for now (unless you first debug the reason for each failure to know whether it is an intended BC break in PHP 7 or a regression)

ciaranmcnulty:
True, it's probably best to worry more about it later.

frankdejonge:
I agree with @stof, especially at this time it's good to be able to install it, even if it fails. This way bugs can be found and fixed. Also, I reckon you'd want to get the project up to speed with that version eventually anyway.

PR #657, opened 2015-03-11
Adopt PSR-2 coding standards

ciaranmcnulty:
* Added `CONTRIBUTING.md` specifying PSR-2
* Added `php_codesniffer` to travis
* Fixed all errors

This will make me less grumpy when people tell me my brackets are 'wrong'

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,25 @@
+Contributing
+============
+
+PhpSpec is an open source, community-driven project. If you'd like to contribute,
+feel free to do this, but remember to follow this few simple rules:
+
+Branching strategy
+-------------------
+
+- __Always__ base your changes on the `master` branch (all new development happens here),
+- When you create Pull Request, always select `master` branch as target, otherwise it
+will be closed (this is selected by default).
+
+Coverage
+--------
+
+- All classes that interact solely with the core logic should be covered by Specs
+- Any infrastructure adaptors should be covered by integration tests using PHPUnit
+- All features should be covered with .feature descriptions automated with Behat
+
+Code style / Formatting
+-----------------------
+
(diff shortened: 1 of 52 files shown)


GrahamCampbell:
Why not use https://styleci.io?

ciaranmcnulty:
@GrahamCampbell Looks good - I want a way to check locally myself before commit - is that possible?

stof:
@ciaranmcnulty stylecu.io just runs PHP-CS-Fixer

-1 for running the style checks on Travis itself. I prefer keeping the Travis build status for the status of tests only.
An option would be to add a failure condition on Scrutinizer to make it report a failed analysis when there are new CS issues.

ciaranmcnulty:
I'm confused why this is working locally for me and failing, BTW

I'll remove the CI elements and look at styleci.io or scrutinizer options, thanks guys

ciaranmcnulty:
OK I've taken out the CI element so the PR is just fixes

GrahamCampbell:
> -1 for running the style checks on Travis itself. I prefer keeping the Travis build status for the status of tests only.

Yep, agreed. You can use https://styleci.io/ to check pushed commits and php-cs-fixer locally.

GrahamCampbell:
If you set styleci to use the "psr2" preset, that's the same as using the psr2 level on php-cs-fixer.

GrahamCampbell:
We can't use .phpcs config files on styleci because for obvious reasons, we can't allow user provided php code to be executed on our workers.

ciaranmcnulty:
Can I exclude the `src` folder some other way? If not, we can't use style

GrahamCampbell:
We're working on that, but we'd accept a pull request to add that feature in the meanwhile.

PR #660, opened 2015-03-15
Adding PSR2 inspections to scrutinizer

ciaranmcnulty:


GrahamCampbell:
The trouble with scrutinizer is that it never actually fails the build for bad cs, so it's kind of pointless.

ciaranmcnulty:
@GrahamCampbell I've deliberately broken this build to show that it does in fact fail for bad CS if you ask it to

GrahamCampbell:
https://i.starbs.net/wL73 Looks like it's passing to me?

ciaranmcnulty:
I meant I broke it to test, I fixed again before merge.

Any issues labelled 'Coding Style' cause a build failure.

GrahamCampbell:
Nice.

GrahamCampbell:
Still, scrutinizer doesn't provide a diff for every fix, infact it doesn't for most fixes, unlike styleci, which provides a diff for everything.

ciaranmcnulty:
True. It does suppose php-ci-fixer but not sure if it patches.

If styleci supports excluded paths in future I'll consider a switch if you want to PR. At the moment we're running a scrutiniser build anyhow so it makes sense to use it.

GrahamCampbell:
Sure. I'll let you know when we ship that feature. The thing holding us is just none of us have enough free time, lol.

Issue #658, opened 2015-03-14
Extended diff for mock call with exact parameters


avant1:
Hello.

I think it would be nice if wrong mock call arguments (defined as exact values) could be outputed in a more detailed way.
For now, if exact value is string and it's length is more than 50 characters, only first 50 characters are displayed (even in verbose mode) for both expected and actual arguments (see https://github.com/phpspec/prophecy/blob/master/src/Prophecy/Util/StringUtil.php#L57)
And we get output like
```
  18  ! it can handle event
      method call:
        Double\Acme\Context\P2->set("data", "some really long string, containing more than 50 "
...)
      was not expected.
      Expected calls are:
        - set(exact("data"), exact("some really long string, containing more than 50 "...))
```
which is pretty useless if difference between strings is after 50 character. Even if difference is in first characters, it is not always easy to find it, cause strings are not aligned:
```
  18  it can handle event
      expected "some really long strin, c"..., but got "some really long string, "....

      @@ -1,1 +1,1 @@
      -some really long strin, containing more than 25 characters
      +some really long string, containing more than 25 characters
```

If I understand source code correctly, extended expected and actual arguments diff can be implemented as extension (via replacing default *formatter.presenter* by one which can detect `\Prophecy\Exception\Call\UnexpectedCallException`) and generating diffs for arguments.

I found no existing extensions which implement similar functionality. I could try to create one.
But I think it makes sence to be implemented in core. What do you think?

ciaranmcnulty:
:+1: It makes sense to me to improve the reporting in core. Do you want to try it?

avant1:
I'd love to.
Probably, I can find free time during next week.

PR #664, opened 2015-03-17
refs #658: show diff for unexpected calls arguments

avant1:
This pull request depends on recent phpspec/prophecy commit (https://github.com/phpspec/prophecy/commit/a874f56d601496b39002f16eb7ed01c64968119e).
Seems like there is no other way to get expected call arguments from ```UnexpectedCallException```.

Tests will fail until that commit will be released (i've manually added ```\Prophecy\Argument\ArgumentsWildcard::getTokens()``` method in my vendors dir, and tests passed).

### DIFF
--- a/features/formatter/developer_is_shown_diffs.feature
+++ b/features/formatter/developer_is_shown_diffs.feature
@@ -154,187 +154,187 @@ Feature: Developer is shown diffs
 
   Scenario: Unexpected call arguments string diffing
     Given the spec file "spec/Diffs/DiffExample4/ClassUnderSpecificationSpec.php" contains:
-    """
-    <?php
+      """
+      <?php
 
-    namespace spec\Diffs\DiffExample4;
+      namespace spec\Diffs\DiffExample4;
 
-    use PhpSpec\ObjectBehavior;
-    use Prophecy\Argument;
-    use Diffs\DiffExample4\ClassBeingMocked;
+      use PhpSpec\ObjectBehavior;
+      use Prophecy\Argument;
+      use Diffs\DiffExample4\ClassBeingMocked;
 
-    class ClassUnderSpecificationSpec extends ObjectBehavior
-    {
-        function it_can_do_work(ClassBeingMocked $objectBeingMocked)
-        {
-            $objectBeingMocked->setValue('some really really long string, and even more, and more!')->shouldBeCalled();
-            $this->doWork($objectBeingMocked);
-        }
-    }
-    """
+      class ClassUnderSpecificationSpec extends ObjectBehavior
+      {
+          function it_can_do_work(ClassBeingMocked $objectBeingMocked)
+          {
+              $objectBeingMocked->setValue('some really really long string, and even more, and more!')->shouldBeCalled();
+              $this->doWork($objectBeingMocked);
+          }
+      }
+      """
     And the class file "src/Diffs/DiffExample4/ClassUnderSpecification.php" contains:
-    """
-    <?php
+      """
+      <?php
 
-    namespace Diffs\DiffExample4;
+      namespace Diffs\DiffExample4;
 
-    class ClassUnderSpecification
-    {
-        public function doWork(ClassBeingMocked $objectBeingMocked)
-        {
-            $objectBeingMocked->setValue('some really really long string, and even more, and more');
-        }
-    }
-    """
+      class ClassUnderSpecification
+      {
+          public function doWork(ClassBeingMocked $objectBeingMocked)
+          {
+              $objectBeingMocked->setValue('some really really long string, and even more, and more');
+          }
+      }
+      """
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
You can change the composer dep to `dev-master`, and we won't merge until there's a valid tag (this is what we have done in #579)

avant1:
I've bumped phpspec/prophecy version up to dev-master, tests are green.
So, before this PR can be merged, I should set stable prophecy dependency version, am I right?

ciaranmcnulty:
@avant1 Prophecy has released 1.4.0 now - when I merge #579 it will conflict with your change to the prophecy version, you should be able to resolve it fairly simply.

avant1:
Done.

features/formatter/developer_is_shown_diffs.feature
```diff
@@ -150,3 +150,191 @@ Feature: Developer is shown diffs
             +    'i' => 2
             +    's' => 'bar'
       """
+
+
+  Scenario: Unexpected call arguments string diffing
+    Given the spec file "spec/Diffs/DiffExample4/ClassUnderSpecificationSpec.php" contains:
+    """
```
ciaranmcnulty: Can you indent the pystrings in all features please?

src/PhpSpec/Formatter/Presenter/StringPresenter.php
```diff
@@ -359,4 +366,75 @@ private function presentCallable($value)
 
         return sprintf('[%s()]', $value);
     }
+
+    /**
+     * @param UnexpectedCallException $exception
+     *
+     * @return string
+     */
+    private function presentCallArgumentsDifference(UnexpectedCallException $exception)
```
ciaranmcnulty: This method is very large and should be broken into several smaller ones

src/PhpSpec/Formatter/Presenter/StringPresenter.php
```diff
@@ -359,4 +366,75 @@ private function presentCallable($value)
...
+        $text = '';
+
+        $actualArguments = $exception->getArguments();
+        $methodProphecies = $exception->getObjectProphecy()->getMethodProphecies($exception->getMethodName());
+        $presentedMethodProphecy = $this->findMethodProphecyOfFirstNotExpectedCall($methodProphecies, $exception);
+        $expectedTokens = $presentedMethodProphecy->getArgumentsWildcard()->getTokens();
+
+        if (count($expectedTokens) !== count($actualArguments)) {
+
+            //parameters count mismatch will be shown in exception message, no need in diff
```
ciaranmcnulty: I'd rather not have inline comments, can you try to make it explicit from the code? Extracting the conditional to a method might help in this instance.

src/PhpSpec/Formatter/Presenter/StringPresenter.php
```diff
@@ -359,4 +366,75 @@ private function presentCallable($value)
...
+
+            //this call is expected
+            if (count($calls)) {
+                continue;
+            }
+
+            return $methodProphecy;
+        }
+
+        throw new \LogicException("UnexpectedCallException has been thrown, but no unexpected calls found.");
```
ciaranmcnulty: Is there a scenario where this is likely, or is it an extra guard? If the latter, it's ok to remove it.

src/PhpSpec/Formatter/Presenter/StringPresenter.php
```diff
@@ -359,4 +366,75 @@ private function presentCallable($value)
...
+                $expectedArguments[] = (string)$token;
+            }
+        }
+
+        foreach($actualArguments as $i => $actualArgument) {
+            $expectedArgument = $expectedArguments[$i];
+            if (is_null($actualArgument)) {
+                $actualArgument = 'null';
+            }
+            if (is_null($expectedArgument)) {
```
ciaranmcnulty: you could inline this to $expectedArgument ?: null in the method call below

ciaranmcnulty:
I've added some more comments.

2.2.0-BETA is now tagged but this could still make it into the release if you have time. I'd like 2.2.0 to be released in the next couple of weeks.

src/PhpSpec/Formatter/Presenter/StringPresenter.php
avant1: All arguments that are equal to false (empty string/false/numeric zero) will be outputed as 'null' if I do so. I will use ternary operator instead of `if` here:
```php
$actualArgument = is_null($actualArgument) ? 'null' : $actualArgument;
```

avant1:
@ciaranmcnulty Thanks for reviewing. I've made appropriate changes.

ciaranmcnulty:
Thanks for the contribution! This will be in the next release (either  BETA2 or RC1)

PR #667, opened 2015-03-26
Made autoload path check a single if statement.

RyanHavoc:
I'm not sure why there were two if statements here?

In my case the first if statement found the autoload file and the second one didn't, which lead to it throwing the standard error and not running phpspec.

This change to the file fixed the issue for me so wanted to put it forward as a pull request unless theres an non-obvious reason why it's setup like this?

### DIFF
--- a/bin/phpspec
+++ b/bin/phpspec
@@ -7,9 +7,7 @@ if (is_file($autoload = getcwd() . '/vendor/autoload.php')) {
     require $autoload;
 } elseif (is_file($autoload = getcwd() . '/../../autoload.php')) {
     require $autoload;
-}
-
-if (is_file($autoload = __DIR__ . '/../vendor/autoload.php')) {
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
If phpspec is compiled as a phar, or installed globally via composer, we need to load both the project's autoloader (the first `if`), and PhpSpec's own autoloader (the second `if`)

ciaranmcnulty:
Appreciate the contribution though!

RyanHavoc:
Ah. That makes sense then. I'm using Laravel so I imagine it works with just the first if for me because phpspec is part of the Laravel autoload.

The question is then, under those circumstances how do I stop it choking on the second if?

ciaranmcnulty:
Choking? What error are you having? Maybe open as an issue

RyanHavoc:
Sure. Explained here: https://github.com/phpspec/phpspec/issues/668

PR #676, opened 2015-03-31
Fix unexpected method call exception presenting

avant1:
There is a serious problem with PR #664.
When there is no method prophecy for called method, fatal error is generated (Call to a member function on a non-object).

This PR fixes problem and adds feature coverage.

### DIFF
--- a/features/formatter/developer_is_shown_diffs.feature
+++ b/features/formatter/developer_is_shown_diffs.feature
@@ -152,7 +152,7 @@ Feature: Developer is shown diffs
       """
 
 
-  Scenario: Unexpected call arguments string diffing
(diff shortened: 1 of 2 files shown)


features/formatter/developer_is_shown_diffs.feature
```diff
@@ -208,7 +208,7 @@ Feature: Developer is shown diffs
       """
 
 
-  Scenario: Unexpected call arguments array diffing
+  Scenario: Unexpected method argumets call arguments array diffing
```
ciaranmcnulty: 'arguments' typo here + in others

ciaranmcnulty:
:+1:

PR #679, opened 2015-04-03
Remove sensitivity to time warnings in formatter output

ciaranmcnulty:
This has caused intermittent failure


features/bootstrap/Matcher/ApplicationOutputMatcher.php
```diff
@@ -42,6 +43,13 @@ public function positiveMatch($name, $subject, array $arguments)
         }
     }
 
+    private function normalise($string)
```
stof: US english, like in the codebase ?

ciaranmcnulty: Argh

PR #682, opened 2015-04-07
Add Caller extension point

shanethehat:
The reason behind this change is that some frameworks make use of the magic accessor methods, and so define them in a way that might be undesirable for use by phpspec. Magento is an example of this, where the base object class will attempt to bootstrap the entire application when a call to a magic accessor is made.

Currently Magespec works around this problem by implementing it's own version of Caller: https://github.com/MageTest/MageSpec/blob/master/src/MageTest/PhpSpec/MagentoExtension/Wrapper/Subject/VarienCaller.php. All of this exists so that the two methods named above can be overridden, and it could be made much simpler with much less duplication if the responsibility of these methods is delegated to another object that can be replaced..

### DIFF
--- a/src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
+++ b/src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Phpspec\CodeAnalysis;
+
+interface AccessInspectorInterface
+{
+    /**
+     * @param object $object
+     * @param string $property
+     * @param bool $withValue
+     * @return bool
+     */
+    public function isPropertyAccessible($object, $property, $withValue);
+
+    /**
+     * @param object $object
+     * @param string $method
+     * @return bool
+     */
+    public function isMethodAccessible($object, $method);
+}
--- a/src/PhpSpec/CodeAnalysis/MagicAwareAccessInspector.php
+++ b/src/PhpSpec/CodeAnalysis/MagicAwareAccessInspector.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace PhpSpec\CodeAnalysis;
+
+use ReflectionMethod;
+use ReflectionProperty;
+
+final class MagicAwareAccessInspector implements AccessInspectorInterface
+{
+    /**
+     * @param object $object
+     * @param string $property
+     * @param bool $withValue
+     * @return bool
+     */
+    public function isPropertyAccessible($object, $property, $withValue = false)
+    {
+        if (!is_object($object)) {
+            return false;
+        }
+
+        if (method_exists($object, $withValue ? '__set' : '__get')) {
+            return true;
+        }
+
+        if (!property_exists($object, $property)) {
+            return false;
+        }
+
+        $propertyReflection = new ReflectionProperty($object, $property);
+        return $propertyReflection->isPublic();
+    }
+
+    /**
+     * @param object $object
+     * @param string $method
+     * @return bool
+     */
+    public function isMethodAccessible($object, $method)
+    {
+        if (!is_object($object)) {
+            return false;
+        }
+
+        if (method_exists($object, '__call')) {
+            return true;
+        }
+
+        if (!method_exists($object, $method)) {
+            return false;
+        }
+
+        $methodReflection = new ReflectionMethod($object, $method);
+        return $methodReflection->isPublic();
+    }
+}
--- a/src/PhpSpec/Wrapper/Subject/Caller.php
+++ b/src/PhpSpec/Wrapper/Subject/Caller.php
@@ -13,6 +13,8 @@
 
 namespace PhpSpec\Wrapper\Subject;
 
+use PhpSpec\CodeAnalysis\MagicAwareAccessInspector;
+use Phpspec\CodeAnalysis\AccessInspectorInterface;
 use PhpSpec\Exception\ExceptionFactory;
 use PhpSpec\Loader\Node\ExampleNode;
 use PhpSpec\Wrapper\Subject;
@@ -21,8 +23,6 @@ use PhpSpec\Wrapper\Unwrapper;
 use PhpSpec\Event\MethodCallEvent;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface as Dispatcher;
 use ReflectionClass;
-use ReflectionMethod;
-use ReflectionProperty;
 use ReflectionException;
 
 class Caller
@@ -47,26 +47,33 @@ class Caller
      * @var \PhpSpec\Exception\ExceptionFactory
      */
     private $exceptionFactory;
+    /**
+     * @var AccessInspectorInterface
+     */
+    private $accessInspector;
 
     /**
-     * @param WrappedObject    $wrappedObject
-     * @param ExampleNode      $example
-     * @param Dispatcher       $dispatcher
+     * @param WrappedObject $wrappedObject
+     * @param ExampleNode $example
+     * @param Dispatcher $dispatcher
      * @param ExceptionFactory $exceptions
-     * @param Wrapper          $wrapper
+     * @param Wrapper $wrapper
+     * @param AccessInspectorInterface $accessInspector
      */
     public function __construct(
         WrappedObject $wrappedObject,
         ExampleNode $example,
         Dispatcher $dispatcher,
         ExceptionFactory $exceptions,
-        Wrapper $wrapper
+        Wrapper $wrapper,
+        AccessInspectorInterface $accessInspector = null
     ) {
         $this->wrappedObject    = $wrappedObject;
         $this->example          = $example;
         $this->dispatcher       = $dispatcher;
         $this->wrapper          = $wrapper;
         $this->exceptionFactory = $exceptions;
+        $this->accessInspector  = $accessInspector ?: new MagicAwareAccessInspector();
     }
 
     /**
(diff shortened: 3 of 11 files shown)


ciaranmcnulty:
Rather than encourage inheritance, can we shift this to composition perhaps?

An object that determines accessibility would make sense to me.

src/PhpSpec/CodeAnalysis/AccessInspector.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace PhpSpec\CodeAnalysis;
+
+use ReflectionMethod;
+use ReflectionProperty;
+
+class AccessInspector implements AccessInspectorInterface
```
ciaranmcnulty: Naming: What makes this concrete implementation specific?

ciaranmcnulty: You can probably make this class final

src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+namespace Phpspec\CodeAnalysis;
+
+interface AccessInspectorInterface
```
ciaranmcnulty: I'm tempted to say we should move towards not using `*Interface`, although that would introduce inconsistency

src/PhpSpec/CodeAnalysis/AccessInspector.php
shanethehat: It's hard to answer what makes this specific without reflecting on the intention behind the PR, which I'm trying not to do. maybe DetailedAccessInspector, MagicAccessInspector, GenericAccessInspector?

src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
shanethehat: I did consciously decide to follow the existing naming convention, however if you want to start moving towards not using `*Interface` for a v3 release, I'm happy to start now.

ciaranmcnulty:
Sorry @shanethehat I left you hanging on this one - looks like there are some merge problems?

src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
ciaranmcnulty: I think let's not drop `*Interface` yet for consistency, but try and make a conscious effort to name the concrete implementations uniquely.

shanethehat:
@ciaranmcnulty I don't think the issue is mergability. The build is broken because the changes to the signature of Wrapper are incompatible with phpspec2-expect. I've fixed this locally by making the change to expect as well, but I haven't raised a PR for it yet. If you approve this change in principle then I'll go ahead with an accompanying change to expect.

Last time we talked about it I ended up investigating a deeper refactoring of the whole wrapper/subject area, but that didn't really get very far.

src/PhpSpec/CodeAnalysis/AccessInspector.php
ciaranmcnulty: What will make the Mage one different, maybe that'll give an idea what makes this version unique

ciaranmcnulty:
Is it just a case that the dependency isn't injected in phpspec2-expect?

shanethehat:
Yes, in Wrapper if I remember correctly

src/PhpSpec/CodeAnalysis/AccessInspector.php
shanethehat: The Mage Caller doesn't allow magic __get, __set or __call methods to be used for determining if properties and methods are accessible.

ciaranmcnulty:
In terms of BC it might make sense to make that an optional dependency for Caller

shanethehat:
I'm not sure if that helps to be honest, because if it's not provided Caller wouldn't behave like it does now. We could keep the current implementation in Caller if the inspector isn't provided I suppose, and mark the methods as deprecated?

ciaranmcnulty:
Caller could internally instantiate the default object if one isn't provided.

This is crap but it provides a BC layer

src/PhpSpec/Wrapper/Subject/Caller.php
```diff
@@ -47,26 +48,33 @@ class Caller
...
-        Wrapper $wrapper
+        Wrapper $wrapper,
+        AccessInspectorInterface $accessInspector
     ) {
         $this->wrappedObject    = $wrappedObject;
         $this->example          = $example;
         $this->dispatcher       = $dispatcher;
         $this->wrapper          = $wrapper;
         $this->exceptionFactory = $exceptions;
+        $this->accessInspector  = $accessInspector;
```
ciaranmcnulty: this could be:

```php
$this->accessInspector  = $accessInspector :? new AccessInspector();
```

ciaranmcnulty:
It was fine to break BC in the constructor of caller when I thought there were no extensions using it ;-) 

Now there is one we have a few options:

1. Make argument optional and instantiate internally (see above)
2. Do something similar via an adaptor
3. Think of some other way of adding the extension point

So I'm leaning towards 1

src/PhpSpec/CodeAnalysis/AccessInspector.php
ciaranmcnulty: Maybe it's fine as it is

ciaranmcnulty: Also this can be `final`

shanethehat:
@ciaranmcnulty done!

src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
stof: IMO, the implementation should not be named ``AccessInspector``. Otherwise it is a pain to change the convention later. We should already name the class as if the name ``AccessInspector`` was reserved by the interface itself

shanethehat: `MagicAwareAccessInspector`?

ciaranmcnulty: Maybe? It's not great but I don't have a superior suggestion

ciaranmcnulty:
Great, I see the build now passes too. I think it's just naming stuff that's stopping me merging.

src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
shanethehat: I don't really like it because it implies the existence of an access inspector that is not magic aware, and that's only going to exist in the magespec repo.

Sometimes if you really can't name something, then the implementation might not be right. Maybe we need several Inspectors for the specific tasks that the current one is doing, like `MagicAccessInspector`, `ReflectionAccessInspector`, `GenericAccessInspector`. They can be iterated over until one returns true, and then all Magespec has to do is remove the one for magic methods from the list.

shanethehat: That's going to make the BC supporting hack we put in a bit more unpleasant though.

ciaranmcnulty: That's something we can refactor to later, now the interface is there.

`MagicAwareAccessInspector` makes sense to me, you sold me on it.

shanethehat: Fair enough, then I'll rename it

ciaranmcnulty:
So is this PR enough for you to make the improvements you want in `MageSpec`? Is it worth making a `MageSpec` branch that uses this branch of `PhpSpec` to verify that?

src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
ciaranmcnulty: Wait I see what you're saying.

We could have `MagicAccessInspector` decorate `VisibilityAccessInspector`. That way MageSpec just needs to mess with the container config and remove that decorator.

ciaranmcnulty: (plus it simplifies the logic)

shanethehat: Yeah, that's where I was going, but it does mean that in order to support BC that hack in the Caller constructor would have to do the same decoration stuff. I think it's something to aim for in refactoring when you can make a BC break.

The existing solution will work well enough for MageSpec, it can just replace the MagicAwareAccessInspector for a MagicUnawareAccessInspector. Probably not named quite like that though.

ciaranmcnulty: True. Man, BC is annoying.

ciaranmcnulty: I think after this is in a stable release we can modify phpspec2-expects and then do a survey of other extensions to look at a controlled BC break on the Caller constructor

shanethehat:
Renamed. This will be fine for Magespec to just replace the implementation of the AccessInspector with something a bit more Magento friendly.

src/PhpSpec/CodeAnalysis/AccessInspectorInterface.php
```diff
@@ -0,0 +1,21 @@
...
+{
+    /**
+     * @param object $object
+     * @param string $property
+     * @param bool $withValue
+     * @return bool
+     */
+    public function isPropertyAccessible($object, $property, $withValue);
```
ciaranmcnulty: Now I look at this again... this method is overloaded.

What we're interested in is whether a property is `writable` or `readable`; these should be separate methods (which would probably have some shared functionality internally).

Similarly the other method should perhaps reflect that we care it's `callable`

src/PhpSpec/CodeAnalysis/MagicAwareAccessInspector.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * @param object $object
+     * @param string $property
+     * @param bool $withValue
+     * @return bool
+     */
+    public function isPropertyAccessible($object, $property, $withValue = false)
+    {
+        if (!is_object($object)) {
```
ciaranmcnulty: The fact it might not be an object is counterintuitive because the parameter is typehinted as an object, and stems from a bad design decision inside `Caller`

Let's keep the `is_object` check localised to `Caller` and not let it spill over to this new class.

ciaranmcnulty:
Thanks @shanethehat! This'll be in 2.3.0 in a few(?) weeks so you can plan your next MageSpec appropriately.

Issue #681, opened 2015-04-06
Documentation: Are paths relative to where the config file is located?


mahalay:
According to this documentation as of this writing:

> You can use a different config file name and path with the --config option:

http://web.archive.org/web/20150406135047/http://www.phpspec.net/en/latest/cookbook/configuration.html

At least for "src_path" it says:
> Paths are relative to the location of the config file.

Suppose in composer.json I setup "Foo" namespace (psr-4) to be autoloaded (because I'd like to move the folder "Foo" around (e.g., share it)) from:
> "\<root_app_dir\>/Foo/src"

And thus have my phpspec config inside it:
> "\<root_app_dir\>/Foo/phpspec.yml

Since the default value of "src_path" is "src" - relative to the config, assuming I described a class as "Foo\Bar\Baz", when I run the following:
> \<root_app_dir\> $ bin/phpspec run --config Foo/phpspec.yml

I would then expect that the generated classes will be stored in "Foo/src/Bar/Baz.php". However this is not the case. Generated classes will be generated under "\<root_app_dir\>/src/Bar/Baz". I get an infinite loop prompting me if I want to generate the class "Foo\Bar\Baz" and if I want to overwrite "Baz.php" file because it already exists.

To get around this problem, I need to have the following config:
```
suites:
    foo_suite:
        namespace: Foo
        src_path: Foo/src
        psr4_prefix: Foo
        spec_path: Foo
```
If I am not missing something, this then proves that the documentation is wrong. In fact, based on my observation, paths defined in the config is relative to which folder phpspec got executed.

I believe this becomes a limiting factor for phpspec, because one would need to know beforehand which folder to execute phpspec from - in this example, in the "\<root_app_dir\>". You can't even run it under "\<root_app_dir\>/Foo". i.e.,:
```
<root_app_dir>/Foo $ ../bin/phspec run
```

Thanks for this great tool

PR #683, opened 2015-04-09
Fixed issue #681 - src|spec  paths are now relative to the config file

mahalay:
Doc says src_path is relative to the config - unfortunately not. Fixed src_path so its relative to the config, I've also fixed spec_path, convinced that it should also be relative to the config.

With this fix, regardless where phpspec was run, src_path and spec_path is relative to the config


mahalay:
crap! how embarassing.

PR #684, opened 2015-04-09
Fixed issue #681 - src|spec paths are made relative to the config file

mahalay:
Doc says src_path is relative to the config - unfortunately not. Fixed src_path so its relative to the config, I've also fixed spec_path, convinced that it should also be relative to the config.

With this fix, regardless where phpspec was run, src_path and spec_path is relative to the config

### DIFF
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -163,12 +163,42 @@ class Application extends BaseApplication
             if (!file_exists($customPath)) {
                 throw new RuntimeException('Custom configuration file not found at '.$customPath);
             }
+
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Can you fix the coding style issue please?

Am I right in thinking this breaks backwards compatibility for some setups? If it does, it's not something I feel we can change before 3.0 (we could change the documentation to reflect reality of course).

I'd like to see some feature coverage around config files, is that something you think you could add?

mahalay:
I'm calling it a night... Yup I agree that a feature coverage would be better. Closing it for now.

PR #685, opened 2015-04-11
Added use_config_path suite config option (Feature)

mahalay:
The use_config_path allows a dev to specify if src or spec paths are relative to the phpspec.yml or the current directory where phpspec was invoked. 

This effectively resolves issue #681 

By default it is set to false - phpspec will behave like it was before (i.e., paths re relative to directory where phpspec was invoked)

When set to true, paths will be relative to the config file used. Very useful when used in conjunction with "--config" option.

See "features/config/paths_can_be_relative_to_config.feature" for further details

### DIFF
--- a/features/config/paths_can_be_relative_to_config.feature
+++ b/features/config/paths_can_be_relative_to_config.feature
@@ -1,0 +1,76 @@
+Feature: Paths can be relative to config or current working directory
+    As a Developer
+    I want to be able to specify if paths are relative to the directory of the config or current working directory
+    So I may specify a custom location for the config file
+
+    Scenario: Generating a class using a custom config with path being relative to the config directory
+        Given the config file located in "Awesome" contains:
+            """
+            suites:
+              behat_suite:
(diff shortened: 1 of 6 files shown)


features/config/paths_can_be_relative_to_config.feature
```diff
@@ -0,0 +1,76 @@
+Feature: Paths can be relative to config or current working directory
```
ciaranmcnulty: This probably needs a scenario where the new option is omitted

ciaranmcnulty:
Why close this?

mahalay:
Because I don't know how to make changes to a pull request :P

GrahamCampbell:
@mahalay Just push to your branch, and they'll show up here. :)

ciaranmcnulty:
Ah :-) Just push future commits to your same branch and they will be part of the PR

GrahamCampbell:
@ciaranmcnulty 1 second difference. :)

PR #686, opened 2015-04-12
Added use_config_path suite config option (Feature) - fixes issue #681

mahalay:
I'm too lazy to write...

See conversation in https://github.com/phpspec/phpspec/pull/685

### DIFF
--- a/features/bootstrap/ApplicationContext.php
+++ b/features/bootstrap/ApplicationContext.php
@@ -86,6 +86,21 @@ class ApplicationContext implements Context, MatchersProviderInterface
     }
 
     /**
+     * @Given I have started describing the :class class with the :config (custom) config
+     * @Given I start describing the :class class with the :config (custom) config
+     */
+    public function iDescribeTheClassWithTheConfig($class, $config)
+    {
+        $arguments = array(
+            'command' => 'describe',
+            'class' => $class,
+            '--config' => $config
+        );
+
+        expect($this->tester->run($arguments, array('interactive' => false)))->toBe(0);
+    }
+
+    /**
      * @When I run phpspec (non interactively)
      * @When I run phpspec using the :formatter format
      * @When I run phpspec with the :option option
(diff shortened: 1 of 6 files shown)


mahalay:
Sorry guys... Newbie here. My first time to contribute to an open source project. Just realized that this is fun to do.

And thanks for the tip.

ciaranmcnulty:
I've been thinking about this change, my main hesitation has been that this adds another configuration flag for us to manage and document.

I don't see particularly what problem it solves, where you have:

```
# src/Awesome/phpspec.yml;
suites:
  some_suite:
    namespace: MilkyWay\OrionCygnusArm
    use_config_path: true
```

You can have:

```
# src/Awesome/phpspec.yml;
suites:
  some_suite:
    namespace: MilkyWay\OrionCygnusArm
    src_path: src/Awesome
```

I think that has the same result?

I'm really loathe to close this one because you said it's one of your first Open Source contributions, but suite config is something that's overcomplex right now and adding more stuff is not the direction I want to take it in.

mahalay:
The problem this solves is: ability to run the tests regardless of the current working directory. Wouldn't it be cool if you can share your codes (modules) without requiring your users to run the tests from a specific directory?

ciaranmcnulty:
Right,I can definitely see the use case.

Perhaps we can make the path relativity explicit by having parameters that can be dropped in, the way Behat does:

```yml
suites:
  some_suite:
    namespace: MilkyWay\OrionCygnusArm
    src_path: %config_dir%/Awesome
```

This would avoid adding another key

ciaranmcnulty:
we could have

```
%config_dir%
%bin_dir%
%working_dir%
```

mahalay:
@ciaranmcnulty, you know what, I knew that'd be the better approach, I actually started working on it. but... 

let's agree on using the following:
```
%paths.config%
%paths.bin%
%paths.base%
```

I got the inspiration from Behat, it uses %paths.base% for the "current working dir". I don't have problems using the ones you suggested, this is in fact trivial, I'm just assuming that phpspec users are likely Behat users so using that convention they'll feel at home. What  do you think?

ciaranmcnulty:
Using the same as Behat makes perfect sense :+1:

mahalay:
One last clarification... 

bin_dir being the one defined in composer.json?

ciaranmcnulty:
I think `%paths.bin%` is 'the location of the `phpspec` binary'

PR #688, opened 2015-04-13
close #687 empty prefix is allowed

### LINKED ISSUE #687: The class creation adds a backslash on empty spec prefix
When the prefix is set as empty the spec class creation uses a backslash prefixed namespace.

Our use case is the following, we do not want to split the namespaces for the specs and the source code, this way we simplify a lot of imports, we have separated the specs and source code by folders, we do not need to separate by namespaces, the specs are already suffixed on the classes, so we do not need to prefix by namespace too.

Markcial:
With this PR the spec prefix can be empty, this way the namespace does not start with a backslash, which is a not valid class name.

### DIFF
--- a/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
+++ b/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
@@ -58,7 +58,6 @@ class PSR0LocatorSpec extends ObjectBehavior
         $this->beConstructedWith('Cust\Ns', '', dirname(__DIR__), __DIR__);
 
         $this->getSpecNamespace()->shouldReturn('Cust\Ns\\');
-        $this->getSpecNamespace()->shouldNotReturn('\Cust\Ns\\');
 
         $this->getFullSpecPath()->shouldReturn(__DIR__.DIRECTORY_SEPARATOR.'Cust'.DIRECTORY_SEPARATOR.'Ns'.DIRECTORY_SEPARATOR);
     }
--- a/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
+++ b/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
@@ -53,6 +53,16 @@ class PSR0LocatorSpec extends ObjectBehavior
         $this->getFullSrcPath()->shouldReturn(dirname(__DIR__).DIRECTORY_SEPARATOR);
     }
 
+    function it_should_not_have_backslash_on_missing_prefix()
+    {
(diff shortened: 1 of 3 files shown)


spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
```diff
@@ -53,6 +53,16 @@ function it_generates_proper_fullSrcPath_even_from_empty_namespace()
         $this->getFullSrcPath()->shouldReturn(dirname(__DIR__).DIRECTORY_SEPARATOR);
     }
 
+    function it_should_not_have_backslash_on_missing_prefix()
+    {
+        $this->beConstructedWith('Cust\Ns', '', dirname(__DIR__), __DIR__);
+
+        $this->getSpecNamespace()->shouldReturn('Cust\Ns\\');
+        $this->getSpecNamespace()->shouldNotReturn('\Cust\Ns\\');
```
ciaranmcnulty: If it returns the other value you don't need to specify a value it doesn't return

spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
```diff
@@ -53,6 +53,16 @@ function it_generates_proper_fullSrcPath_even_from_empty_namespace()
...
+
+        $this->getFullSpecPath()->shouldReturn(__DIR__.DIRECTORY_SEPARATOR.'Cust'.DIRECTORY_SEPARATOR.'Ns'.DIRECTORY_SEPARATOR);
```
ciaranmcnulty: This should perhaps be a different example

ciaranmcnulty:
This maybe could do with a `.feature` illustrating the impact this has on user behaviour

Markcial:
Just a moment please, i will update the PR with a feature instead of a spec test

ciaranmcnulty:
No keep the spec too ;-)

ciaranmcnulty:
Thanks for the feature, it is a lot clearer to me now what this PR is about :+1:

Markcial:
Oh sorry, i will add again the spec then. Thanks to you (the phpspec folks), the gherkin syntax is very cool and easy to use, at least the way that is used in this project :D

acasademont:
This would come handy! +1

ciaranmcnulty:
Thanks for the contributution!

Is there anything that needs to go in the docs for this?

Markcial:
only a slight mention about the new feature of shared namespace between specs and classes might be fine. Thanks for the merge :)

PR #694, opened 2015-04-28
Fix haveKeyWithValue matcher for arrays with correct key, but wrong values

felixkiss:
There was a bug, that made tests using the haveKeyWithValue matcher with arrays
pass, if the array had the specified key, but the wrong value.

e.g. before this patch:
```php
// implementation
public function getAttributes()
{
    return ['name' => 'Some Name'];
}

// should be a failing test, but is actually passing
function it_returns_the_right_name()
{
    $this->getAttributes()->shouldHaveKeyWithValue('name', 'Foo Bar');
}
```

Feature originally introduced in #640 

Let me know if you want a Behat feature as well. I don't know what the policy on this project is about what code paths get acceptance tests (the happy path is already covered by an acceptance test).


felixkiss:
Functionality did already work with `ArrayObject`s. Only `array`s were affected.

ciaranmcnulty:
Thanks for the fix - the level of coverage looks fine to me

PR #696, opened 2015-05-01
docs: bootstrapping: fixed the config key

MacDada:
`console.io.bootstrap: path/to/different-bootstrap.php` does not work (does not load the bootstrap file).

`bootstrap: path/to/different-bootstrap.php` does work.

phpspec version 2.2.0


ciaranmcnulty:
Thanks!

PR #697, opened 2015-05-05
Added `public` to the speced method to keep consistency

gsdevme:
Seems good practice to explicitly define the scope as you would in your application.

### DIFF
--- a/features/code_generation/developer_generates_spec.feature
+++ b/features/code_generation/developer_generates_spec.feature
@@ -16,7 +16,7 @@ Feature: Developer generates a spec
 
       class MarkdownSpec extends ObjectBehavior
       {
-          function it_is_initializable()
+          public function it_is_initializable()
           {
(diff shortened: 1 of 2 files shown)


avant1:
Reasons to omit `public` keyword can be found here: https://github.com/phpspec/phpspec/pull/590#issuecomment-67995169

gstaniforthsessiondigital:
So if you have public properties in your specs you use `var` because thats more expressive than `public/private/protected` in the sense of meaning `variable`...

Seems unrealistic to wave the magic hand and say they are not code... because they are, I don't see why we can't have consistency and write code the same way. 

Is there a roadmap which replaces the specs with an expressive whitespace parsed language perhaps? :trollface:

jakzal:
Closing for good reasons Marcello already gave in https://github.com/phpspec/phpspec/pull/590#issuecomment-67995169.

PR #699, opened 2015-05-07
Add new duringInstantiation() method for named constructs closes #610 

### LINKED ISSUE #610: Expect exceptions from static factory methods using beConstructedThrough()
I'm using static factory methods to construct objects.

Is there any way to expect exceptions that are thrown from the factories?

```php
class User {
    private $email;

    private function __construct() {
    }

    public static function createWithEmail($email) {
        $user = new User();
        if(!preg_match("/.+\@.+\..+/", $email)) {
            throw new InvalidArgumentException("Invalid email address");
        }
        $user->email = $email;
        return $user;
    }
}
```

```php
class UserSpec extends ObjectBehavior {
    const USER_EMAIL = "jason@gmail.com";
    const USER_INVALID_EMAIL = "notanemail";

    function it_can_be_created_with_an_email_address() {
        $this->beConstructedThrough('createWithEmail', [self::USER_EMAIL]);
        $this->shouldHaveType('User');
        $this->email()->shouldBe(self::USER_EMAIL);
    }

    /* What should this be?? */
    function it_should_reject_invalid_email_addresses() {
        $this->shouldThrow('InvalidArgumentException')
            ->beConstructedThrough('createWithEmail', [self::USER_INVALID_EMAIL]);
    }
}
```

Any way to do this? Or is there a better way that I should design my factories?

acasademont:
As said in #610 this implements the a new `duringInstantiation()` method to allow checking for exceptions thrown during the construction of an object.

### DIFF
--- a/features/matchers/developer_uses_throw_matcher.feature
+++ b/features/matchers/developer_uses_throw_matcher.feature
@@ -163,3 +163,53 @@ Feature: Developer uses throw matcher
     """
     When I run phpspec
     Then the suite should pass
+
+  @issue610
+  Scenario: Throwing an exception during object construction
+    Given the spec file "spec/Runner/ThrowExample5/MarkdownSpec.php" contains:
+      """
+      <?php
+
+      namespace spec\Runner\ThrowExample5;
+
+      use PhpSpec\ObjectBehavior;
+      use Prophecy\Argument;
+
+      class MarkdownSpec extends ObjectBehavior
+      {
+          function it_throws_an_exception_using_during_instantiation_syntax()
+          {
+              $this->beConstructedWith(1, 2);
+              $this->shouldThrow('\InvalidArgumentException')->duringInstantiation();
+          }
+
+          function it_throws_an_exception_using_during_named_instantiation_syntax()
+          {
+              $this->beConstructedThrough('defaultNumber2', array(1));
+              $this->shouldThrow('\InvalidArgumentException')->duringInstantiation();
+          }
+      }
+
+      """
+    And the class file "src/Runner/ThrowExample5/Markdown.php" contains:
+      """
+      <?php
+
+      namespace Runner\ThrowExample5;
+
+      class Markdown
+      {
+          public function __construct($num1, $num2)
+          {
+              throw new \InvalidArgumentException();
+          }
+
+          public static function defaultNumber2($num1, $num2 = 2)
+          {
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
Thanks for this, it looks good!

I'm intending to release a patch to fix the problems with arrayKeyHasValue (see #702) which is the only reason this hasn't been looked at yet

acasademont:
great!

features/matchers/developer_uses_throw_matcher.feature
```diff
@@ -163,3 +163,51 @@ Feature: Developer uses throw matcher
     """
     When I run phpspec
     Then the suite should pass
+
+  @issue610
+  Scenario: Throwing an exception during object construction
+    Given the spec file "spec/Runner/ThrowExample5/MarkdownSpec.php" contains:
+    """
```
ciaranmcnulty: Can you indent the PyStrings a couple of characters please?

ciaranmcnulty:
I like the overall idea, I'm not quite on board with the syntax yet.

Rather than complicating the arguments to `duringInstantiation`, it might be better to make it argumentless and expect the spec author to indicate how the object will be instantiated:

```php
$this->beConstructedWith(1,2,3);
$this->shouldThrow('InvalidArgumentException')->duringInstantiation();
```

For me this is clearer what's going on, and removes the complexity of having `beConstructedWith` with different arguments to the `duringInstantiation`. What do you think?

features/matchers/developer_uses_throw_matcher.feature
acasademont: sure!

acasademont:
Yes, much cleaner, i'll try to do it ASAP!

acasademont:
@ciaranmcnulty done!

acasademont:
btw, if you're ok with the syntax and the code changes i'll update the docs too to reflect the new method in the `Throw` matcher :)

ciaranmcnulty:
Looks good to me; doc changes in the same PR make sense.

I won't merge for a while in case there is any fallout from `2.2.1` but we can get ready for `2.3.0` pretty soon

acasademont:
Updated docs, feel free to merge now :)

ciaranmcnulty:
Great work @acasademont, thanks for the contribution!

acasademont:
:+1: welcome, and thank you too!

PR #702, opened 2015-05-11
Fixed exception thrown by ArrayKeyValueMatcher

gquemener:
I see several problems with the current implementation of the ArrayKeyValueMatcher:

- `$this->method()->shouldHaveKeyWithValue('foo', 'bar')` will throw exception telling that no matcher was found for "haveKeyWithValue", even if the matcher is registered. This is because the matcher is only used when the method returns an array or an ArrayObject. This is not convenient at all, as I expect my method body to be empty (and thus return `null`) when I'm writing my spec. PhpSpec should tell me that expecting result was an array or instance or ArrayObject and that he received something else.
- When my method returns an array that doesn't contain the expected key, PhpSpec will raise a notice error: `Undefined index on line 79`. BTW, this error was no captured earlier because of the use of `shouldThrow()` without argument. It indeed throws an exception but not the good one! 

This PR covers these lacks.

Another concern of mine is what to expect when I do `$this->method()->shouldNotHaveKeyWithValue('abc', 123)` and `$this->method()` returns `[]`? Should it fail because the method didn't return an array with the expected key? Was the key expected at all (as we are asserting the content of the key, I'd say yes, we expect the value of the key not to be `123`, but we at least expect the key to exist, don't you think?)?

### DIFF
--- a/spec/PhpSpec/Matcher/ArrayKeyValueMatcherSpec.php
+++ b/spec/PhpSpec/Matcher/ArrayKeyValueMatcherSpec.php
@@ -39,29 +39,9 @@ class ArrayKeyValueMatcherSpec extends ObjectBehavior
         $this->supports('haveKeyWithValue', array(), array('', ''))->shouldReturn(true);
     }
 
-    function it_does_not_matches_null_value()
+    function it_does_not_respond_to_haveKeyWithValue_with_non_array_subject()
     {
-        $this->shouldThrow(new FailureException('Expected array or instance of \ArrayAccess, got NULL'))->duringPositiveMatch('haveKeyWithValue', null, array('abc', 123));
-    }
-
-    function it_does_not_matches_string_value()
-    {
-        $this->shouldThrow(new FailureException('Expected array or instance of \ArrayAccess, got string'))->duringPositiveMatch('haveKeyWithValue', 'foo', array('abc', 123));
-    }
-
-    function it_does_not_matches_integer_value()
-    {
-        $this->shouldThrow(new FailureException('Expected array or instance of \ArrayAccess, got integer'))->duringPositiveMatch('haveKeyWithValue', 42, array('abc', 123));
-    }
-
-    function it_does_not_matches_double_value()
-    {
-        $this->shouldThrow(new FailureException('Expected array or instance of \ArrayAccess, got double'))->duringPositiveMatch('haveKeyWithValue', 1.2, array('abc', 123));
-    }
-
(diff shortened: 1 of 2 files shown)


gquemener:
Ping @echochamber @felixkiss I think you might have an opinion on this one.

gquemener:
Ping @ciaranmcnulty 
Hello, I think you might be interested in this one too :)

felixkiss:
Looks good. More detailed error messages are always an improvement.

As for `$this->method()->shouldNotHaveKeyWithValue('abc', 123)`: I would expect it to pass with `[]` based on the wording of the matcher.

ciaranmcnulty:
The logic looks sound, and I can't see any coding style problems to mention :+1: 

I'm not sure I agree with your changes to `supports`. It's a different strategy than we've used in other matchers and it may be better to change the way we report these errors to users, rather than making matchers have to do this level of type checking and reporting.

What I mean is we could include type information in the message: "No matcher shouldHaveKeyWithValue() found for value null" or something similar.

stof:
The issue about supports could happen with any matcher out there though (for instance, ``shouldHaveCount`` can trigger different behavior depending on whether you are countable or no, and whether you have a method ``hasCount`` or no)

ciaranmcnulty:
If matchers are too open about what types they can operate, there's less point in having `supports`. It will introduce a lot of duplicative type checking code between matchers.

I can see the opposite argument however.

gquemener:
IMO, it should be explicit that the matcher expected a certain type and that the given one is not correct. Otherwise, just stating that no matcher "foo" was registered is confusing, as it sounds like phpspec is buggy.

I agree to switch back `supports` implementation, but to make the "no matcher" error message more explicit.

ciaranmcnulty:
I 100% agree it should be more clear to the user why the matcher isn't usable. The only bit I'm arguing for is doing that checking outside the individual matchers somehow, otherwise we rely on every matcher advertising the reasons it can't be applied.

gquemener:
After rethinking this PR, I discovered that the exception thrown in case of "no matcher found" is pretty clear: https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Runner/MatcherManager.php#L77-L87

It states that no matcher for `shouldHaveKeyWithValue(array:2) was found for value null`.

I've restored the type checking inside the matcher.

ciaranmcnulty:
Thanks for spotting and fixing this, @gquemener.

I'll make a 2.2.1 release soon for this fix.

PR #708, opened 2015-05-23
One more fix of unexpected method call verbose presenting

avant1:
That's so embarassing, but I've found another bug in current implementation of #658. 
Fatal error (Call to a member function  on a non-object) can be generated in some cases (see [scenario](https://github.com/avant1/phpspec/blob/fix-invalid-parameter-presentation/features/formatter/developer_is_shown_diffs.feature#L403) for example).


ciaranmcnulty:
Thanks for following up with the fix :+1:

PR #711, opened 2015-05-30
Improve fatal error message when running specs

Elfiggo:
Helps identify the scenario where a fatal error was caused

Includes:

Creation of a new Listener to record current example being executed
Upon a Fatal Error happening a message is displayed identifying which example had a fatal error


ciaranmcnulty:
@Elfiggo We support PHP 5.3 so your use of short array syntax is breaking the build

features/bootstrap/IsolatedProcessContext.php
```diff
@@ -73,4 +73,35 @@ public function theTestsShouldBeRerun()
...
+    public function iRunPhpspec()
+    {
+        $process = new Process(
+            $this->buildPhpSpecCmd() . ' run'
+        );
+
+        $process->run();
+        $this->lastOutput = $process->getOutput();
+
+//        expect((bool)$process->getErrorOutput())->toBe(false);
```
ciaranmcnulty: Please remove this line from the commit

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,49 @@
...
+        $this->message = $message;
+        $this->formatter = $formatter;
+    }
+
+    public function registerShutdown()
+    {
+        ini_set('display_errors', '0');
+        error_reporting(E_NOTICE);
+        register_shutdown_function(array($this, 'updateConsole'));
+        return TRUE;
```
ciaranmcnulty: Please use lowercase `true` here

src/PhpSpec/Message/MessageInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+namespace PhpSpec\Message;
+
+interface MessageInterface
+{
+    public function setMessage($currentExample);
+
+    public function getMessage();
+}
```
ciaranmcnulty: Linebreak here

src/PhpSpec/Console/Application.php
```diff
@@ -82,6 +85,7 @@ public function doRun(InputInterface $input, OutputInterface $output)
 
         $this->container->get('console.io')->setConsoleWidth($this->getTerminalWidth());
 
+        $this->container->get('process.shutdown')->registerShutdown();
         return parent::doRun($input, $output);
```
ciaranmcnulty: Line break before the return statement here

src/PhpSpec/Process/Shutdown/Shutdown.php
ciaranmcnulty: Why is this returning anything anyhow? It doesn't seem to be used outside of the specs.

features/exception_handling/developer_is_told_which_scenario_fataled.feature
```diff
@@ -0,0 +1,45 @@
+@isolated
+Feature: As a Developer
```
ciaranmcnulty: This needs a scenario covering what happens when an error occurs outside of a spec execution (e.g. syntax error in the spec so it fails at the load stage)

src/PhpSpec/Formatter/FatalFormatter.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace PhpSpec\Formatter;
+
+use PhpSpec\Console\IO;
+use PhpSpec\Formatter\Presenter\PresenterInterface;
+use PhpSpec\Listener\StatisticsCollector;
+use PhpSpec\Message\Example;
+use PhpSpec\Message\MessageInterface;
+
+class FatalFormatter extends ConsoleFormatter
```
ciaranmcnulty: This is not a formatter - formatters are selected with the `--formatter` CLI option and handle the output of the whole test execution.

You're not really using any features of ConsoleFormatter so you should drop the dependency and rename this class.

src/PhpSpec/Listener/StateListener.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+namespace PhpSpec\Listener;
+
+use PhpSpec\Event\ExampleEvent;
+use PhpSpec\Event\SuiteEvent;
+use PhpSpec\Message\MessageInterface;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class StateListener implements EventSubscriberInterface {
```
ciaranmcnulty: `StateListener` is very generic, you can be more specific about what state it's listening for. It's something like 'where we are in the lifecycle of the test suite'?

src/PhpSpec/Message/Fatal.php
```diff
@@ -0,0 +1,28 @@
...
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Message;
+
+class Fatal implements MessageInterface {
```
ciaranmcnulty: a 'Fatal` `Message` doesn't sound like 'the object that knows which example we are in' to me. This needs to be renamed.

src/PhpSpec/Message/Fatal.php
```diff
@@ -0,0 +1,28 @@
...
+    private $currentExample;
+
+    public function setMessage($currentExample)
```
ciaranmcnulty: setCurrentExample surely.

src/PhpSpec/Listener/StateListener.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * @var Example
+     */
+    private $message;
+
+    public static function getSubscribedEvents()
+    {
+        return array(
+            'beforeExample' => array('exampleMessage', -20),
+            'afterExample' => array('exampleMessage', -20),
```
ciaranmcnulty: After the example you probably want to record that we're no longer in an Example

spec/PhpSpec/Listener/StateListenerSpec.php
```diff
@@ -0,0 +1,45 @@
...
+    {
+        $this->exampleMessage($example, $message);
+    }
+
+    function it_should_call_afterSuite(SuiteEvent $example, MessageInterface $message)
+    {
+        $this->suiteMessage($example, $message);
+    }
+
+
```
GrahamCampbell: too many blank lines :P

src/PhpSpec/Console/Application.php
```diff
@@ -72,6 +74,7 @@ public function doRun(InputInterface $input, OutputInterface $output)
         $assembler = new ContainerAssembler();
         $assembler->build($this->container);
 
+
```
GrahamCampbell: ?

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -464,6 +483,7 @@ function (ServiceContainer $c) {
             }
             $c->set('event_dispatcher.listeners.formatter', $formatter);
         });
+
```
GrahamCampbell: ?

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,48 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Process\Shutdown;
+
+use PhpSpec\Formatter\FatalFormatter;
+use PhpSpec\Message\MessageInterface;
+
+class Shutdown {
```
GrahamCampbell: braces need to be on a new line

spec/PhpSpec/Process/Shutdown/ShutdownSpec.php
```diff
@@ -0,0 +1,27 @@
...
+
+    function it_should_register_the_shutdown_method()
+    {
+        $this->registerShutdown()->shouldReturn(true);
+    }
+
+    function it_should_update_the_formatter_on_shutdown(MessageInterface $message)
+    {
+        $message->getMessage()->willReturn('Hello');
+        $this->updateConsole();
```
ciaranmcnulty: There's no matcher or mock here, it's unclear that you're testing anything.

src/PhpSpec/Console/Application.php
```diff
@@ -14,6 +14,8 @@
 namespace PhpSpec\Console;
 
 use PhpSpec\Console\Prompter\Factory;
+use PhpSpec\Message\Example;
```
ciaranmcnulty: I don't think this is used

features/exception_handling/developer_is_told_which_scenario_fataled.feature
```diff
@@ -0,0 +1,45 @@
...
+  I want to know in which scenario or example my script was running
+  So that I can better trace where my changes caused a fatal error
+
+  Scenario: Spec attempts to call an undeclared function
+    Given the spec file "spec/Message/Fatal3/FatalSpec.php" contains:
+      """
+      <?php
+
+      namespace spec\Message\Fatal3;
```
ciaranmcnulty: 3? ;-)

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        ini_set('display_errors', '0');
+        error_reporting(E_NOTICE);
+        register_shutdown_function(array($this, 'updateConsole'));
+        return TRUE;
+    }
+
+    public function updateConsole()
+    {
+        $this->formatter->displayFatal($this->message);
```
ciaranmcnulty: Rather than directly couple this to the display of the error message, I'd like to see an extension point here so we can do more actions than just shutting down.

src/PhpSpec/Message/Fatal.php
Elfiggo: Not according to the message interface, might be worth renaming the variable instead. Thoughts?

ciaranmcnulty: `currentExample` is probably right because that's what you're passing in the only use case. I'm unclear why the interface is called Message

Elfiggo: My thoughts were to allow other developers to build messages for any further integration (e.g a Warning Message, Error Message....)

src/PhpSpec/Process/Shutdown/Shutdown.php
Elfiggo: Would those changes be better in a Shutdown Runner on Fatal to allow further configuration?

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,48 @@
...
+    public function __construct(MessageInterface $message, CurrentExampleWriter $currentExampleWriter)
+    {
+        $this->message = $message;
+        $this->currentExampleWriter = $currentExampleWriter;
+    }
+
+    public function registerShutdown()
+    {
+        ini_set('display_errors', '0');
+        error_reporting(E_NOTICE);
```
ciaranmcnulty: Why `E_NOTICE`?

stof:
Looks like the same commits are appearing multiple time in the history for this PR. It would be great to clean the commit history a bit

features/exception_handling/developer_is_told_which_scenario_fataled.feature
stof: ``As a developer`` should not be the feature title but the first line of the description (and the feature needs a proper title)

src/PhpSpec/Formatter/CurrentExampleWriter.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public function displayFatal(MessageInterface $message)
+    {
+        $error = error_get_last();
+
+        if (!empty($error) && $message->getMessage()) {
+            $this->output->writeln("Fatal error happened while executing the following example");
+            $this->output->writeln($message->getMessage());
+            $this->output->writeln(($error['message']));
```
stof: you have extra braces here

src/PhpSpec/Process/Shutdown/Shutdown.php
stof: changing the error reporting level seems totally wrong to me

ciaranmcnulty: It's to suppress the PHP error output, but `display_errors` is probably more appropriate?

stof: but other errors should still be reported to the error handler, not only notices

ciaranmcnulty: Agreed

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,48 @@
...
+        register_shutdown_function(array($this, 'updateConsole'));
+    }
+
+    public function updateConsole()
```
ciaranmcnulty: As discussed offline, this is very specific to the implementation.

Delegate the console updating to another object w/ appropriate interface, and have this class just trigger it, so we can add future shutdown handlers to do things like offer to fix fatals etc.

src/PhpSpec/Message/MessageInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Message;
+
+interface MessageInterface
```
ciaranmcnulty: I don't think this interface is really needed. 'Message' is a very generic concept; you would be better off naming the methods in `CurrentExample` to `setCurrentExample` and `getCurrentExample`.

src/PhpSpec/Listener/CurrentExampleListener.php
```diff
@@ -0,0 +1,57 @@
...
+
+    public function beforeExampleMessage(ExampleEvent $event)
+    {
+        $this->message->setMessage($event->getTitle());
+    }
+
+    public function afterExampleMessage(ExampleEvent $event)
+    {
+        $message = 'After the example ' . $event->getTitle();
+        $this->message->setMessage($message);
```
ciaranmcnulty: Shouldn't you be doing something like `clearCurrentExample` ? When this event fires we are no longer in an example at all. You're re-setting it to the example it's already set to.

ciaranmcnulty:
@Elfiggo please declare your new classes as `final`

src/PhpSpec/Listener/CurrentExampleListener.php
Elfiggo: What if it fatals between Last Example and New Example? extreme edge case

src/PhpSpec/Message/MessageInterface.php
Elfiggo: Will remove

src/PhpSpec/Listener/CurrentExampleListener.php
ciaranmcnulty: Then saying it errored during the last example would be incorrect

Elfiggo:
closing and resubmitting more streamlined PR

PR #714, opened 2015-06-07
Fix loop when generating classes and autoloader is not present 

### LINKED ISSUE #698: Possible Infinite loop on case insensitive filesystem
Working with an existing namespace you can end up miss typing the wrong case and get into an infinite loop on the class not existing on a case insensitive filesystem.

To solve it, it could read the spec filepath back before creating the actual class to silently fix the error for the user or just throw an error to alert the user to there mistake. 

Maybe it should know its already tried to create it even, rather than looping

Image

Notice the `namespace` & `shouldHaveType` are wrong however its in the correct Folder due to the folder already existing

image

Happy to go fix this.

ciaranmcnulty:
screen shot 2015-06-08 at 08 06 25
Fixes the common infinite loop error where the project autoloader cannot load classes that have been generated so offers to generate them again.

Previous attempts tried to do this in the same process, but were stymied by the fact composer caches misses on class_exists.

This solution passes a context between executions and checks in the new process that the types are now loadable.

Closes #698

### DIFF
--- a/src/PhpSpec/Process/Prerequisites/SuitePrerequisites.php
+++ b/src/PhpSpec/Process/Prerequisites/SuitePrerequisites.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Process\Prerequisites;
+
+use PhpSpec\Process\Context\ExecutionContextInterface;
+
+final class SuitePrerequisites implements SuitePrerequisitesInterface
+{
+    /**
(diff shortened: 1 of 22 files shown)


ciaranmcnulty:
I'm using the generic Exception presenter here, which should probably be aware of the block width and use a different wrapping mode, but that's a different PR

src/PhpSpec/Process/Prerequisites/SuitePrerequisites.php
```diff
@@ -0,0 +1,54 @@
...
+        foreach ($this->executionContext->getGeneratedTypes() as $type) {
+            if (!class_exists($type) && !interface_exists($type)) {
+                $undefinedTypes[] = $type;
+            }
+        }
+
+        if ($undefinedTypes) {
+            throw new PrerequisiteFailedException(sprintf(
+                "The type%s %s was generated but could not be loaded. Do you need to configure an autoloader?\n",
+                count($undefinedTypes) > 1 ? 's' : '',
```
jon-acker: Since you're pluralising type(s) why not was/were ?

ciaranmcnulty: True. I wish English handled this case better.

PR #715, opened 2015-06-07
Offer to generate private constructor after named constructor

### LINKED ISSUE #631: Generating private constructors
This was included in @thecrimpmaster's PR but got dropped to be re-addressed later.

After a named constructor is generated from beConstructedThrough, we should offer to generate a private constructor.

 * Would you like me to add a private constructor to User for you? [Y/n]

This would only trigger if there was no existing `__construct`

ciaranmcnulty:
screen shot 2015-06-08 at 08 08 48
After successfully creating a named constructor, if there is no 'regular' constructor offer to make a private one

Closes #631

### DIFF
--- a/src/PhpSpec/Listener/NamedConstructorNotFoundListener.php
+++ b/src/PhpSpec/Listener/NamedConstructorNotFoundListener.php
@@ -79,6 +79,17 @@ class NamedConstructorNotFoundListener implements EventSubscriberInterface
                     'arguments' => $arguments
                 ));
                 $event->markAsWorthRerunning();
+
+                if (!method_exists($classname, '__construct')) {
+                    $message = sprintf('Do you want me to make the constructor of %s private for you?', $classname);
+
+                    if ($this->io->askConfirmation($message)) {
+                        $this->generator->generate($resource, 'private-constructor', array(
+                            'name' => $method,
+                            'arguments' => $arguments
+                        ));
+                    }
+                }
             }
         }
     }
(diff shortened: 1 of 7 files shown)


ciaranmcnulty:
@thecrimpmaster included this in his named constructor PR but there was some discussion, so I advised him to remove it and resubmit later.

Possibly he hasn't found the time since then.

src/PhpSpec/Listener/NamedConstructorNotFoundListener.php
```diff
@@ -79,6 +79,17 @@ public function afterSuite(SuiteEvent $event)
                     'arguments' => $arguments
                 ));
                 $event->markAsWorthRerunning();
+
+                if (!method_exists($classname, '__construct')) {
```
shanethehat: Should this also detect a public constructor?

ciaranmcnulty: At the moment if the author has put an existing public constructor in, it might not be appropriate to modify it to private so I'm completely skipping if any constructor exists.

shanethehat: It might not be appropriate, but isn't that why the tool asks rather than just doing it?

ciaranmcnulty: It'd be a different message, and a different generator... it just seemed like an edge case to me?

By the same token, could be added in future.

PR #717, opened 2015-06-08
Add Fatal Error Message feature

Elfiggo:
Provides new functionality to display which example caused a fatal error and how.


src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -52,6 +55,9 @@ public function build(ServiceContainer $container)
         $this->setupResultConverter($container);
         $this->setupRerunner($container);
         $this->setupMatchers($container);
+        $this->setupMessage($container);
+        $this->setupShutdown($container);
+        $this->setupShutdownAction($container);
```
ciaranmcnulty: `Actions` please

src/PhpSpec/Listener/CurrentExampleListener.php
```diff
@@ -0,0 +1,57 @@
...
+        return array(
+            'beforeExample' => array('beforeExampleMessage', -20),
+            'afterExample' => array('afterExampleMessage', -20),
+            'afterSuite' => array('suiteMessage', -20),
+        );
+    }
+
+    public function __construct(CurrentExample $message)
+    {
+        $this->message = $message;
```
ciaranmcnulty: You should call this $currentExample everywhere. The 'message' stuff is confusing

src/PhpSpec/Listener/CurrentExampleListener.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public function beforeExampleMessage(ExampleEvent $event)
+    {
+        $this->message->setCurrentExample($event->getTitle());
+    }
+
+    public function afterExampleMessage()
+    {
+        $message = "";
```
ciaranmcnulty: empty string doesn't make sense to me.

src/PhpSpec/Process/Shutdown/ShutdownActionInterface.php
```diff
@@ -0,0 +1,20 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Process\Shutdown;
+
+interface ShutdownActionInterface
+{
+    public function runAction();
+    public function output();
```
ciaranmcnulty: linebreak between these please.

src/PhpSpec/Process/Shutdown/UpdateConsoleAction.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        $this->message = $message;
+        $this->currentExampleWriter = $currentExampleWriter;
+    }
+
+    public function runAction()
+    {
+        ini_set('display_errors', '0');
+        error_reporting(E_ERROR);
+        register_shutdown_function(array($this, 'output'));
```
ciaranmcnulty: This shouldn't be registering itself as a shutdown function, if we're going to have lots of `ShutdownActionInterface` instances.

Have one shutdown handler, that then calls all of the registered `ShutdownActionInterface`s

spec/PhpSpec/Listener/CurrentExampleListenerSpec.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    function it_should_implement_event_subscriber_interface()
+    {
+        $this->shouldHaveType('Symfony\Component\EventDispatcher\EventSubscriberInterface');
+    }
+
+    function it_should_call_beforeExample(ExampleEvent $example, CurrentExample $message)
+    {
+        $localMessage = 'in before example';
```
ciaranmcnulty: What's a Message?

src/PhpSpec/Formatter/CurrentExampleWriter.php
```diff
@@ -0,0 +1,45 @@
...
+
+        if (!empty($error))
+        {
+            $hhvmFatalCode = 16777217;
+
+            $fatal = isset($_ENV['HHVM']) ? ($hhvmFatalCode == $error['type']) : (E_ERROR == $error['type']);
+
+            if ($message->getCurrentExample() && $fatal) {
+                $this->output->writeln(sprintf('<failed>✘ %s</failed>', "Fatal error happened while executing the following example"));
+                $this->output->writeln(sprintf('<failed>    %s</failed>', $message->getCurrentExample()));
```
ciaranmcnulty: You're not handling the case where this isn't in an example

Elfiggo: Should there be a separate SuiteListener that is handled differently?

ciaranmcnulty: I'm just saying the message needs to be different

src/PhpSpec/Process/Shutdown/ShutdownActionInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Process\Shutdown;
+
+interface ShutdownActionInterface
+{
+    public function runAction();
+
+    public function output();
```
ciaranmcnulty: Not all shutdown actions will output something, this does not need to be part of the interface

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,48 @@
...
+        error_reporting(E_ERROR);
+        register_shutdown_function(array($this, 'runShutdown'));
+    }
+
+    public function registerAction(ShutdownActionInterface $action)
+    {
+        $this->actions[] = $action;
+    }
+
+    public function count()
```
ciaranmcnulty: What is this for?

src/PhpSpec/Listener/CurrentExampleListener.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+namespace PhpSpec\Listener;
+
+use PhpSpec\Event\ExampleEvent;
+use PhpSpec\Event\SuiteEvent;
+use PhpSpec\Message\CurrentExample;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class CurrentExampleListener implements EventSubscriberInterface {
```
ciaranmcnulty: As this implements an interface you can make it `final` easily

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return count($this->actions);
+    }
+
+    public function runShutdown()
+    {
+        foreach ($this->actions as $shutdownActions) {
```
ciaranmcnulty: I've been thinking about these ShutdownActions...

Maybe `shutdownAction` is too generic (although we may need it in future); perhaps they should be `fatalErrorAction`s?

That way you could move the `get_last_error` logic into this method, and explicitly pass the error details through to the actions when they're run, otherwise I can see us having repetition once we get more than one fatal handler.

Elfiggo: Yeah I was looking to move the error and hip hop checks elsewhere yesterday

src/PhpSpec/Process/Shutdown/Shutdown.php
Elfiggo: Classic trap

ciaranmcnulty:
I merged some other PRs which seems to have caused a conflict here, you'll need to merge or rebase from master

src/PhpSpec/Process/Shutdown/Shutdown.php
Elfiggo: Um  [InvalidArgumentException]                                                                             
  String "" is not a valid class name.                                                                   
  Please see reference document: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-0.md

Elfiggo:
More Open, Honest Feedback required :)

CHANGES.md
```diff
@@ -1,11 +1,12 @@
 Upcoming
 ========
 
-* Adds `duringInstantiation` to more easily test constructor exceptions
+* Adds `duringInstantiation()` to more easily test constructor exceptions
```
ciaranmcnulty: this diff shouldn't be showing here.

src/PhpSpec/Formatter/CurrentExampleWriter.php
```diff
@@ -0,0 +1,37 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Formatter;
+
+use PhpSpec\Console\IO;
+use PhpSpec\Message\CurrentExample;
+
+class CurrentExampleWriter
```
ciaranmcnulty: This feels more like a fatal error writer, right?

ciaranmcnulty: It might make sense to introduce an interface here (and finalise the class) - we may want to present fatals in different ways in future.

src/PhpSpec/Message/CurrentExample.php
```diff
@@ -0,0 +1,29 @@
...
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Message;
+
+class CurrentExample
```
ciaranmcnulty: Because this class is a simple value you can finalise it (although you'd have to change the tests to use a concrete value rather than a double).

src/PhpSpec/Formatter/CurrentExampleWriter.php
```diff
@@ -0,0 +1,37 @@
...
+    public function __construct(IO $io)
+    {
+        $this->output = $io;
+    }
+
+    public function displayFatal(CurrentExample $currentExample, $error = null)
+    {
+        if (!empty($error) && $currentExample->getCurrentExample())
+        {
+            $this->output->writeln(sprintf('<failed>✘ %s</failed>', "Fatal error happened while executing the following"));
```
ciaranmcnulty: This needs a non-decorated callback for plaintext consoles (see how most Formatters handle this). You can check isDecorated on the IO.

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,59 @@
...
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Process\Shutdown;
+
+class Shutdown
```
ciaranmcnulty: This can be final

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,59 @@
...
+    public function runShutdown()
+    {
+        $error = $this->getFatalError();
+
+        foreach ($this->actions as $fatalErrorActions) {
+            $fatalErrorActions->runAction($error);
+        }
+    }
+
+    public function getFatalError()
```
ciaranmcnulty: This is only called from the same class so should default to `private`

ciaranmcnulty:
Nearly there :-)

There's an odd diff; the branch history could be cleaner so maybe check the rebase.

Elfiggo:
Suggested changes made.

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -648,4 +667,38 @@ private function setupRerunner(ServiceContainer $container)
             return new PhpExecutableFinder();
         });
     }
+
+    /**
+     * @param ServiceContainer $container
+     */
+    private function setupMessage(ServiceContainer $container)
```
ciaranmcnulty: Please remove the 'message' references in the service configuration

src/PhpSpec/Formatter/FatalErrorWriter.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public function displayFatal(CurrentExample $currentExample, $error = null)
+    {
+        if (!empty($error) && $currentExample->getCurrentExample()) {
+            $failedOpen = ($this->output->isDecorated()) ? '<failed>' : '';
+            $failedClosed = ($this->output->isDecorated()) ? '</failed>' : '';
+            $failedCross = ($this->output->isDecorated()) ? '✘' : '';
+
+            $this->output->writeln(sprintf("$failedOpen$failedCross %s$failedClosed", "Fatal error happened while executing the following"));
```
ciaranmcnulty: This mixture of using double quote expansion and `sprintf` to both do placeholders is confusing; pick one or the other ;-)

src/PhpSpec/Formatter/WriterInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+namespace PhpSpec\Formatter;
+
+use PhpSpec\Message\CurrentExample;
+
+interface WriterInterface
+{
+    public function displayFatal(CurrentExample $currentExample, $error = null);
+}
```
ciaranmcnulty: Needs a newline here

src/PhpSpec/Formatter/WriterInterface.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Formatter;
+
+use PhpSpec\Message\CurrentExample;
+
+interface WriterInterface
```
ciaranmcnulty: `FatalPresenterInterface` would be more consistent with other naming

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function __construct()
+    {
+        $this->actions = array();
+    }
+
+    public function registerShutdown()
+    {
+        ini_set('display_errors', '0');
+        error_reporting(E_ERROR);
```
ciaranmcnulty: Is this needed? Errors aren't displayed so I would expect you can take this out.

src/PhpSpec/Process/Shutdown/ShutdownActionInterface.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Process\Shutdown;
+
+interface ShutdownActionInterface
+{
+    public function runAction($error = null);
```
ciaranmcnulty: Why make the error optional?

ciaranmcnulty: Oh, in case some shutdown actions don't want it? Fair enough.

src/PhpSpec/Process/Shutdown/Shutdown.php
Elfiggo: I suppose this would interfere with Devs who have this on in development. Will remove.

ciaranmcnulty:
There is a problem with this feature, it blocks any other fatal errors from showing.

e.g. if there is a parse error in a Spec I get silent output from phpspec.

This isn't something I can merge; these categories of error should all be shown.

src/PhpSpec/Formatter/FatalErrorWriter.php
```diff
@@ -0,0 +1,40 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Formatter;
+
+use PhpSpec\Console\IO;
+use PhpSpec\Message\CurrentExample;
+
+final class FatalErrorWriter implements FatalPresenterInterface
```
ciaranmcnulty: So I've been thinking... :-)

This should maybe be the responsibility of the actual selected Formatter. For instance the JUnit formatter might choose to do something different to the progress formatter in the instance of a fatal.

This would require making a new Formatter interface with an `onFatal` and handling the cases where users have a different formatter that uses the old Interface.

This solution where the Formatter and the FatalPresenter are separated out probably doesn't work for all the different formatters users might select.

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    private function getFatalError()
+    {
+        $error = error_get_last();
+        $fatal = false;
+
+        if (!empty($error)) {
+            $fatal = isset($_ENV['HHVM']) ? (self::HHVM_FATAL_ERROR == $error['type']) : (E_ERROR == $error['type']);
```
stof: strict comparisons should be used IMO

src/PhpSpec/Process/Shutdown/Shutdown.php
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        return ($fatal) ? $error : null;
```
stof: braces are not needed

src/PhpSpec/Process/Shutdown/Shutdown.php
stof: why checking in ``$_ENV['HHVM']`` ? ``defined('HHVM_VERSION')`` is the way to detect that you are running on HHVM, not an environment variable being set

src/PhpSpec/Process/Shutdown/Shutdown.php
ciaranmcnulty: This looks like it could be a lot clearer with some boolean logic reshuffling

src/PhpSpec/Process/Shutdown/Shutdown.php
Elfiggo: Will make the update. 
Travis CI passed the example, I accepted it was the solution.

Elfiggo:
Please check last commit as a way checking for parse errors

Elfiggo:
From the what I can tell I can't see output from the process because it fails before the rubber boots up. So I end up testing PHP functionality. 

I seem to be missing a point here, should I be changing the boot up sequence or changing the isolated context to get the PHP Parse error in which case aren't I just testing PHP?

This is the circle of logic I am stuck in. Help.

ciaranmcnulty:
Rubber?

ciaranmcnulty:
The IsolatedProcessContext involves two processes: The one in which Behat is running, and the one in which PhpSpec is running.

What you want to test is that the PhpSpec process outputs an error. This partly relies on PHP but also checks your changes are still allowing the error to be displayed to the user.

What you're currently testing is that if you explicitly include the file in the Behat process, the Behat process experiences an error. This is just testing PHP.

Elfiggo:
I think the Container Assembler needs dicing up.

ciaranmcnulty:
What needs to change about container assembler?

features/exception_handling/developer_is_notified_which_example_caused_a_fatal_error.feature
```diff
@@ -0,0 +1,85 @@
...
+          }
+      }
+
+      """
+    When I run phpspec
+    Then I should see "Fatal error happened while executing the following"
+    And  I should see "it fatals when calling an undeclared function"
+
+    @isolated
+	Scenario: Spec attempts to call an undeclared function with a differenet formatter
```
ciaranmcnulty: Odd indentation here?

ciaranmcnulty: The scenario title should reflect the outcome and reason, e.g. something like 'Fatal error is not shown when formatter does not support it'

spec/PhpSpec/Listener/CurrentExampleListenerSpec.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $this->shouldHaveType('PhpSpec\Listener\CurrentExampleListener');
+    }
+
+    function it_should_implement_event_subscriber_interface()
+    {
+        $this->shouldHaveType('Symfony\Component\EventDispatcher\EventSubscriberInterface');
+    }
+
+    function it_should_call_beforeCurrentExample(ExampleEvent $example)
```
ciaranmcnulty: This example doesn't test anything

spec/PhpSpec/Listener/CurrentExampleListenerSpec.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $currentExample = new CurrentExample();
+        $fatalError = 'Fatal error happened before example';
+        $example->getTitle()->willReturn($fatalError);
+        $currentExample->setCurrentExample($fatalError);
+        $this->beforeCurrentExample($example);
+    }
+
+    function it_should_call_afterCurrentExample(ExampleEvent $example)
```
ciaranmcnulty: This example doesn't test anything

spec/PhpSpec/Listener/CurrentExampleListenerSpec.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $currentExample = new CurrentExample();
+        $currentExample->setCurrentExample(null);
+        $example->getTitle()->willReturn(null);
+        $this->afterCurrentExample($example);
+    }
+
+    function it_should_call_afterSuiteEvent(SuiteEvent $example)
```
ciaranmcnulty: This example doesn't test anything

src/PhpSpec/Process/Shutdown/ShutdownActionInterface.php
ciaranmcnulty: I don't think you need the default here, it's never called with 0 parameters

src/PhpSpec/Formatter/FatalErrorWriter.php
ciaranmcnulty: Have you had a chance to address this point, @Elfiggo ?

Elfiggo: I believe I have on the latest commits

ciaranmcnulty:
@Elfiggo There are conflicts blocking this from being merged

CHANGES.md
```diff
@@ -1,4 +1,14 @@
-2.3.0-beta1 / 2015-07-04
+2.3.0-beta3 / 2015-08-08
```
ciaranmcnulty: This should not be part of your PR

bin/phpspec
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env php
 <?php
 
-define('PHPSPEC_VERSION', '2.3.0-beta');
+define('PHPSPEC_VERSION', '2.3.0-beta3');
```
ciaranmcnulty: This should not be part of your PR

features/bootstrap/IsolatedProcessContext.php
```diff
@@ -10,7 +13,18 @@
 class IsolatedProcessContext implements Context, SnippetAcceptingContext
 {
     /**
-     * @var Process
+     * @var Filesystem
+     * @beforeSuite
```
ciaranmcnulty: @beforeSuite on a variable?

features/bootstrap/IsolatedProcessContext.php
```diff
@@ -69,4 +83,45 @@ public function iShouldSeeAnErrorAboutTheMissingAutoloader()
...
+    /**
+     * @Then I should see :message
+     */
+    public function iShouldSee($message)
+    {
+        expect(strpos($this->lastOutput, $message))->toNotBe(false);
+    }
+
+    /**
+     * @Given the isolated file :file contains:
```
ciaranmcnulty: This should be in filesystemcontext and ideally integrated with the other existing methods

Elfiggo:
Going to recreate a new PR at some point. Closing for now.

PR #722, opened 2015-06-18
Delay registration of subscribers

shanethehat:
By delaying the subscriber registrations we can avoid circular dependency issues that arise when injecting the dispatcher into an object that a subscriber depends on.


ciaranmcnulty:
Makes sense, looks safe :+1:

Issue #720, opened 2015-06-16
Add Config option to auto-add files to git


kayladnls:
I propose a config option in phpspec.yml that allows for the generated spec file to be automatically added to git when you run `phpspec desc Some/Class/Name`

Thoughts?

ciaranmcnulty:
Sounds reasonable I guess. And something similar after creating new classes/interfaces?

We could make it a generic hook for 'a command to run after a file is generated' rather than git-specific

kayladnls:
Oh yes, that's a much better solution. Would we be able to pass a variable (the name of the file just created) to the command? 

I want to avoid having to use `git add --all`

ciaranmcnulty:
Yeah I'm sure some placeholder could be inserted.

I'm just thinking about whether this makes sense in the core rather than in an extension.

kayladnls:
It's something I run into constantly. If it's not a larger issue, I'd be happy to write an extension for it.

ciaranmcnulty:
Having thought about it it's a bit out of scope for the core too. If @shanethehat's #723 is merged (no reason why not right now) there'll be an event an Extension can hook into to do this, if you're up for working on one.

PR #723, opened 2015-06-19
Dispatch event when files are created

shanethehat:
Requires https://github.com/phpspec/phpspec/pull/722 to be merged first to avoid a circular dependency in the service container.

The goal of this PR is to lay the groundwork for https://github.com/phpspec/phpspec/issues/720 by exposing an event when a new file is created. This would allow a future implementation for acting on the event, either in core or as an extension.

### DIFF
--- a/src/PhpSpec/CodeGenerator/Generator/PromptingGenerator.php
+++ b/src/PhpSpec/CodeGenerator/Generator/PromptingGenerator.php
@@ -15,6 +15,7 @@ namespace PhpSpec\CodeGenerator\Generator;
 
 use PhpSpec\Console\IO;
 use PhpSpec\CodeGenerator\TemplateRenderer;
+use PhpSpec\Event\FileCreationEvent;
 use PhpSpec\Process\Context\JsonExecutionContext;
 use PhpSpec\Process\Context\ExecutionContextInterface;
 use PhpSpec\Util\Filesystem;
@@ -158,19 +159,23 @@ abstract class PromptingGenerator implements GeneratorInterface
      */
     private function generateFileAndRenderTemplate(ResourceInterface $resource, $filepath)
     {
-        $isNewFile = $this->fileAlreadyExists($filepath);
+        $fileExists = $this->fileAlreadyExists($filepath);
         $content = $this->renderTemplate($resource, $filepath);
 
         $this->filesystem->putFileContents($filepath, $content);
         $this->io->writeln($this->getGeneratedMessage($resource, $filepath));
 
-        if ($isNewFile) {
+        if (!$fileExists) {
             $this->dispatchFileCreationEvent($filepath);
         }
     }
 
+    /**
+     * @param string $filepath
+     */
     private function dispatchFileCreationEvent($filepath)
     {
-
+        $event = new FileCreationEvent($filepath);
+        $this->dispatcher->dispatch('afterFileCreation', $event);
     }
 }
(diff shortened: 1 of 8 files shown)


src/PhpSpec/CodeGenerator/Generator/PromptingGenerator.php
```diff
@@ -39,20 +41,29 @@
...
+    /**
+     * @param IO $io
      * @param TemplateRenderer $templates
-     * @param Filesystem       $filesystem
+     * @param EventDispatcherInterface $dispatcher
+     * @param Filesystem $filesystem
+     * @param ExecutionContextInterface $executionContext
      */
-    public function __construct(IO $io, TemplateRenderer $templates, Filesystem $filesystem = null, ExecutionContextInterface $executionContext = null)
+    public function __construct(IO $io, TemplateRenderer $templates, EventDispatcherInterface $dispatcher, Filesystem $filesystem = null, ExecutionContextInterface $executionContext = null)
```
ciaranmcnulty: To avoid BC breakage I'm not happy with changing the method signature in an incompatible way.

In other cases we've added optional arguments to the end of the dependency list, and you'd have to handle the case where an extension doesn't pass the dispatcher (by not dispatching that event).

src/PhpSpec/CodeGenerator/Generator/PromptingGenerator.php
```diff
@@ -148,9 +159,23 @@ private function createDirectoryIfItDoesExist($filepath)
      */
     private function generateFileAndRenderTemplate(ResourceInterface $resource, $filepath)
     {
+        $fileExists = $this->fileAlreadyExists($filepath);
         $content = $this->renderTemplate($resource, $filepath);
 
         $this->filesystem->putFileContents($filepath, $content);
         $this->io->writeln($this->getGeneratedMessage($resource, $filepath));
+
+        if (!$fileExists) {
```
ciaranmcnulty: What is this check guarding against? The file already existed but we generated it anyway?

ciaranmcnulty:
No major objections; will this be useful for MageSpec XML generation?

src/PhpSpec/CodeGenerator/Generator/PromptingGenerator.php
ciaranmcnulty: OR actually this takes the total dependency count quite high, maybe it could be achieved via Decoration instead (just an idea)?

src/PhpSpec/CodeGenerator/Generator/PromptingGenerator.php
shanethehat: If the file exists and the user chooses to overwrite it then it would be a change, not a file creation. There could potentially be an event dispatched when an existing file changes if a use case arises.

shanethehat:
This probably won't be useful for Magespec at the moment. Magespec uses the `afterExample` and `afterSuite` events to determine whether any XML updates need to be made, and I think that's more robust. It will however mean that Magespec needs some work to dispatch the `fileCreated` event, because currently non of Magespec's generators use the PromptingGenerator abstract.

src/PhpSpec/CodeGenerator/Generator/PromptingGenerator.php
shanethehat: Hmm... doing it by decoration could be a viable option. Maybe the prompting generator could also be a generator instead of an abstract....

ciaranmcnulty:
Looks good, thanks!

PR #724, opened 2015-06-22
Create constructors as first method in class

shanethehat:
Currently when allowing phpspec to create a constructor, it places it as the last method in a class. This is because it uses the same logic as for placing a standard method, but by convention a constructor should be the first method in a class if one exists.


ciaranmcnulty:
:+1: This has bugged me for ages.

ciaranmcnulty:
Should probably do the same thing for named constructors, right?

shanethehat:
Yeah, I probably should do that. What would be the convention there, private constructor first or named constructor method first?

ciaranmcnulty:
I would say `__construct` first, then the others, then the normal methods

* New `__construct` goes at the top
* New private constructor goes at the top, but after `__construct` if it exists
* Other methods go at the end

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
```diff
@@ -23,6 +23,9 @@
  */
 class MethodGenerator implements GeneratorInterface
 {
+    const METHOD_PLACEMENT = '/}[ \n]*$/';
+    const CONSTRUCTOR_PLACEMENT = '/\n(?=\s*(private|protected|public)?\s?function)/';
```
ciaranmcnulty: May contain `static`

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
```diff
@@ -111,4 +113,27 @@ protected function getTemplate()
...
+        return preg_replace(self::METHOD_PLACEMENT, rtrim($snippetToInsert)."\n}\n", trim($code));
+    }
+
+    /**
+     * @param $code
+     * @return bool
+     */
+    private function codeContainsAFunction($code)
+    {
+        return false !== strpos($code, 'function');
```
ciaranmcnulty: This doesn't guarantee the code contains a function...

shanethehat: it doesn't? Can you give an example of where it will fail?

ciaranmcnulty: ^ edited in

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
shanethehat: Or `final` ... anything else I've missed?

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
shanethehat: Regex it is then I suppose. I was trying to avoid that.

ciaranmcnulty: Regex still isn't that safe; we should probably use the tokenizer more for this stuff

features/code_generation/developer_generates_method.feature
```diff
@@ -42,7 +42,6 @@ Feature: Developer generates a method
 
       class Markdown
       {
-
```
ciaranmcnulty: This makes me very happy

Issue #727, opened 2015-06-23
Cannot specify a closure as a callable to beConstructedThrough


araines:
Documentation suggests this should be able to take a callable, but closures do not work:

```
$this->beConstructedThrough(
    function () {
        return new Class();
    }
);
$this->getProperty()->shouldBe('foo');
```

```
        warning: method_exists() expects parameter 2 to be string, object given in
        /vendor/phpspec/phpspec/src/PhpSpec/Wrapper/Subject/Caller.php line 317
```

ciaranmcnulty:
Fixed by #728

PR #728, opened 2015-06-23
Fixes issue #727

araines:
must check that the method is a string before attempting to call method_exists, otherwise it could be a callable


ciaranmcnulty:
The way you have it's great, sorry I didn't merge this earlier!

PR #729, opened 2015-06-26
MISC Remove expect usage in IsolatedProcessContext

mike182uk:
This just removes the need for `expect` when running steps from `IsolatedProcessContext`


ciaranmcnulty:
Really nice work @mike182uk :+1: 

Now we have this we don't need the separate `smoke` profile in `behat.yml`; the `smoke` suite can move into the default profile.

I'm not really keen on adding the @php-version to everything tagged smoke or isolated, it feels like something we should do elsewhere. One way would be to get the 5.3.3 travis build to run with `--tags=~@smoke` or something similar.

This certainly qualifies for the two-🍺 bounty I mentioned :-)

stof:
does it work on Windows too ? and on PHP versions compiled with ``--enable-sigchild`` ? ``proc_open`` is not exactly a cross-platform API (which is exactly why the Symfony Process component exists to normalize the behavior)

stof:
As solution might be to use symfony/process 2.6+, which allows passing a stream for the input, which could then allow to write to this stream.

mike182uk:
@stof `symfony/process` uses `proc_open` under the hood. I agree though that using `symfony/process` would be better as it will handle more edge cases than my helper class. 

I have swapped out my helper for `symfony/process`.

composer.json
```diff
@@ -34,8 +33,8 @@
         "behat/behat":           "^3.0.11",
         "bossa/phpspec2-expect": "~1.0",
         "symfony/filesystem":    "~2.1",
-        "symfony/process":       "~2.1",
-        "phpunit/phpunit":       "~4.4"
+        "phpunit/phpunit":       "~4.4",
+        "symfony/process":       "2.6"
```
GrahamCampbell: Why have you locked this at 2.6.0?

mike182uk: Ah, i meant to lock it at `2.6.*`. `2.7` requires `php-5.3.9` but travis still runs tests against `php-5.3.3` (so composer install would fail using `2.7` on `5.3.3`). I've updated the PR.

ciaranmcnulty:
@mike182uk Thanks for the efforts here.

ciaranmcnulty:
@mike182uk I will buy you those drinks when I see you next :-)

composer.json
```diff
@@ -34,8 +33,8 @@
         "behat/behat":           "^3.0.11",
         "bossa/phpspec2-expect": "~1.0",
         "symfony/filesystem":    "~2.1",
-        "symfony/process":       "~2.1",
-        "phpunit/phpunit":       "~4.4"
+        "phpunit/phpunit":       "~4.4",
+        "symfony/process":       "2.6.*"
```
stof: why not ``~2.6`` to allow using maintained versions ?

ciaranmcnulty: Good catch, thanks. I'll change in master when I'm at a desktop

Issue #700, opened 2015-05-07
[Request] Allow FQCN as parameter to run associated spec


hackel:
I keep finding myself in a situation where I want to run e.g. `phpspec run 'MyApp\Whatever'` instead of `phpspec run spec/MyApp/WhateverSpec.php`.  Would it be possible to add the ability to detect that?  E.g., if the parameter isn't found as a spec file on disk, check for a PSR-4 class instead?  This would be really convenient!

jakubzapletal:
Because the issue with backslashes isn't in phpspec but how arguments are handled in the shell, then the only solution would be to add a feature `phpspec run MyApp/Whatever`.

jakubzapletal:
I forgot another solutions could be `phpspec run MyApp\\Whatever` or `phpspec run "MyApp\Whatever"`, but it depends if this functionality is desirable.

jakzal:
As @jakubzapletal reminded this is an issue with shell, not PhpSpec. PhpSpec will receive the value with no slashes as your shell will use them as escape sequences. For better explanations and some solutions see https://github.com/phpspec/phpspec/issues/326#issuecomment-38039058

hackel:
This isn't an issue with backslashes, however.  I'm used to escaping classes with quotes already, which, as mentioned in that bug report, works fine with `describe`.  It does not work fine with `run`, however.

When I run `phpspec run 'MyApp\Whatever'`, phpspec looks for the file `MyApp/Whatever` (e.g. it's converting backslashes to forward slash directory separators), and not even looking in the `spec_path` directory.  I suppose it's trying to maintain compatibility for Windows users or something.  In any case, this returns a file not found.  What I'm proposing is that when this file not found is encountered, it checks to see if it can find a spec file associated with the class name, as it is able to do when running `phpspec run` without any arguments.

Even `phpspec run 'spec\MyApp\WhateverSpec'` doesn't work because of this issue, it needs a full path, regardless of whether it's using back or forward slashes.

PR #730, opened 2015-06-27
Run a named spec using fully qualified class name

shanethehat:
Attempts to address https://github.com/phpspec/phpspec/issues/700 by checking whether the spec argument passed to the run command can be translated into a fully qualified class name for an existing spec file. If it can then the identified spec is executed, otherwise the original behaviour is used.


ciaranmcnulty:
Looks great - is it possible to add this to the documentation too?

PR #732, opened 2015-07-05
Fix: Add missing param annotation to docblock

localheinz:
This PR

* [x] adds a missing parameter annotation to a constructor docblock


ciaranmcnulty:
Thanks!

PR #733, opened 2015-07-05
Fix: SuitePrerequisites vs. SuitePrerequisitesInterface

localheinz:
This PR

* [x] fixes docblocks which refer to an implementation (`SuitePrerequisites `), while the corresponding interface is probably intended to be used (`SuitePrerequisitesInterface`)


ciaranmcnulty:
Yep this looks correct, thanks

PR #734, opened 2015-07-05
Fix: Simplify FQCN

localheinz:
This PR

* [x] simplifies a few unnecessary fully qualified class names

### DIFF
--- a/src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
+++ b/src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
@@ -26,17 +26,17 @@ use PhpSpec\CodeGenerator\Writer\TokenizedCodeWriter;
 class MethodGenerator implements GeneratorInterface
 {
     /**
-     * @var \PhpSpec\Console\IO
+     * @var IO
      */
     private $io;
 
     /**
-     * @var \PhpSpec\CodeGenerator\TemplateRenderer
+     * @var TemplateRenderer
      */
     private $templates;
 
     /**
-     * @var \PhpSpec\Util\Filesystem
+     * @var Filesystem
      */
     private $filesystem;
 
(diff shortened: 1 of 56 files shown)


localheinz:
@ciaranmcnulty 

To be honest, I'm a bit confused - there are so many fully qualified class names, while at the same time, these are imported, is there a reason I'm unaware of why that is done?

ciaranmcnulty:
No idea, to be honest. I think some IDEs have historically been poor at resolving class aliases, which may be a reason?

GrahamCampbell:
I'm REALLY :-1: on this. PHPDoc should be human readable IMO. It should be immediately obvious to a human on first read what the types are.

ciaranmcnulty:
@GrahamCampbell Does the FQCN make it more obvious what the type is? I'm not convinced it does.

ciaranmcnulty:
@localheinz Merging your other PRs seems to have caused a problem with the merge here

GrahamCampbell:
> Does the FQCN make it more obvious what the type is? I'm not convinced it does.

`Foo` is totally non obvious, but `Vendor\Foo\Baz\Foo` is super clear what the type is.

ciaranmcnulty:
In typehints or class instantiation, we use short names because it is easier to grok what the code is doing if we're not seeing FCQNs all over the place. For me the same logic applies to docblocks.

localheinz:
@ciaranmcnulty 

Rebased!

everzet:
@GrahamCampbell if `Foo` doesn't give you a clear idea on what type is it, then `Foo` is a bad name. I see this all the time, people think that namespace is part of the class name - IMO it is not, it is a context.

ciaranmcnulty:
Thanks for this and the other improvements @localheinz

localheinz:
You're welcome, @ciaranmcnulty!

PR #739, opened 2015-07-10
add an empty line at the end of generated files

nsitbon:
[PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#22-files) requires each PHP files to be ended with an empty line. This PR enforces this rule.

### DIFF
--- a/features/code_generation/developer_generates_class.feature
+++ b/features/code_generation/developer_generates_class.feature
@@ -17,6 +17,7 @@ Feature: Developer generates a class
       {
       }
 
+
       """
 
   @issue269
@@ -40,6 +41,7 @@ Feature: Developer generates a class
     {
     }
 
+
     """
 
   @issue127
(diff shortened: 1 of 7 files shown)


GrahamCampbell:
There is already one isn;t there?

nsitbon:
running 
`vendor/bin/phpspec desc Foo`
generates a spec file without ending empty line.

ciaranmcnulty:
We currently have a LF at the end, not a blank line. I personally don't like having a blank line and don't see many projects checking for that, but it does seem to be a PSR-2 requirement.

ciaranmcnulty:
Some debate here https://groups.google.com/forum/#!topic/php-fig/PSzmcFVQst0

ciaranmcnulty:
We currently end with `}\n` and PSR-2 seems to imply we should end with `}\n\n`

Even though that's stupid IMO

nsitbon:
My bad : vim was hiding the trailing empty line. I'm sorry for having wasted your time.
The good point is that it allows me to catch a bug in behat.

Issue #741, opened 2015-07-10
It won't create method for me if the class followed by a comment.


erdemkeren:
Hello, here is an example php file structure:

    <?php
        namespace Foo;

        class Bar
        {
            [...]
        }

    // End of bar class.

And the spec:

    <?php
        class BarSpec extends ObjectBehavior
        {

            function it_is_initializable()
            {
                $this->shouldHaveType('Foo\Bar');
            }

            function it_calls_a_method_to_do_sth()
            {
                $this->doSth();
            }
            
        }
        
If you are describing a class inside a file structure like that, phpspec will endlessly ask to create the undefined method for you, but it won't be able to do it. Like this:

> Foo/Bar
>  9  - it calls a method to do sth
>      method Foo\Bar::doSth not found.
>
>                        50%                              1
> 1 specs
> 1 examples (1 passed, 1 broken)
> 1ms
>
>
>  Do you want me to create
>  `Foo\Bar::doSth()` for you?
>                                                 [Y/n]
> Y
>   Method Foo\Bar::doSth() has been created.
> 
> Foo/Bar
>  28  - it calls a method to do sth
>      method Foo\Bar::doSth not found.
>
>                        50%                              1
> 1 specs
> 1 examples (1passed, 1 broken)
> 2ms
>
>
>  Do you want me to create
>  `Foo\Bar::doSth()` for you?
>                                                 [Y/n]
> Really?
> fooproj [master] $

ismatkurt:
+1

shanethehat:
Which version of phpspec are you using? This might be fixed in the 2.3.0 beta because the logic for placing the new method was rewritten.

erdemkeren:
Hmm I'll check it out. Using 2.1 at the moment.

erdemkeren:
I updated to 2.3.0 beta; yes, the behavior changed but still buggy. When I add the comment at the end of the file, it placed the new function between the namespace and the use lines, causing a syntax error like;

    <?php
    namespace Foo\Bar;

    use Foo\Baz;
    use Foo\Qux;
    use Foo\Ba

        public function bugOnMethodCreate()
        {
            // TODO: write logic here
        }

    rBaz;
    use ...;

    /**
     * Class starts here
     */
    class Quux 
    {
        // Class body here.
    }

    // End of quux class.

ciaranmcnulty:
Hanks for catching it. I don't think we need to fix in older versions because the model is so different but it should be fixed for 2.3

[...]

PR #745, opened 2015-07-14
#741 Place methods specifically after the last one

shanethehat:
Fixes an issue raised in #741 which seems to have been caused by the existing method in a SUT being followed by a newline. This adds specific detection for the end of the last method of the class, ignoring any trailing newlines.


ciaranmcnulty:
Thanks!

PR #747, opened 2015-07-21
Refactoring of the Presenter objects

shanethehat:
Breaking the existing large presenter object up into smaller collaborating objects will make it easier to modify the behaviour of individual parts, either as part of phpspec core or by extensions. One such change that this would enable would be to stop truncating strings in failure output when a verbose flag is passed.

Scrutinizer will not pass until https://github.com/phpspec/phpspec/pull/748 is merged.

### DIFF
--- a/src/PhpSpec/Formatter/Presenter/Exception/GenericPhpSpecExceptionPresenter.php
+++ b/src/PhpSpec/Formatter/Presenter/Exception/GenericPhpSpecExceptionPresenter.php
@@ -13,9 +13,7 @@
 
 namespace PhpSpec\Formatter\Presenter\Exception;
 
-final class GenericPhpSpecExceptionPresenter
-    extends AbstractPhpSpecExceptionPresenter
-    implements PhpSpecExceptionPresenter
+final class GenericPhpSpecExceptionPresenter extends AbstractPhpSpecExceptionPresenter implements PhpSpecExceptionPresenter
 {
     /**
      * @var ExceptionElementPresenter
--- a/src/PhpSpec/Formatter/Presenter/Exception/GenericPhpSpecExceptionPresenter.php
+++ b/src/PhpSpec/Formatter/Presenter/Exception/GenericPhpSpecExceptionPresenter.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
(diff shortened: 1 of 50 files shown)


.scrutinizer.yml
```diff
@@ -20,7 +20,12 @@ tools:
...
 
+checks:
+  php:
+    duplication: false
+
 filter:
   excluded_paths:
     - docs/*
     - vendor/*
+    - src/PhpSpec/Formatter/Presenter/StringPresenter.php
```
stof: why are you excluding it ? This looks bad

src/PhpSpec/Formatter/Presenter/Exception/GenericPhpSpecExceptionPresenter.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Formatter\Presenter\Exception;
+
+final class GenericPhpSpecExceptionPresenter
+    extends AbstractPhpSpecExceptionPresenter
+    implements PhpSpecExceptionPresenter
```
stof: this should stay on a single line

src/PhpSpec/Formatter/Presenter/Value/ValuePresenter.php
```diff
@@ -0,0 +1,19 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: shane
+ * Date: 21/07/15
+ * Time: 08:44
```
stof: this comment is wrong

.scrutinizer.yml
shanethehat: This is part of a temp commit that was introduced to show the improvement the refactoring has on the code quality metrics. It should be removed before merging.

src/PhpSpec/Formatter/Presenter/Exception/GenericPhpSpecExceptionPresenter.php
ciaranmcnulty: Won't that make it too wide?

stof: There is no hard restriction on line length in PSR-2, only a recommendation to avoid making lines too long.

and all other code in phpspec keeps the class declaration on a single line IIRC (btw, the Symfony CS, which are built on top of PSR-2, are indeed mandating to keep in on a single line)

PR #748, opened 2015-07-22
Disable superglobal rule checking

shanethehat:
A recently added Scrutinizer rule is making builds fail because it believes that superglobals are bad.


shanethehat:
The build says it's broken for an issue that is now ignored, but if I trigger a rebuild by rebasing it will forget that the issue was ignored.

PR #752, opened 2015-07-30
ValidatingExtensionTrait first implementation

gajdaw:
Hi,
I think that developing  extensions is quite difficult right now. That's because when you declare an incorrect service, for example matcher with `StdClass()` in your extension `load()` method:

     /**
     * @param ServiceContainer $container
     */
    public function load(ServiceContainer $container)
    {
        $container->set('matchers.abc', function (ServiceContainer $c) {
            return new \StdClass();
        });
    }

then `phpspec` will silently die. It took me quite a while to debug this.
The trait in this PR aims to make writing extensions a little easier.

### DIFF
--- a/src/PhpSpec/ServiceContainer.php
+++ b/src/PhpSpec/ServiceContainer.php
@@ -234,7 +234,7 @@ class ServiceContainer
      *
      * @return array
      */
-    private function getPrefixAndSid($id)
+    public function getPrefixAndSid($id)
     {
         if (count($parts = explode('.', $id)) < 2) {
             return array(null, $id);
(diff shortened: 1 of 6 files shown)


src/PhpSpec/ServiceContainer.php
```diff
@@ -234,7 +234,7 @@ public function configure()
      *
      * @return array
      */
-    private function getPrefixAndSid($id)
+    public function getPrefixAndSid($id)
```
stof: I'm not a fan of exposing this method. It is an internal implementation detail.

gajdaw: I solved this by creating identical private method `getPrefixAndSid()` in the trait.

The code is in my branch:
https://github.com/gajdaw/phpspec/tree/base-extension-trait

Because PR is closed the code is not updated here. I wasn't sure if I would be able to fix broken tests on PHP 5.3. But I did it. 

If you consider this solution valuable can you, please, reopen this PR?

PR #753, opened 2015-07-30
Doc: info about throwing an exception inside inline matcher

gajdaw:
An example how to produce more verbose info within inline matcher.

### DIFF
--- a/docs/cookbook/matchers.rst
+++ b/docs/cookbook/matchers.rst
@@ -477,3 +477,40 @@ You can create custom matchers using the Inline matcher.
             ];
         }
     }
+
+In order to print a more verbose error message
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Thanks!

ladhadha:
Hi, I try to make my inline matchers based on [this](http://www.phpspec.net/en/latest/cookbook/matchers.html).

```
use PhpSpec\ObjectBehavior;
use PhpSpec\Matcher\InlineMatcher;
use PhpSpec\Exception\Example\FailureException;
```

But I can't find this class `PhpSpec\Matcher\InlineMatcher`. So where it is? Am I miss something dependency to install?

PR #756, opened 2015-08-04
Developer uses extension - feature

gajdaw:
As of now, I belive this is not covered by tests.

### DIFF
--- a/features/extensions/developer_uses_extension.feature
+++ b/features/extensions/developer_uses_extension.feature
@@ -1,0 +1,160 @@
+Feature: Developer uses extension
+  As a Developer
+  I want to use my extension
+
+  Scenario: Extension can provide a new matcher
+    Given the config file contains:
+    """
+    extensions:
+      - Example1\PhpSpec\MatcherExtension\Extension
+    """
+    And the class file "src/Example1/PhpSpec/MatcherExtension/Extension.php" contains:
+    """
+    <?php
+
+    namespace Example1\PhpSpec\MatcherExtension;
+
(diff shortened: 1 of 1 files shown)


stof:
AFAICT, these scenarios are not testing anything. If we disable the loading of extensions in PhpSpec, your scenarios will still pass as there is nothing relying on the extension in the suite you are running

features/extensions/developer_uses_extension.feature
```diff
@@ -0,0 +1,96 @@
...
+    use PhpSpec\Matcher\IdentityMatcher;
+
+    class Extension implements ExtensionInterface
+    {
+        /**
+         * @param ServiceContainer $container
+         */
+        public function load(ServiceContainer $container)
+        {
+          $container->set('matchers.anotherIdentityMatcher', function (ServiceContainer $c) {
```
stof: wrong indentation

ciaranmcnulty:
I agree with @stof that this test doesn't really show that anything works.

The real test is a real extension passing or failing. Perhaps we should be doing some sort of large build against known/supported extensions before a major release?

gajdaw:
Thanks for you comments. I haven't thought about using the extension. But, of course, you are right. Including an extension in config doesn't call any methods defined in the extension.

I included a new test case `it_should_succeed_in_using_new_matcher()` that uses the new matcher defined in the extension. This required the complete implementation of a new matcher.

stof:
this scenario is covering that extensions can provide matchers. This is what its description should be

gajdaw:
@stof Done.

everzet:
:+1: good job

PR #757, opened 2015-08-08
Move `symfony/process` dependency to non-dev dependencies

sroze:
This PR moves the `symfony/process` dependency in the non-dev dependencies of PhpSpec.

Else, if the SUS do not require this package, PhpSpec won't run anymore because the `Symfony\Component\Process\PhpExecutableFinder` class is required for PhpSpec to run.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,6 +24,7 @@
         "sebastian/exporter":       "~1.0",
         "symfony/console":          "~2.3",
         "symfony/event-dispatcher": "~2.1",
+        "symfony/process":          "^2.6",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
This was incorrectly moved in #729 

Because Travis installs the dev dependencies, nothing failed there but it will fail for the PHAR or indeed any projects that include it as a dependency

ciaranmcnulty:
Thanks for catching this @sroze !

sroze:
:+1:

Issue #762, opened 2015-08-24
Reporter truncates expected result so so that it is not readable.


lukasoppermann:
Hey,

I get errors like this:

```
✘ should automatically add an about link
        expected [array:1] to have value "http://dev.formandsystem...." for about
        key, but found "http://dev.formandsystem....".
```

As you can imagine, this is not very handy, could you maybe add an option to not truncate at all?

mheki:
Thanks for reporting @lukasoppermann. If you expect string and increase verbosity level (`-v`), you will see full strings compared. Looks like this doesn't work properly for arrays.

PR #763, opened 2015-08-24
Do not truncate string values to 25 characters when diffing arrays

mheki:
This PR solves an issue https://github.com/phpspec/phpspec/issues/762
Helps to see longer strings while comparing arrays in verbose mode.


avant1:
:+1:

stof:
you are still truncating lines in the case of multiline strings, which is inconsistent

mheki:
thanks @stof, that's also changed now.

ciaranmcnulty:
This looks great, thanks!

PR #764, opened 2015-08-28
Update class and interface template

umpirsky:
Classes used to have one extra line before first method/constructor in the class:

```php
<?php

namespace App\Domain;

class Book
{

    public function getTitle()
    {
        // TODO: write logic here
    }
}
```

After my update, it looks like this:

```php
<?php

namespace App\Domain;

class Book
{
    public function getTitle()
    {
        // TODO: write logic here
    }
}
```

### DIFF
--- a/src/PhpSpec/CodeGenerator/Generator/templates/class.template
+++ b/src/PhpSpec/CodeGenerator/Generator/templates/class.template
@@ -1,5 +1,4 @@
 <?php%namespace_block%
 
 class %name%
-{
-}
+{}
--- a/src/PhpSpec/CodeGenerator/Generator/templates/interface.template
+++ b/src/PhpSpec/CodeGenerator/Generator/templates/interface.template
@@ -1,5 +1,4 @@
 <?php%namespace_block%
 
 interface %name%
-{
-}
+{}


Ocramius:
@umpirsky this fixes a symptom, not the cause. And it breaks PSR-2 compat on empty classes.

umpirsky:
@Ocramius True, I know, i just thought that in most cases class will not be empty. :)

I guess more appropriate fix would be to have first method template and add empty line only if it's not the first thing after `{`.

Ocramius:
Indeed, that would be the correct fix.

umpirsky:
@Ocramius But harder to implement. :trollface:

Ocramius:
Computering is hard: deal with it.

umpirsky:
:smile:

ciaranmcnulty:
I believe the work that @shanethehat did for 2.3.0 addresses this; if it doesn't I think an extension to his approach is appropriate rather than breaking the template in a non-PSR2 way.

ciaranmcnulty:
@umpirsky Can you check this is still the behaviour for `master` and if so can you please open an issue so we can discuss the best way to fix?

I appreciate your contribution!

umpirsky:
@ciaranmcnulty Oh, it actually works in master, cool! Thanks. :+1:

PR #765, opened 2015-08-28
ValidatingExtensionTrait first implementation

gajdaw:
This PR contains a trait that can help to validate services injected inside an extension.

Right now it only validates matchers but it can be extended by assigning new values to the map stored `$map` private property.

The whole concept was to allow validation of injected service only for user created services injected in the extension. The services injected within `setupMatchers` method in `ContainerAssembler` class are left intact.

### DIFF
--- a/features/extensions/developer_uses_incorrect_extension.feature
+++ b/features/extensions/developer_uses_incorrect_extension.feature
@@ -1,0 +1,56 @@
+@php-version @php5.4
+Feature: Developer uses incorrect extension
+  As a Developer
+  I want to use my extension
+  In order to avoid spending a lot of time debugging I want to see a clear error message
+
+  Scenario: Using extension with incorrect matcher
+    Given the config file contains:
+    """
+    extensions:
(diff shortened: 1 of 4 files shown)


ciaranmcnulty:
What's the motivation for this feature? I might be wrong but isn't this protected already via type hints?

gajdaw:
If I am not mistaken right now you can pass just anything as a matcher service. Even a callable that returns an object that does not implement PhpSpec\Matcher\MatcherInterface.
In that case phpspec will silently fail.
Using the trait included in this PR you will get an exception with clear message.

So it could be of some value for extension developers. However, the price is the additional step to validate the service.

ciaranmcnulty:
I'm afraid after a lot of thought I don't think this is appropriate to core - I don't see the value of checking these things at runtime rather than, for instance, writing tests for your extension.

However if you want to open-source it in your own repository for other extension developers to make use of I'd be happy to link to it in the docs.

gajdaw:
Yes, that's good idea. I wasn't sure if this should go to the core either.

PR #766, opened 2015-08-28
Add string contain matcher

mheki:
This PR adds matcher which helps to check if the string contains given substring.
It allows to write: `$this->getSlug()->shouldContain('foo');` and can help to write more readable and simpler examples.


ciaranmcnulty:
Can you please add a feature to the behat coverage as well?

ciaranmcnulty:
And come to think of it, the docs could be updated as part of this same PR

mheki:
Sure @ciaranmcnulty

mheki:
Both added

ciaranmcnulty:
Looks good, thanks!

PR #772, opened 2015-09-09
Remove default formatters from config

ciaranmcnulty:
These are in there from someone's personal preference, and are not my preferred formatters.

Rather than impose my own preferences I've removed them so that users can use config of their choice.


GrahamCampbell:
Nice change. :+1:

Issue #710, opened 2015-05-27
Array type hint in constructor raises invalid argument exception


Sukhrob:
In class file:

    public function __construct(array $array)
    {
        $this->array= $array;
    }

In spec file:

    function let(array $array) // Tried with and without "array" type hint
    {
        $this->beConstructedWith($array);
    }

Raises
error: Argument 1 passed to __construct() must be of type array, object given.

avant1:
phpspec assumes that everything you want to get in example (or let/letgo methods) arguments is a "mock" (or a spy, i'm actually not good at terminology) of actual object you will use in that example.

When you don't type hint `let` method argument, special object (which is instanceof `\stdClass`) is being passed to your constructor. But if `__construct` of actual class has `array` type hint, then php will generate error (cause object is not array obviously).

When you type hint spec method argument as `array`, special object is being passed to spec method too, and error is being raised for the same reason.

You can solve this problem passing regular array (created manually) to `beConstructedWith`.
Your spec file could contain:
```php
function let()
{
    $this->beConstructedWith(array());
}
```
Usually constructor array parameter infuences the object behaviour, so it makes sense to pass different arrays and test them separately. So usually you don't ask phpspec "just give me any array I could create my object". But if you need "any array" - you can easily create empty array manually. 

Also I agree it is a bit confusing for phpspec newcomers. Probably more verbose error should be raised when example argument is type hinted as array.

Sukhrob:
You are right, @avant1. Thanks

PR #775, opened 2015-09-17
Invalid collaborator type hinting error handling

avant1:
When collaborator is defined with array or callable type hinting spec execution raises not really clear php warnings (something like *Argument 1 passed to method() must be array, object given*).

This PR makes such situations more clear (this can be espesially important for newcomers).

I've faced this problem when started working with phpspec, and I am not alone (see #710).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,7 @@ install:
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
 before_script:
-  - echo "<?php if (PHP_VERSION_ID >= 50400) echo ',@php5.4';" > php_version_tags.php
+  - echo "<?php if (PHP_VERSION_ID >= 50400) echo ',@php5.4'; if (PHP_VERSION_ID >= 70000) echo ',@php7';" > php_version_tags.php
 
 script:
    - bin/phpspec run --format=pretty
--- a/src/PhpSpec/Exception/Wrapper/InvalidCollaboratorTypeException.php
+++ b/src/PhpSpec/Exception/Wrapper/InvalidCollaboratorTypeException.php
@@ -20,8 +20,8 @@ class InvalidCollaboratorTypeException extends CollaboratorException
     public function __construct(\ReflectionParameter $parameter, \ReflectionFunctionAbstract $function)
     {
         $message = sprintf(
-            'Collaborator cannot be array or callable: argument %s defined in %s. ' .
-            'You can create arrays/callables manually.',
+            'Collaborator must be an object: argument %s defined in %s. ' .
+            'You can create non-object values manually.',
             $parameter->getPosition(),
             $this->fetchFunctionIdentifier($function)
         );
--- a/src/PhpSpec/Exception/Wrapper/InvalidCollaboratorTypeException.php
+++ b/src/PhpSpec/Exception/Wrapper/InvalidCollaboratorTypeException.php
@@ -1,0 +1,44 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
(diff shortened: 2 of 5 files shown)


avant1:
Fixed problem on php5.3 (callable type was introduced in 5.4).

ciaranmcnulty:
This is a great addition. Perhaps it can also check for scalar type hints in PHP7, as they're also not something that's mockable.

avant1:
Thought about that too. In php7 was added [`ReflectionType` class](http://php.net/manual/en/class.reflectiontype.php), so check becomes even easier. I'll add appropriate changes later.

src/PhpSpec/Exception/Wrapper/InvalidCollaboratorTypeException.php
```diff
@@ -0,0 +1,44 @@
...
+namespace PhpSpec\Exception\Wrapper;
+
+
+class InvalidCollaboratorTypeException extends CollaboratorException
+{
+
+    public function __construct(\ReflectionParameter $parameter, \ReflectionFunctionAbstract $function)
+    {
+        $message = sprintf(
+            'Collaborator cannot be array or callable: argument %s defined in %s. ' .
```
stof: I would say ``Collaborator must be an object`` instead. Scalar typehints are not supported either

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
```diff
@@ -134,6 +137,20 @@ private function generateCollaborators(CollaboratorManager $collaborators, \Refl
         }
     }
 
+    private function ensureParameterHasSupportedTypeHinting(
+        \ReflectionParameter $parameter,
+        \ReflectionFunctionAbstract $function
+    ) {
+        $isUnsupportedTypeHinting = $parameter->isArray();
+        if (version_compare(PHP_VERSION, '5.4.0', '>')) {
```
stof: Use ``PHP_VERSION_ID >= 50400``. OPCache can optimize such check at compile time while it cannot for a function call.

avant1:
Added necessary changes.

.travis.yml
```diff
@@ -33,7 +33,7 @@ install:
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
 before_script:
-  - echo "<?php if (PHP_VERSION_ID >= 50400) echo ',@php5.4';" > php_version_tags.php
```
ciaranmcnulty: Hm, we need a cleaner way to handle this don't we?

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
```diff
@@ -134,6 +137,22 @@ private function generateCollaborators(CollaboratorManager $collaborators, \Refl
         }
     }
 
+    private function ensureParameterHasSupportedTypeHinting(
+        \ReflectionParameter $parameter,
+        \ReflectionFunctionAbstract $function
+    ) {
+        $isUnsupportedTypeHinting = $parameter->isArray();
```
ciaranmcnulty: You can extract this to a private to make the conditionals clearer

.travis.yml
avant1: Any suggestions? I don't have any ideas. This looks a little like a hack, but it is a small and easy to maintain hack :smile:

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
avant1: Agree.

avant1:
Added some changes, also rebased on master.

.travis.yml
ciaranmcnulty: Yeah it'll be ok for now. I'll probably write a behat extension or something that does this better

ciaranmcnulty:
Thanks - great addition!

PR #780, opened 2015-10-03
Update composer.json

ciaranmcnulty:


GrahamCampbell:
Isn't the syntax `|` not `,`?

ciaranmcnulty:
Probably

GrahamCampbell:
I think `,` is the same as `&&`, ie, both conditions must be matched.

PR #781, opened 2015-10-04
Add ext-tokenizer to composer dependencies

Ormin:
Sometimes, php is distributed with ext-tokenizer disabled ( this was a case for me when trying to run phpspec from a temporary workstation with cygwin ). Without it, it will fail with following message:

Fatal error: Call to undefined function token_get_all() in [...]/src/PhpSpec/Locator/PSR0/PSR0Locator.php on line 277.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "symfony/process":          "^2.6",
         "symfony/finder":           "~2.1",
         "symfony/yaml":             "~2.1",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Thanks, this is a good idea.

Is there a particular version of the extension that we need? IIRC we are using `token_get_all` and `token_name`.

ciaranmcnulty:
Version doesn't seem significant, thanks for the contrib.

PR #782, opened 2015-10-06
Added variadic arguments docBlock to ObjectBehavior and Wrapper\Subject

maryo:
`$this->beConstructedWith($collaborator);`
PhpStorm was complaining "Method call uses 1 parameters, but method signature uses 0 parameters"

### DIFF
--- a/src/PhpSpec/ObjectBehavior.php
+++ b/src/PhpSpec/ObjectBehavior.php
@@ -28,7 +28,7 @@ use ArrayAccess;
  * wrap the results into PhpSpec subjects. This results will then be able to
  * be matched against expectations.
  *
- * @method void beConstructedWith()
+ * @method void beConstructedWith(...$arguments)
  * @method void beConstructedThrough($factoryMethod, array $constructorArguments)
  * @method void beAnInstanceOf($class)
  * @method void shouldHaveType($type)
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Hm, is there any compatibility impact for older versions?

maryo:
I'd be very surprised if so. But I'm using phpspec for the first time today so I am not sure.

stof:
This phpdoc is here for a better IDE support, so the question is how much this is supported by IDEs.

maryo:
Exactly. I don't know how many IDEs support it. But looks like it's valid phpDoc. See https://github.com/phpDocumentor/phpDocumentor2/commit/14510b1833ed3ca3467169e8e9f5f6924342a977

But it's probably undocumented.

ciaranmcnulty:
Yeah the open question is whether older versions of IDEs would have a problem - we don't want to add something that helps PhpStorm users but hurts others (plus the use of docblocks here is a hack until real IDE support appears).

maryo:
That's hard to answer since I don't expect people to report issues when it's not merged. So the only solution is probably to test it.
What IDEs you want me to check?

ciaranmcnulty:
It's a hack because the docblock is never going to be completely correct; the real behaviour is dynamic and to support it fully IDEs have to be able to comprehend PhpSpec's configuration.


However this is a small addition and will help some

maryo:
Not sure I agree with the hack part :) but thanks for merging :+1:

PR #787, opened 2015-10-10
Add --ignore-pending run command flag & config file support

toby-griffiths:
Added behat feature file including scenarios to cover command line flag & config file, & also without this flag/config

Updated ResultConverter to use flag/config to prevent non-zero exit code being returned.

Should we also need some way of setting `--ignore-pending=false` if the config file has it set to true?

### DIFF
--- a/features/bootstrap/Matcher/ApplicationStatusCodeMatcher.php
+++ b/features/bootstrap/Matcher/ApplicationStatusCodeMatcher.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Matcher;
+
+use PhpSpec\Exception\Example\FailureException;
+use PhpSpec\Matcher\MatcherInterface;
+use Symfony\Component\Console\Tester\ApplicationTester;
+
+
+/**
+ * Matcher class to test the status code from a console command
+ *
+ * @package Matcher
+ */
+class ApplicationStatusCodeMatcher implements MatcherInterface
+{
+
+    /**
+     * Checks if matcher supports provided subject and matcher name.
+     *
+     * @param string $name
+     * @param mixed  $subject
+     * @param array  $arguments
+     *
+     * @return Boolean
+     */
+    public function supports($name, $subject, array $arguments)
+    {
+        return ($name == 'haveStatusCode' && $subject instanceof ApplicationTester);
+    }
+
+
+    /**
+     * Evaluates positive match.
+     *
+     * @param string $name
+     * @param mixed  $subject
+     * @param array  $arguments
+     */
+    public function positiveMatch($name, $subject, array $arguments)
+    {
+        $expected = $arguments[0];
+        $actual = $subject->getStatusCode();
+        if ($expected !== $actual) {
+            throw new FailureException(sprintf(
+                "Application status code did not contain expected '%s'. Actual output:\n'%s'" ,
+                $expected,
+                $actual
+            ));
+        }
+    }
+
+
+    /**
+     * Evaluates negative match.
+     *
+     * @param string $name
+     * @param mixed  $subject
+     * @param array  $arguments
+     */
+    public function negativeMatch($name, $subject, array $arguments)
+    {
+        $expected = $arguments[0];
+        $actual = $subject->getStatusCode();
+        if ($expected === $actual) {
+            throw new FailureException(sprintf(
+                "Application status code did not contain expected '%s'. Actual output:\n'%s'" ,
+                $expected,
+                $actual
+            ));
+        }
+    }
+
+
+    /**
+     * Returns matcher priority.
+     *
+     * @return integer
+     */
+    public function getPriority()
+    {
+        return 51;
+    }
+}
--- a/features/bootstrap/Matcher/ApplicationStatusCodeMatcher.php
+++ b/features/bootstrap/Matcher/ApplicationStatusCodeMatcher.php
@@ -6,7 +6,6 @@ use PhpSpec\Exception\Example\FailureException;
 use PhpSpec\Matcher\MatcherInterface;
 use Symfony\Component\Console\Tester\ApplicationTester;
 
-
 /**
  * Matcher class to test the status code from a console command
  *
@@ -14,7 +13,6 @@ use Symfony\Component\Console\Tester\ApplicationTester;
  */
 class ApplicationStatusCodeMatcher implements MatcherInterface
 {
-
     /**
      * Checks if matcher supports provided subject and matcher name.
      *
@@ -26,9 +24,9 @@ class ApplicationStatusCodeMatcher implements MatcherInterface
      */
     public function supports($name, $subject, array $arguments)
     {
+
         return ($name == 'haveStatusCode' && $subject instanceof ApplicationTester);
     }
-
 
     /**
      * Evaluates positive match.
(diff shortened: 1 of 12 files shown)


ciaranmcnulty:
We should keep the behaviour the same as it is currently when there is no flag in the config or at the CLI, in order to avoid breaking existing CI builds.

You may therefore want to reword the config flags to be a positive case: 'fail on pending' or similar.

toby-griffiths:
But it already fails on pending, does it not? That's why I wanted to add
this as my gulp watcher kept failing for spec's I'd not completed yet.

Please correct me if I'm wrong?

Toby

On Sat, 10 Oct 2015 at 14:41, Ciaran McNulty <notifications@github.com>
wrote:

> We should keep the behaviour the same as it is currently when there is no
> flag in the config or at the CLI, in order to avoid breaking existing CI
> builds.
>
> You may therefore want to reword the config flags to be a positive case:
> 'fail on pending' or similar.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phpspec/phpspec/pull/787#issuecomment-147090131>.
>
-- 

Toby Griffiths
*Cubic Mushroom*
m: 07533 777371
e: toby@cubicmushroom.co.uk  |  skype: toby_griffiths  |  tw: @cubicmushroom
w: http://cubicmushroom.co.uk

-- 

Cubic Mushroom Ltd. is a company registered in England and Wales. 
Registered number: 7891099. Registered office: Unit G9 Castle Cavendish 
Business Centre, Dorking Road, Nottingham NG7 5PN.  VAT Registration 
Number: GB 199 6865 21

toby-griffiths:
Thinking about it, should the runner return different codes based on whether tests are pending/skipped/failed/broken?  Perhaps that's the topic for another PR.

ciaranmcnulty:
No you're correct, I was confusing skipped and pending.

features/bootstrap/Matcher/ApplicationStatusCodeMatcher.php
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Returns matcher priority.
+     *
+     * @return integer
+     */
+    public function getPriority()
+    {
+        return 51;
+    }
+}
```
ciaranmcnulty: Newline here please

src/PhpSpec/Console/ResultConverter.php
```diff
@@ -21,6 +21,23 @@
 class ResultConverter
 {
     /**
+     * @var IO
+     */
+    private $io;
+
+
```
ciaranmcnulty: Extra line break

src/PhpSpec/Console/ResultConverter.php
```diff
@@ -21,6 +21,23 @@
...
+     * ResultConverter constructor.
+     *
+     * @param IO $io
+     */
+    public function __construct(IO $io)
+    {
+        $this->io = $io;
+    }
+
+
```
ciaranmcnulty: Extra line break

src/PhpSpec/Console/ResultConverter.php
```diff
@@ -33,6 +50,12 @@ public function convert($result)
...
                 return 0;
+
+            case ExampleEvent::PENDING:
+                if ($this->io->isIgnorePendingEnabled()) {
+                    return 0;
+                }
+                break;
         }
 
         return 1;
```
ciaranmcnulty: While you're modifying this, it makes sense to me for this do the the `default` of the switch

ciaranmcnulty:
My hesitation now is whether the flag/option specifically for pending is appropriate - the next guy along might want to have Failure passing, etc...

Is there a generic way the same problem can be solved?

stof:
@ciaranmcnulty in Behat, we have a ``--strict`` option to make the suite fail on pending steps (by default, they are just displayed but not considered as failures).
Marking failures as passing should not be in core. It should be left to a phpspec-vw extension :smile:

ciaranmcnulty:
@stof :-)

Yes but 'strict' is pretty vague. I feel like this 'pending-is-passing' is an edge case that if we put one flag in we'll end up with lots in future?

src/PhpSpec/Console/ResultConverter.php
toby-griffiths: Do you mean there is an extra one?  I'm guessing 1 line between variables & method definition then?

ciaranmcnulty: Yes please

features/bootstrap/Matcher/ApplicationStatusCodeMatcher.php
toby-griffiths: Before the return value?  Should that also relate to the supports() method in the same file as well?

ciaranmcnulty: At the end of the file

src/PhpSpec/Console/ResultConverter.php
toby-griffiths: Is this different to the comment above?  Am confused?

ciaranmcnulty: It's at a different point in the file ;-) In general single line between methods please

toby-griffiths:
I'm finding it useful to disable the 'failure' for pending tests during development, but want them to still error during final build checks.  Happy to implement another solution that works in this scenario, if we can agree on one?.

src/PhpSpec/Console/ResultConverter.php
toby-griffiths: Ah, OK.

toby-griffiths:
I guess another possibility is to have different bitwise error codes for each type of failure, and the final error code could AND the bits from all tests… or does that sound like bad practice?

src/PhpSpec/Console/ResultConverter.php
toby-griffiths: If we are keep the pending option (or adding others later), we would then need to ensure that each case statement returns something from each conditional control path if we move the default return into the `default` case.  Leaving outside means less control logic within the switch structure.
Having said that, if each error type was going to return it's own error code, this would make more sense?

ciaranmcnulty:
We currently take the maximum of the test results to be the result of the entire suite, which means there's an explicit ranking of how 'bad' a test result is.

Perhaps the way to model it would be as a threshold?

toby-griffiths:
Perhaps another option would be to define 'environments' or some such in the config file, that defined what is considered a pass, and then use a single command line switch to select which set of rules to use each time?

ciaranmcnulty:
Actually, a question I should have asked at the beginning: what is your reason for treating pending steps as passes?

toby-griffiths:
@ciaranmcnulty I'm finding it useful to disable these 'failures' during development, as my test runner watch tasks would keep telling me I have failures, when actually they're just tests I've not got around to.

Perhaps an alternative solution would be to have different error codes & have gulp (or any other test runner) respond differently to each code?

ciaranmcnulty:
Having considered this, I'm not sure it's a usage I'm keen on supporting in core (having lots of empty examples seems wrong to me). This could be implemented as an extension, or I'd be open to more discussion of returning multiple exit codes.

PR #791, opened 2015-10-16
Add resource name to specification template.

dbrumann:
This PR aims to make the resource name available to the specification-template to allow for:

```
<?php

namespace %namespace%;

use %subject%;
use PhpSpec\ObjectBehavior;
use Prophecy\Argument;

class %name% extends ObjectBehavior
{
    {
        $this->shouldHaveType(%resource%::class);
    }
}
```

### DIFF
--- a/spec/PhpSpec/CodeGenerator/Generator/SpecificationGeneratorSpec.php
+++ b/spec/PhpSpec/CodeGenerator/Generator/SpecificationGeneratorSpec.php
@@ -40,24 +40,26 @@ class SpecificationGeneratorSpec extends ObjectBehavior
     function it_generates_spec_class_from_resource_and_puts_it_into_appropriate_folder(
         $io, $tpl, $fs, ResourceInterface $resource
     ) {
-        $resource->getSpecName()->willReturn('App');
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Is this not achievable using `%namespace%` and `%name%`?

dbr-foodpanda:
No, %name% is the name of the spec. So it will return something like `MyClassSpec` instead of `MyClass`.

ciaranmcnulty:
ok, you can use the placeholder `\%subject%::class` here for the same thing. As such I'm going to close this PR

dbr-foodpanda:
But class will give you the fully qualified class name like `Acme\Domain\MyClass` I just want to have `MyClass` and import the fully qualified class via use-statement.

ciaranmcnulty:
You convinced me ;-)

Please rename it from %resource% to something that reflects it's the short class name

dbr-foodpanda:
Updated PR with renamed template variable `%subject_class%` instead of `%resource%`.

ciaranmcnulty:
Looks good, thanks!

PR #793, opened 2015-10-23
Remove broken example that is actually not needed

ciaranmcnulty:
I believe this example is duplicated by `it_presents_invokable_object_as_string` in the same spec.

@docteurklein I don't suppose you remember the original intention of this spec?

During work on PHP7 support I found this example is actually broken - it passes `$this` to a method that is type hinted as requiring a `Subject`, but our typehint error handling hides the problem.

### DIFF
--- a/spec/PhpSpec/Formatter/Presenter/StringPresenterSpec.php
+++ b/spec/PhpSpec/Formatter/Presenter/StringPresenterSpec.php
@@ -4,6 +4,7 @@ namespace spec\PhpSpec\Formatter\Presenter;
 
 use PhpSpec\ObjectBehavior;
 use PhpSpec\Formatter\Presenter\Differ\Differ;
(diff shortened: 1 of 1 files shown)


docteurklein:
@ciaranmcnulty sorry, I don't remember indeed.

docteurklein:
Here is the commit that introduced it: https://github.com/ciaranmcnulty/phpspec/commit/d1df48fd3b778818eac965291860f3a104f22745

ciaranmcnulty:
@docteurklein Yep; To me it seems that the later examples added duplicate it so it's safe to remove?

docteurklein:
yes, it seems the invokable example is covered, no problem for me to remove it :)

ciaranmcnulty:
Good, that's what I thought ;-)

Issue #792, opened 2015-10-19
Missing documentation how to run specific suite?


PWalkow:
I've got a phpspec.yml defined in repository with all of the suites like:
```
suites:
    nameOne: {
        namespace: Project\Bundle\OneBundle, 
        psr4_prefix: Project\Bundle\OneBundle, 
        spec_path: src/Project\Bundle\OneBundle, 
        src_path: src/Project\Bundle\OneBundle
   }
   nameTwo: {
        namespace: Project\Bundle\TwoBundle, 
        psr4_prefix: Project\Bundle\TwoBundle, 
        spec_path: src/Project\Bundle\TwoBundle, 
        src_path: src/Project\Bundle\TwoBundle
  }
```  
Standard `bin/phpspec run` it runs all of the suites as it should.
How I can run the suite with the name 'nameTwo'?

mheki:
You can run that explicitly with your spec path specified: `phpspec run src/Project/Bundle/TwoBundle`

PWalkow:
@mheki it does not to see any of the specs inside of this bundle.
```
0 specs
0 examples 
0ms
```

mheki:
Sorry, my fault, try adding `spec` at the end of your spec path. 
So `phpspec run src/Project/Bundle/TwoBundle/spec` should do the trick here.

However I can see your point with `-s suitename` option

PWalkow:
@mheki Unfortunately same results (no specs found)

mheki:
There is a small problem in your example. Your spec path is the same as src_path. 
Thanks for reporting, looks like there is an issue in file locating for this particular use case.
I have played with different scenarios locally and found a quick fix for that. 
Will send PR soon.

Saeven:
This would work! :)
https://github.com/phpspec/phpspec/pull/850

PR #794, opened 2015-10-24
Fix running spec with given spec path same as src path

mheki:
This PR tries to solve issue https://github.com/phpspec/phpspec/issues/792 
Basically when user configured `spec_path` to be the same as `src_path` and tried to run `phpspec spec-path` he got a message: 
```
0 specs
0 examples 
0ms
```
while running simply `phpspec` without any argument, specs were run.


ciaranmcnulty:
Thanks for the fix!

Issue #760, opened 2015-08-17
Notice: Undefined index: HOME while running phpspec via Jenkins on Windows Server.


adamelso:
So this is by no means fatal to running phpspec in this setup, I see the successful results of the run command. However if it's worth addressing, I get this PHP notice appear twice: once before and after execution.

> [exec] Notice: Undefined index: HOME in C:\Program Files (x86)\Jenkins\jobs\xxxx-develop\workspace\vendor\phpspec\phpspec\src\PhpSpec\Console\ContainerAssembler.php on line 265

https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Console/ContainerAssembler.php#L279

avant1:
The same notice occurs when running phpspec using gitlab CI multi runner:

> PHP Notice:  Undefined index: HOME in D:\path\to\project\vendor\phpspec\phpspec\src\PhpSpec\Console\Application.php on line 243

The file and line are not the same because my project still uses phpspec 2.0.1 (see https://github.com/phpspec/phpspec/blob/2.0.1/src/PhpSpec/Console/Application.php#L243)

ciaranmcnulty:
This seems solved by #796

adamelso:
:+1:  I can confirm that since updating to 2.4, our Jenkins server no longer complains about this.

ciaranmcnulty:
Great, thanks for confirming

Issue #227, opened 2013-12-08
phpspec on Windows throws Undefined Index notice when running tests


estelsmith:
I've installed commit 73828a76098c88e464ea4006d328eb4bdf2f13d4 via Composer. Any time I run phpspec on Windows, I receive the following notice:

```
PHP Notice:  Undefined index: HOME in {package}\vendor\phpspec\phpspe
c\src\PhpSpec\Console\Application.php on line 207
```

I looked at the file in question, and it seems that phpspec is assuming that $_SERVER['HOME'] exists. This does not seem like a safe assumption.

PR #796, opened 2015-10-26
Use environment variable for home directory, instead of the server superglobal.

adamelso:
This should fix issue #760 (and issue #227) where PR #206 doesn't.

Interestingly, it is already being used by the [Application class](https://github.com/phpspec/phpspec/blob/af867af3105c5292083263eeca9310a14e1f34d4/src/PhpSpec/Console/Application.php#L176).


ciaranmcnulty:
We really need some sort of windows CI coverage

adamelso:
@ciaranmcnulty Way ahead of you ;) #797

PR #797, opened 2015-10-26
Use AppVeyor for continuous integration on Windows

adamelso:
I thought there might be some benefits to testing PhpSpec on Windows. However, most of the failing test cases will most likely involve strict differences in how Windows works to Unix, but not necessarily cause blockers in the application - these being for example line endings or directory separators.

Here's what the output looks like: https://ci.appveyor.com/project/archfizz/phpspec/build/1.0.5/job/99y8hxkfbb42gvrt
If there's value in having this, then I can continue working on it.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,50 @@
+build: false
+shallow_clone: true
+platform: x86
+clone_folder: c:\projects\phpspec
+
+environment:
+    matrix:
+        - PHP_DOWNLOAD_FILE: php-5.6.14-nts-Win32-VC11-x86.zip
+
+matrix:
+    allow_failures:
+        - PHP_DOWNLOAD_FILE: php-5.6.14-nts-Win32-VC11-x86.zip
+
+skip_commits:
+    message: /\[ci skip\]/
+
+cache:
+    - c:\php -> appveyor.yml
+    - '%LOCALAPPDATA%\Composer'
+    - vendor
+
+init:
+    - SET PATH=c:\php;%PATH%
+    - SET COMPOSER_NO_INTERACTION=1
+    - SET PHP=1
+    - SET ANSICON=121x90 (121x90)
+
+install:
+    - IF EXIST c:\php (SET PHP=0) ELSE (mkdir c:\php)
+    - cd c:\php
+    - IF %PHP%==1 appveyor DownloadFile http://windows.php.net/downloads/releases/archives/%PHP_DOWNLOAD_FILE%
+    - IF %PHP%==1 7z x %PHP_DOWNLOAD_FILE% -y > 7z.log
+    - IF %PHP%==1 echo @php %%~dp0composer.phar %%* > composer.bat
+    - appveyor DownloadFile https://getcomposer.org/composer.phar
+    - copy php.ini-production php.ini /Y
+    - echo date.timezone="UTC" >> php.ini
+    - echo extension_dir=ext >> php.ini
+    - echo extension=php_openssl.dll >> php.ini
+    - echo extension=php_curl.dll >> php.ini
+    - echo extension=php_mbstring.dll >> php.ini
+    - echo extension=php_fileinfo.dll >> php.ini
+    - cd c:\projects\phpspec
+    - SET COMPOSER_ROOT_VERSION=dev-master
+    - composer update --no-progress --ansi
+
+test_script:
+    - cd c:\projects\phpspec
+    - php bin\phpspec run --format=pretty
+    - php vendor\phpunit\phpunit\phpunit --testdox
+    - php vendor\behat\behat\bin\behat --format=pretty --tags="~@php-version,@php5.4"


ciaranmcnulty:
This will be great to have; we can have it as an allowed failure for now and then Windows users who want to contribute can work through the bugs :-) I'd be happy to make a Windows issue tag etc.

avant1:
Wow, great work!
I've tried to set up similar script too, but I failed to unzip archive using PowerShell :smile: 

There are mostly errors with directory separators and EOL's (there are some places in code where *\n* is being used).

appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+    - IF %PHP%==1 7z x php-5.5.7-nts-Win32-VC11-x86.zip -y > 7z.log
+    - IF %PHP%==1 echo @php %%~dp0composer.phar %%* > composer.bat
+    - appveyor DownloadFile https://getcomposer.org/composer.phar
+    - copy php.ini-production php.ini /Y
+    - echo date.timezone="UTC" >> php.ini
+    - echo extension_dir=ext >> php.ini
+    - echo extension=php_openssl.dll >> php.ini
+    - echo extension=php_curl.dll >> php.ini
+    - echo extension=php_mbstring.dll >> php.ini
+    - IF %PHP_EXT%==1 echo extension=php_intl.dll >> php.ini
```
stof: PHPSpec does not rely on intl, so doing a test matrix to run it without and with intl is useless

appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+    - SET COMPOSER_ROOT_VERSION=dev-master
+    - composer update --prefer-source --no-interaction --no-progress --ansi
+    - echo ^<^?php if ^(PHP_VERSION_ID ^>^= 50400^) echo ',@php5.4'^; if ^(PHP_VERSION_ID ^>^= 70000^) echo ',@php7'^; > php_version_tags.php
+    - php php_version_tags.php > php_version_tags.txt
+
+test_script:
+    - cd c:\projects\phpspec
+    - php bin\phpspec run --format=pretty
+    - php vendor\bin\phpunit --testdox
+    - set /p PHP_VERSION_TAGS=<php_version_tags.txt
```
stof: this should not be a test script IMO

stof:
@ciaranmcnulty please enable it on the repo (for branches containing the config flag), so that this PR can run on appveyor

ciaranmcnulty:
I think I've configured the web hook.. https://ci.appveyor.com/project/ciaranmcnulty/phpspec

You may need to push to this branch again?

adamelso:
I'm back from holiday (and illness) so I'll continue working on this from today

adamelso:
Not really sure how the `allow_failures` matrix works on appveyor. I was hoping to use `platform` as the condition, where the value is constant but I think only conditions defined in the matrix can be used, ie, the `PHP_DOWNLOAD_FILE` variable that I've defined for each group.

Also, would be a good idea to host the PHP Windows binary archives ourselves rather than relying on [windows.php.net](http://windows.php.net/downloads/releases/)? For example, Phinx uses their own repository on AWS S3. A script may be required to check the lastest version of PHP 7.0 from http://windows.php.net/downloads/snaps/php-7.0/

ciaranmcnulty:
I suspect `windows.php.net` is something we can rely on more than our own solution and the hassles that would involve.

ciaranmcnulty:
This looks good enough to merge and start work on the windows issues :+1: 

How can we configure it so that the windows-failure is visible, but isn't getting on the way of normal deployments?

adamelso:
Alright, it's ready! Except, shall I squash the commits before you merge it?

incidentally, I managed to get Behat to run, but all the scenarios will fail because calling [`chdir()`](https://github.com/phpspec/phpspec/blob/acc41e9588760c09a9bbd8c84fb00bc22f31c3c3/features/bootstrap/FilesystemContext.php#L40) into a directory on Windows prevents that directory [from being removed later](https://github.com/phpspec/phpspec/blob/acc41e9588760c09a9bbd8c84fb00bc22f31c3c3/features/bootstrap/FilesystemContext.php#L54) as pointed out by (Bernhard Schussek)[http://stackoverflow.com/a/33457617/558866].

adamelso:
Also two more issues:

* The `allow_failures` matrix does not result in an overall success when used with the `fast_finish` flag - [possibly a bug](http://help.appveyor.com/discussions/problems/462-allow-failures-behavior-is-problematic).

* The builds overall take nearly an hour to run, most of that time is spent on runnning `composer update`, even though the [Composer cache directory is cached](https://github.com/phpspec/phpspec/pull/797/files#diff-180360612c6b8c4ed830919bbb4dd459R26).

appveyor.yml
```diff
@@ -0,0 +1,64 @@
...
+clone_folder: c:\projects\phpspec
+
+environment:
+    matrix:
+        - PHP_DOWNLOAD_FILE: php-5.3.3-nts-Win32-VC9-x86.zip
+        - PHP_DOWNLOAD_FILE: php-5.3.29-nts-Win32-VC9-x86.zip
+        - PHP_DOWNLOAD_FILE: php-5.4.44-nts-Win32-VC9-x86.zip
+        - PHP_DOWNLOAD_FILE: php-5.5.29-nts-Win32-VC11-x86.zip
+        - PHP_DOWNLOAD_FILE: php-5.6.14-nts-Win32-VC11-x86.zip
+        - PHP_DOWNLOAD_FILE: php-7.0-nts-windows-vc14-x86-r2e38918.zip
```
stof: IMO, you should not test against all PHP versions on appveyor: AppVeyor does not allow to run parallel jobs, which means that all jobs are running one after the other, making the build take a long time. We already cover multiple PHP versions on Travis, so we may not need to do the same on AppVeyor

appveyor.yml
```diff
@@ -0,0 +1,64 @@
...
+    - copy php.ini-production php.ini /Y
+    - echo date.timezone="UTC" >> php.ini
+    - echo extension_dir=ext >> php.ini
+    - echo extension=php_openssl.dll >> php.ini
+    - echo extension=php_curl.dll >> php.ini
+    - echo extension=php_mbstring.dll >> php.ini
+    - echo extension=php_fileinfo.dll >> php.ini
+    - cd c:\projects\phpspec
+    - SET COMPOSER_ROOT_VERSION=dev-master
+    - composer update --prefer-source --no-progress --ansi
```
stof: installing from source forbids composer to use the cache (it caches downloaded archives, but it cannot cache ``git clone``). This makes the cache useless

appveyor.yml
```diff
@@ -0,0 +1,64 @@
...
+        - PHP_DOWNLOAD_FILE: php-5.5.29-nts-Win32-VC11-x86.zip
+        - PHP_DOWNLOAD_FILE: php-5.6.14-nts-Win32-VC11-x86.zip
+        - PHP_DOWNLOAD_FILE: php-7.0-nts-windows-vc14-x86-r2e38918.zip
+
+skip_commits:
+    message: /\[ci skip\]/
+
+cache:
+    - c:\php -> appveyor.yml
+    - '%LOCALAPPDATA%\Composer -> composer.json'
```
stof: no need to invalidate the composer cache on each change in the composer.json. This would force to re-download all archives even if only one of them should change

appveyor.yml
ciaranmcnulty: I agree, this build is about detecting windows issues not issues between PHP versions.

I suggest just checking 5.6.14

ciaranmcnulty:
Thanks for all. The hard work, we don't require squashes although you can if you prefer 

If you can address the comments from @stof it looks like the speed can be increased significantly

adamelso:
Thank you @stof , that's much better!

@ciaranmcnulty AppVeyor is now testing against PHP 5.6 only. Cheers!

appveyor.yml
```diff
@@ -0,0 +1,54 @@
...
+    allow_failures:
+        - PHP_DOWNLOAD_FILE: php-5.6.14-nts-Win32-VC11-x86.zip
+
+skip_commits:
+    message: /\[ci skip\]/
+
+cache:
+    - c:\php -> appveyor.yml
+    - '%LOCALAPPDATA%\Composer'
+    - .phpunit -> phpunit
```
stof: this line is useless (we don't have a ``.phpunit`` folder here. Symfony has it because of its special install of PHPUnit due to the Yaml usage in PHPUnit)

appveyor.yml
```diff
@@ -0,0 +1,54 @@
...
+init:
+    - SET PATH=c:\php;%PATH%
+    - SET COMPOSER_NO_INTERACTION=1
+    - SET PHP=1
+    - SET ANSICON=121x90 (121x90)
+
+install:
+    - IF EXIST c:\php (SET PHP=0) ELSE (mkdir c:\php)
+    - cd c:\php
+    - IF %PHP%==1 (IF %PHP_DOWNLOAD_FILE%==php-7.0-nts-windows-vc14-x86-r2e38918.zip (appveyor DownloadFile http://windows.php.net/downloads/snaps/php-7.0/r2e38918/%PHP_DOWNLOAD_FILE%) ELSE (appveyor DownloadFile http://windows.php.net/downloads/releases/archives/%PHP_DOWNLOAD_FILE%))
```
stof: this can be simplified: ``PHP_DOWNLOAD_FILE`` is never equal to ``php-7.0-nts-windows-vc14-x86-r2e38918.zip``

appveyor.yml
```diff
@@ -0,0 +1,54 @@
...
+    - echo date.timezone="UTC" >> php.ini
+    - echo extension_dir=ext >> php.ini
+    - echo extension=php_openssl.dll >> php.ini
+    - echo extension=php_curl.dll >> php.ini
+    - echo extension=php_mbstring.dll >> php.ini
+    - echo extension=php_fileinfo.dll >> php.ini
+    - cd c:\projects\phpspec
+    - SET COMPOSER_ROOT_VERSION=dev-master
+    - composer update --no-progress --ansi
+    - echo ^<^?php if ^(PHP_VERSION_ID ^>^= 50400^) echo ',@php5.4'^; if ^(PHP_VERSION_ID ^>^= 70000^) echo ',@php7'^; > php_version_tags.php
```
stof: this can be simplified, as we know we are running PHP 5.6

adamelso:
@stof Thanks, updated. The php5.4 tag is passed directly inline now.

ciaranmcnulty:
Great stuff, thanks!

I guess we can start to work through these failures now

PR #800, opened 2015-11-01
Handle case with incomplete use statements

### LINKED ISSUE #799: Typehint namespaces are not resolved correctly when using partial alias
```
namespace Foo;

use Bar\Baz;

class SomethingSpec extends ObjectBehaviour
{
    function it_does_something(Baz\Boz $baz)
    {

    }
}
```

Currently the double is resolved as `Foo\Baz\Boz`, it should be `Bar\Baz\Boz`

ciaranmcnulty:
Closes #799


spec/PhpSpec/CodeAnalysis/TokenizedNamespaceResolverSpec.php
```diff
@@ -80,4 +80,25 @@ class Foo
         $this->resolve('Foo')->shouldReturn('Baz\Foo');
         $this->resolve('Biz')->shouldReturn('Boz\Bar');
     }
+
+    function it_resolves_types_with_partial_use_statements()
+    {$this->analyse('
```
norzechowicz: It looks like ``$this->analyse('`` should be in next line.

PR #803, opened 2015-11-04
Reads from stderr or stdout for new fatal message

Elfiggo:
Isolated Context now reads from ErrorOutput for formatters that do not use the Fatal Presenter and StdOutput for the new Fatal Error message.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,4 +37,4 @@ before_script:
 script:
    - bin/phpspec run --format=pretty
    - ./vendor/bin/phpunit --testdox
-   - ./vendor/bin/behat --format=pretty --tags '~@php-version'`php php_version_tags.php`
+   - ./vendor/bin/behat --format=pretty --strict --tags '~@php-version'`php php_version_tags.php`
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ install:
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
 before_script:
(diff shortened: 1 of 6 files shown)


jon-acker:
Against HHVM the example appears to fail with this:
```
And the class file "src/Message/Fatal/Fatal2.php" contains:                                       # FilesystemContext::theClassOrTraitFileContains()
      """
      <?php
      
      namespace Message\Fatal;
      
      class Fatal2
      {
          public function __construct($param)
          {
              if ($param == 'throw') {
                  throw new \Exception();
              }
          }
      }
      
      """
    When I run phpspec with the "junit" formatter                                                     # IsolatedProcessContext::iRunPhpspecWithThe()
    Then I should see "Call to undefined function"                                                    # IsolatedProcessContext::iShouldSee()
      Did not expect <value>false</value>, but got one. (PhpSpec\Exception\Example\FailureException)
```

ciaranmcnulty:
@Elfiggo this doesn't seem to be passing?

Elfiggo:
Would appreciate a review on this, i'll remove the WIP monicker if happy.

ciaranmcnulty:
You're describing sensible behaviour in the scenarios, but you're not actually testing it. You're directing both outputs into one so we can't tell if the assertion is true.

ciaranmcnulty:
@Elfiggo your scenario still says 'outputs to stderr' but that is not what you are checking. You're checking 'it outputs to stderr or stdout'

ciaranmcnulty:
Thanks!

PR #809, opened 2015-11-25
[WIP] Added new interface and one implementation for discussion

GeeH:
This is literally to provoke discussion on a proposed change.

I've been in discussion with @ciaranmcnulty and @MarcelloDuarte about improving PhpSpec's integration with IDEs, specifically with PhpStorm. PhpStorm uses stub files to tell the IDE what to code complete on when it can't be inferred from the dependency tree, typically this is because of magic methods.

The specific problem I'm looking to address initially is code completion in the spec file on matchers.  In order to do this, we need to be able to programmatically generate a list of matchers that are available at any given time. At the moment I don't see any predictable way of generating this list, some matchers have class constants, some have hard coded arrays etc, so I'm proposing that we add a new method to the `MatcherInterface` that exposes the keywords that this matcher can handle.

This PR gives an example of this new interface method, along with one sample implementation in a matcher.

### DIFF
--- a/src/PhpSpec/Matcher/MatcherInterface.php
+++ b/src/PhpSpec/Matcher/MatcherInterface.php
@@ -50,4 +50,12 @@ interface MatcherInterface
      * @return integer
      */
     public function getPriority();
+
+    /**
+     * Returns an array of matcher names this implementation can provide
+     *
+     * @return array
+     */
+    public function providesMatchers();
+
 }
--- a/src/PhpSpec/Matcher/TypeMatcher.php
+++ b/src/PhpSpec/Matcher/TypeMatcher.php
@@ -96,4 +96,12 @@ class TypeMatcher extends BasicMatcher
             $this->presenter->presentValue($subject)
         ));
     }
+
+    /***
+     * @return array
+     */
+    public function providesMatchers()
+    {
+        return self::$keywords;
+    }
 }


ciaranmcnulty:
This seems sensible, although for BC it will need to be a new interface. Of course, extensions may lag behind in providing support as well.

How would discovery work? PhpSpec only resolves the available matchers at runtime for the spec

GeeH:
Basically, at some point either inside the IDE, or as part of the build
process, a script would run that would resolve all the available matchers,
and run getMatchers() (or similar) on every available matcher, creating an
uber list of all available matchers. It would then generate @method
annotations on a stub file, or if it's in the build process, on the
`ObjectBehavior` class.

Currently I'm working with the IDE team to try and find a way that library
operators can bundle stubs in their own projects, or provide predictable
scripts that the IDE can use to generate it, but it may be that adding the
@methods to the `ObjectBehavior` automatically as part of your build
process is viable.

Anyway, until there's a way to resolve all matcher names it's actually
impossible to do anything :P

G

ciaranmcnulty:
This doesn't seem a bad idea but to retain BC you would need to:

* Make a new interface (that extends the old one) rather than modifying the old one
* Mark the old interface as `@deprecated`
* Update all of the existing matchers to implement the new interface
* Update all of the extensions too

I'm closing this because it's still WIP and am trying to tidy up the PRs but please reopen if you want to tackle the above

PR #812, opened 2015-11-28
Mark arguments as optional

thewilkybarkid:
The arguments on `beConstructedThrough` are optional, so this corrects the PHPDoc.


ciaranmcnulty:
Thanks!

PR #814, opened 2015-11-29
Do not remove typehints not related to specs

### LINKED ISSUE #813: Typehint parsing seems to strip typehints from anonymous class
When running [this](https://github.com/jschreuder/SpotApi/blob/master/spec/Application/Request/HttpRequestParserRouterSpec.php#L64) spec, it errors with the following message:

```
PHP Fatal error:  Declaration of class@anonymous::parseHttpRequest($httpRequest, array $attributes): Spot\Api\Application\Request\Message\RequestInterface must be compatible with Spot\Api\Application\Request\HttpRequestParserInterface::parseHttpRequest(Psr\Http\Message\ServerRequestInterface $httpRequest, array $attributes): Spot\Api\Application\Request\Message\RequestInterface in E:\Sandbox\SpotApi\spec\Application\Request\HttpRequestParserRouterSpec.php on line 64
```

So it looks like the trick used to parse out the typehints from the file also strips the typehint from my anonymous class.

[edit] This is on 2.4.0, when I go back to 2.3.0 it runs fine.

sroze:
This PR aims to fix #813. The rewriter do not removes the typehints of non-spec methods.

It might be interesting to merge the method selection with the one which is in the [ResourceLoader](https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Loader/ResourceLoader.php#L69).

### DIFF
--- a/spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
+++ b/spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
@@ -70,6 +70,31 @@ class TokenizedTypeHintRewriterSpec extends ObjectBehavior
         ');
     }
 
+    function it_removes_typehints_from_single_argument_methods_that_starts_with_its()
+    {
+        $this->rewrite('
+        <?php
+
+        class Foo
+        {
+            public function its_do_bar(\Foo\Bar $bar)
+            {
+            }
+        }
+
+        ')->shouldReturn('
+        <?php
+
+        class Foo
+        {
+            public function its_do_bar($bar)
+            {
+            }
+        }
+
+        ');
+    }
+
     function it_removes_typehints_for_multiple_arguments_in_methods()
     {
         $this->rewrite('
--- a/spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
+++ b/spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
@@ -27,7 +27,7 @@ class TokenizedTypeHintRewriterSpec extends ObjectBehavior
 
         class Foo
         {
-            public function bar()
+            public function it_do_bar()
             {
             }
         }
@@ -37,7 +37,7 @@ class TokenizedTypeHintRewriterSpec extends ObjectBehavior
 
         class Foo
         {
-            public function bar()
+            public function it_do_bar()
             {
             }
         }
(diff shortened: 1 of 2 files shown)


spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
```diff
@@ -27,7 +27,7 @@ function it_leaves_alone_specs_with_no_typehints()
 
         class Foo
         {
-            public function bar()
+            public function it_do_bar()
```
ciaranmcnulty: it_does_bar please

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -171,4 +171,38 @@ private function tokenHasType($token, $type)
...
+     * Returns true if the function should be read with the rewriter.
+     *
+     * @param array $tokens
+     * @param int $index
+     *
+     * @return bool
+     */
+    private function functionShouldBeRead(array $tokens, $index)
+    {
+        return preg_match('/^(it|its)[^a-zA-Z]/', $this->getFunctionName($tokens, $index));
```
ciaranmcnulty: this duplicates logic elsewhere

sroze: Exactly here: https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Loader/ResourceLoader.php#L69

The "problem" is that just isolating this regex in a method might be too much.

sroze:
Closing in favor of #815.

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
jon-acker: could you put this regex in a constant somewhere then?

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -102,7 +102,7 @@ private function stripTypeHints($tokens)
                     if ($this->tokenHasType($token, T_STRING) && !$this->currentClass) {
                         $this->currentClass = $token[1];
                     }
-                    elseif($this->tokenHasType($token, T_FUNCTION)) {
+                    elseif($this->tokenHasType($token, T_FUNCTION) && $this->functionShouldBeRead($tokens, $index)) {
```
jon-acker: would it make more sense, instead of having this exception called functionShouldBeRead() to introduce a new state e.g. STATE_READING_EXAMPLE which gets ignored? that way it would be clear that this exception is about typehints in spec examples.

PR #815, opened 2015-11-29
Do not remove typehints inside a method definition

sroze:
This PR fixes #814 by simply not removing typehints found inside a method definition.

### DIFF
--- a/spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
+++ b/spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
@@ -70,6 +70,39 @@ class TokenizedTypeHintRewriterSpec extends ObjectBehavior
         ');
     }
 
+    function it_do_not_remove_typehints_in_methods()
+    {
+        $this->rewrite('
+        <?php
+
+        class Foo
+        {
+            public function bar(\Foo\Bar $bar)
+            {
+                new class($argument) implements InterfaceName
+                {
+                    public function foo(Foo $foo) {}
+                };
+            }
+        }
+
+        ')->shouldReturn('
+        <?php
+
+        class Foo
+        {
+            public function bar($bar)
+            {
+                new class($argument) implements InterfaceName
+                {
+                    public function foo(Foo $foo) {}
+                };
+            }
+        }
+
+        ');
+    }
+
     function it_removes_typehints_for_multiple_arguments_in_methods()
     {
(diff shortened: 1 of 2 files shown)


stof:
Tests are broken on PHP 7

sroze:
@stof yop yop I'll have a look today

sroze:
@ciaranmcnulty :tada:

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -99,13 +99,34 @@ private function stripTypeHints($tokens)
...
                     }
                     break;
+                case self::STATE_READING_FUNCTION_BODY:
+                    if ('{' == $token) {
+                        $this->currentBodyLevel++;
+                    }
+                    elseif ('}' == $token) {
+                        $this->currentBodyLevel--;
+
+                        if ($this->currentBodyLevel == 0) {
```
stof: this should use a strict comparison

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -99,13 +99,34 @@ private function stripTypeHints($tokens)
...
+                            $this->currentFunction = '';
+                        }
+                    }
+
+                    if (!$this->currentFunction) {
+                        $this->state = self::STATE_READING_CLASS;
```
stof: this state change should happen in the ``$this->currentBodyLevel == 0`` condition IMO. This is where you know you are leaving the function

sroze:
Makes sense, addressed your points @stof.

spec/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriterSpec.php
```diff
@@ -70,6 +70,39 @@ public function bar($bar)
         ');
     }
 
+    function it_do_not_remove_typehints_in_methods()
```
jon-acker: do -> does

ciaranmcnulty: Just this to change from what I can see

sroze:
@ciaranmcnulty I fixed the typo reported by @jon-acker :)

ciaranmcnulty:
Thanks, @sroze!

PR #818, opened 2015-12-04
Fix re-runner on Windows

### LINKED ISSUE #773: Windows: 'PHPSPEC_EXECUTION_CONTEXT' is not recognized as an internal or external command, operable command or batch file
### The issue

On Windows systems, after succesfully auto-generating code, this error message is shown on the command prompt.

> 'PHPSPEC_EXECUTION_CONTEXT' is not recognized as an internal or external command, operable program or batch file.

I have tested this on my Windows 10 machine and a fresh Windows 7 VirtualBox installation, with only php 5.6.13 and composer installed.

phpspec_execution_context

On an Ubuntu system, the message doesn't appear.

### Versions affected

PhpSpec 2.3.0

### Versions unaffected

PhpSpec 2.2.1

### Steps to reproduce

1. Get a Windows machine with php/composer installed.

2. Create an empty folder.

3. `composer require phpspec/phpspec`

4. Set psr-0 autoloading, `composer dumpautoload`, etc.

5. `phpspec describe Test`

6. `phpspec run`

7. Test will fail, PhpSpec will offer to create the class for the user. Answer Y.

8. PspSpec will create the class, but the above mentioned error will be shown.

Same error message when creating a method.

### My findings

Keep in mind that I'm a noobish programmer, and this codebase is overwhelming to me.

What I've found is that an afterSuite event is fired, and ClassNotFoundListener will mark
the test as "worth rerunning". Then, ReRunListener afterSuite() will be called, will check
if the test is worth rerunning, and then rerun the suite. From there, it will end in the
PassThruRerunner reRunSuite() method, which will build a string to rerun the suite. And there
lies the problem: to build the command, it calls a buildArgString() method. This method gets
an executionContext of JsonExecutionContext (the only implementation of the interface AFAIK), and 
asEnv() returns an array, being the key the infamous 'PHPSPEC_EXECUTION_CONTEXT' that confuses Windows, and a Json string. So the method returns this long string…

`PHPSPEC_EXECUTION_CONTEXT="{ generated-types :[ Test ]}"`

…which then gets concatenated to the php executable path, and the arguments, and that is executed by the system via passthru(). Windows doesn't like that.

Now, if I remove the `$this->buildArgString()` part of the `$command` string, everything seems to work fine, but it's obviously there for a reason, so I guess the problem lies elsewhere, but I don't know how to continue from there.
 
Which is where my noobishness shows, and the reason why this is an issue and not a pull request :)

### Questions

Now, questions:

1. Is someone else able to reproduce this, or have I a botched php/phpspec installation or whatever? I'm surprised nobody has complained already.

2. PhpSpec works fine, the error message is annoying but I can live with it. But the question is:
Do I keep trying to understand what's wrong, or maybe someone more clever than me already has the answer?

3. Should I start using Unix-like systems for coding and not even noticing errors like these and live a happier life?

I'll be glad to help with whatever you may need from me. Thanks!

chrisemerson:
Create a new WindowsPassthruRerunner that fixes the command line on windows & ensure it gets used on windows instead of the standard PassthruRerunner. Fixes #773 

The escapeshellarg and escapeshellcmd calls are removed from the new Windows re-runner as they tend to mangle the JSON and the command on windows. I'm not too sure about doing this - but all the data they escape comes from internal processes which makes me a bit more comfortable about it - though they may not be acceptable of course, and a different solution may be needed.

I had a compounded issue with the fact that there is a space in the path to my PHP executable, which this also fixes.

### DIFF
--- a/src/PhpSpec/Process/ReRunner/PassthruReRunner.php
+++ b/src/PhpSpec/Process/ReRunner/PassthruReRunner.php
@@ -44,7 +44,7 @@ class PassthruReRunner extends PhpExecutableReRunner
         return (php_sapi_name() == 'cli')
             && $this->getExecutablePath()
             && function_exists('passthru')
-            && strtolower(substr(PHP_OS, 0, 3)) != "win";
+            && (stripos(PHP_OS, "win") !== 0);
     }
 
     public function reRunSuite()
--- a/src/PhpSpec/Process/ReRunner/WindowsPassthruReRunner.php
+++ b/src/PhpSpec/Process/ReRunner/WindowsPassthruReRunner.php
@@ -44,7 +44,7 @@ class WindowsPassthruReRunner extends PhpExecutableReRunner
         return (php_sapi_name() == 'cli')
             && $this->getExecutablePath()
             && function_exists('passthru')
-            && strtolower(substr(PHP_OS, 0, 3)) == "win";
+            && (stripos(PHP_OS, "win") === 0);
     }
 
     public function reRunSuite()
--- a/src/PhpSpec/Process/ReRunner/WindowsPassthruReRunner.php
+++ b/src/PhpSpec/Process/ReRunner/WindowsPassthruReRunner.php
@@ -16,7 +16,7 @@ namespace PhpSpec\Process\ReRunner;
 use PhpSpec\Process\Context\ExecutionContextInterface;
 use Symfony\Component\Process\PhpExecutableFinder;
 
-class PassthruReRunner extends PhpExecutableReRunner
+class WindowsPassthruReRunner extends PhpExecutableReRunner
 {
     /**
      * @var ExecutionContextInterface
@@ -44,13 +44,14 @@ class PassthruReRunner extends PhpExecutableReRunner
         return (php_sapi_name() == 'cli')
             && $this->getExecutablePath()
             && function_exists('passthru')
-            && strtolower(substr(PHP_OS, 0, 3)) != "win";
+            && strtolower(substr(PHP_OS, 0, 3)) == "win";
     }
 
     public function reRunSuite()
     {
         $args = $_SERVER['argv'];
-        $command = $this->buildArgString() . escapeshellcmd($this->getExecutablePath()).' '.join(' ', array_map('escapeshellarg', $args));
+        $command = $this->buildArgString() . '"' . $this->getExecutablePath().'" '.join(' ', array_map('escapeshellarg', $args));
+
         passthru($command, $exitCode);
         exit($exitCode);
     }
@@ -60,7 +61,7 @@ class PassthruReRunner extends PhpExecutableReRunner
         $argstring = '';
 
         foreach ($this->executionContext->asEnv() as $key => $value) {
-            $argstring .= $key . '=' . escapeshellarg($value) . ' ';
+            $argstring .= 'SET ' . $key . '=' . $value . ' && ';
         }
 
         return $argstring;
--- a/spec/PhpSpec/Process/ReRunner/WindowsPassthruRerunnerSpec.php
+++ b/spec/PhpSpec/Process/ReRunner/WindowsPassthruRerunnerSpec.php
@@ -1,0 +1,39 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace spec\PhpSpec\Process\ReRunner;
+
+use PhpSpec\ObjectBehavior;
+use PhpSpec\Process\Context\ExecutionContextInterface;
+use Prophecy\Argument;
+use Symfony\Component\Process\PhpExecutableFinder;
+
+class WindowsPassthruRerunnerSpec extends ObjectBehavior
+{
+    function let(PhpExecutableFinder $executableFinder, ExecutionContextInterface $executionContext)
+    {
+        $this->beConstructedThrough('withExecutionContext', array($executableFinder, $executionContext));
+    }
+
+    function it_is_a_rerunner()
+    {
+        $this->shouldHaveType('PhpSpec\Process\ReRunner');
+    }
+
+    function it_is_not_supported_when_php_process_is_not_found(PhpExecutableFinder $executableFinder)
+    {
+        $executableFinder->find()->willReturn(false);
+
+        $this->isSupported()->shouldReturn(false);
+    }
+}
--- a/src/PhpSpec/Console/ContainerAssembler.php
+++ b/src/PhpSpec/Console/ContainerAssembler.php
@@ -669,6 +669,12 @@ class ContainerAssembler
                 $c->get('process.executioncontext')
             );
         });
+        $container->setShared('process.rerunner.platformspecific.windowspassthru', function (ServiceContainer $c) {
+            return ReRunner\WindowsPassthruReRunner::withExecutionContext(
+                $c->get('process.phpexecutablefinder'),
+                $c->get('process.executioncontext')
+            );
+        });
         $container->setShared('process.phpexecutablefinder', function () {
             return new PhpExecutableFinder();
         });


src/PhpSpec/Process/ReRunner/PassthruReRunner.php
```diff
@@ -43,7 +43,8 @@ public function isSupported()
     {
         return (php_sapi_name() == 'cli')
             && $this->getExecutablePath()
-            && function_exists('passthru');
+            && function_exists('passthru')
+            && strtolower(substr(PHP_OS, 0, 3)) != "win";
```
ciaranmcnulty: I think we use something from $_SERVER to detect windows in other places I think?

chrisemerson: OK, I'll see if I can find other examples and change this to match.

chrisemerson: I've had a search through the codebase but can't find any other places you are detecting windows any ideas where it might be?

ciaranmcnulty: I guess I was thinking of this https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Console/ContainerAssembler.php#L303-L307

ciaranmcnulty:
I'd rather there was a way to fix the existing rerunner, but this seems to be a decent approach. I don't have a way to verify that it works right now, if someone could independently check on a Windows box it would be a big help.

src/PhpSpec/Process/ReRunner/PassthruReRunner.php
ciaranmcnulty: ok so i'd rather you use `stripos` here

src/PhpSpec/Process/ReRunner/WindowsPassthruReRunner.php
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * @return boolean
+     */
+    public function isSupported()
+    {
+        return (php_sapi_name() == 'cli')
+            && $this->getExecutablePath()
+            && function_exists('passthru')
+            && strtolower(substr(PHP_OS, 0, 3)) == "win";
```
ciaranmcnulty: and `stripos` here please

chrisemerson:
Unfortunately, windows handles environment variables in a different enough way to unix based OS's that I think to fix in the existing rerunner would create a fair amount of mess - you have to call 'set' as a command, separate commands with && and not put the values in quotes (or they become part of the environment var itself).

I'll fix the other couple of items you've mentioned.

ciaranmcnulty:
Decided to trust @chrisemerson as it's not causing any bad effects on non-Windows platforms ;-)

Thanks!

src/PhpSpec/Process/ReRunner/WindowsPassthruReRunner.php
```diff
@@ -0,0 +1,69 @@
...
+        return (php_sapi_name() == 'cli')
+            && $this->getExecutablePath()
+            && function_exists('passthru')
+            && (stripos(PHP_OS, "win") === 0);
+    }
+
+    public function reRunSuite()
+    {
+        $args = $_SERVER['argv'];
+        $command = $this->buildArgString() . '"' . $this->getExecutablePath().'" '.join(' ', array_map('escapeshellarg', $args));
```
stof: the executable path needs to be escaped to account for cases where there are spaces in it.
And ``C:\Program Files\PHP\php.exe`` has a space in it, so things would break if people install PHP in ``C:\Program Files`` (which is an expected location when following the windows conventions)

ciaranmcnulty: @stof Chris seemed to think this fixed that problem for him?

stof: @ciaranmcnulty maybe he does not have spaces in his PHP executable path, meaning he is not affected by the issue I reported.
If removing the escaping solved an issue for him, it means there might be *another* issue which disappears when removing escaping. But this needs to be investigated then. Saying ``all the data they escape comes from internal processes which makes me a bit more comfortable about it`` is wrong:

- the PHP path does not come from internal stuff (it depends on the user setup)
- the escaping is here because of the needs of the shell interpreter, not just because of accepting untrusted input

ciaranmcnulty: I refer to his comment "I had a compounded issue with the fact that there is a space in the path to my PHP executable, which this also fixes." but you are right to be concerned.

Currently this feature doesn't work at all on Windows so at least this PR would make it work for the one person I know who uses windows (@chrisemerson)

chrisemerson: The quote marks either side of the call to the executable sort this out, though I am working on the assumption that the output of the getExecutablePath() method is 'clean' and doesn't have anything within it that needs escaping itself. The PHP escape functions cause the command not to run under windows for some reason. My PHP executable is in Program Files and this change works with it.

stof: this looks suspicious to me, given that spaces need escaping. Can you check what is returning by ``getExecutablePath`` ?

ciaranmcnulty: I'll hold off releasing `2.4.1` until this is resolved

chrisemerson: You don't need to escape spaces within the double quotes - the only issue would occur if getExecutablePath returned something with a double quote mark within it.

stof: ah I missed that you are applying double quotes manually here, which means doing part of the escaping by hand (I'm saying "part of it", because other chars may need escaping too on Windows, but PHP itself does not handle it totally right: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Process/ProcessUtils.php#L39)

What I don't understand is why you are applying quotes by hand instead of using the escaping (you should indeed not do both)

chrisemerson: The escaping itself just didn't work. I did try this first but had no luck trying several different things.

stof: well, this is probably related to the fact that the existing passthru rerunner uses ``escapeshellcmd``, which is *not* meant to escape a single part of the command, and so does not enclose values in quotes when they contain spaces. ``escapeshellarg`` should probably be used instead.

chrisemerson: OK, perhaps it's my misunderstanding of these functions then - do you want me to look further into it and PR again if I get it working using the escapeshellarg functions?

garethellis36:
I'm on Windows 7 and have just started a mini project to try out PHPSpec. I had the same issue which prompted this PR so I included dev-master in composer. `phpspec run` (after `phpspec desc`) results in the yellow "class doesn't exist" message, along with the prompt to create it. If I enter 'Y', the class is generated. However, the process seems to hang at the `passthru()` point (`WindowsPassthruReRunner::reRunSuite()` line 55). If I do CTRL-C out and `phpspec run` again, it still doesn't recognize that the class exists and I'm just stuck in a loop.

ciaranmcnulty:
@garethellis36 The last part of your comment is probably due to a misconfigured autoloader.

ciaranmcnulty:
I'm tempted at this point to remove rerunning from the Windows version until a fix can be found.

garethellis36:
@ciaranmcnulty Yes you're right there. My composer.json wasn't set-up right - thanks!

chrisemerson:
@ciaranmcnulty A fix for what? Or just that you want the command parts escaped differently? I'm happy to look into that further, just awaiting some feedback on whether you want me to or not.

ciaranmcnulty:
@chrisemerson Sorry didn't realise you were waiting for affirmation. If you can get it working with escaping that would be great (and there may be a fix you can apply to the existing rerunner too)

chrisemerson:
I'll take a look at it this evening

chrisemerson:
Apologies for delay - I've changed this now, used escapeshellarg() on the path instead of manually putting the quotes in there. I'll open a new PR.

Issue #816, opened 2015-11-30
Silent loader failing in 2.4


tomphp:
I've just installed 2.4 and found that if you have a supporting class defined in a spec file, and the supporting class either implements an undefined interface or doesn't correctly define the methods in an interface it implements - then the phpspec loader fails silently and phpspec exits with a 255 exit code.

Example: Including the following spec in your test suite will cause phpspec to fail silently and now run any tests:

```php
<?php

namespace spec;

use PhpSpec\ObjectBehavior;
use Prophecy\Argument;

class SomethingSpec extends ObjectBehavior
{
    function it_is_initializable()
    {
        $this->shouldHaveType('Something');
    }
}

class ExampleClass implements NotDefinedInterface
{
}
```

sroze:
@tomphp are you running the suite with PHP 5 or 7?

tomphp:
@sroze 7 but I've tried 5 and had the same response.

Elfiggo:
@tomphp @sroze Thanks for reporting it, I will take a look.

Elfiggo:
Hi @tomphp would you be able to add your error_reporting level here?
It will help with the investigation?

tomphp:
@Elfiggo it's currently set to `E_ALL`. Are you not able to reproduce? I can try and make some time to take a look at it myself if you're not able to recreate the same behaviour.

Elfiggo:
@tomphp I have been able to reproduce.

I have used your code for the Behat scenario to prove it works. I do see the error if I use a different formatter such as the "junit" one.

The fix is in this PR https://github.com/phpspec/phpspec/pull/819

Thanks for getting back to me.

tomphp:
@Elfiggo: Awesome! Thanks for taking the time to investigate and fix!

ciaranmcnulty:
As 819 is now merged I'll close.

PR #819, opened 2015-12-07
Silent loader failing in 2.4 in relation to #816

Elfiggo:
Fix for issue #816 

Please take a moment to verify.

As the fatal error was not inside of a spec, it was not able to display the error.

I have changed the condition so if the issue is inside of a spec file BUT not inside of a spec, then you will still see the fatal error message, as long as it is a Fatal Error.


features/exception_handling/developer_defines_supporting_spec.feature
```diff
@@ -0,0 +1,46 @@
...
+          {
+              if ($param == 'throw') {
+                  throw new \Exception();
+              }
+          }
+      }
+
+      """
+    When I run phpspec
+    Then I should see "Fatal error happened while executing the following"
```
ciaranmcnulty: Line break here please

PR #824, opened 2015-12-15
Improved method for finding the last method of a class. Fixes #823.

### LINKED ISSUE #823: Issue generating methods when the last method contains the keyword "function"
If the class I'm spec-ing contains an anonymous function in its last method then the code generation fails. This is because the the ClassFileAnalyser was looking for the last use of the token `T_FUNCTION`. Example file:
```
<?php
class MyClass
{
    public function testAnAnonymousFunction()
    {
        return function () {
            return 'something';
        };
    }
}
```

Error:
```
 [PhpSpec\Exception\Example\ErrorException]                                                                                             
  notice: Uninitialized string offset: 2 in /vendor/phpspec/phpspec/src/PhpSpec/Util/ClassFileAnalyser.php line 41
```

petemcfarlane:


src/PhpSpec/Util/ClassFileAnalyser.php
```diff
@@ -232,4 +220,32 @@ private function tokenIsFunction($token)
...
+     * @param array $tokens
+     * @param int $index
+     * @return int
+     */
+    public function findEndOfLastMethod(array $tokens, $index)
+    {
+        for ($i = $index - 1; $i > 0; $i--) {
+            if ($tokens[$i] == "}") {
+                return $i + 1;
+                continue;
```
sroze: I'd argue that a `continue` after a `return` is useless :stuck_out_tongue:

Elfiggo:
Looks good.

ciaranmcnulty:
Thanks!

PR #826, opened 2015-12-18
Fix html formatter

bnowack:
Hi,

The "--format html" option seems to be broken at the moment, due to a missing Exception reference and because the HTML Presenter is not returned by the assembler.

Cheers,
Benji


shanethehat:
:+1:

ciaranmcnulty:
Thanks for the fix!

PR #828, opened 2015-12-21
Remove interface name suffixes

shanethehat:
*This PR breaks BC*

All `*Interface` interface names are renamed to something reflecting their purpose without the `Interface` suffix.


stof:
Well, I don't think bumping to PhpSpec 3 just for the removal of the Interface suffix is worth it. It would break all the ecosystem of extensions (and without a way to support both 2.x and 3.x together)

shanethehat:
Yeah, I agree. It seemed like a fun idea, but this is missing a step of deprecating all the old interfaces first so that the extensions have a chance to adapt. I'll make a less destructive PR.

ciaranmcnulty:
I think this is a good step but needs to be done with deprecation first. I'll close but please keep thinking about it :-)

I'm going to publish a release roadmap soon (probably Symfony-style date constrained) so it will be clearer when and how BC-breaking changes can be introduced

ciaranmcnulty:
@shanethehat This is suitable for 3.0 so I'll reopen

ciaranmcnulty:
@shanethehat can you resolve the merge conflicts please?

GrahamCampbell:
Probably could do with being squashed to one commit too.

shanethehat:
@ciaranmcnulty conflicts fixed, expect weirdness removed, history squashed

stof:
@ciaranmcnulty that's not suitable as is though, as we should introduce it first with a BC layer, and then remove the BC layer only after that.

ciaranmcnulty:
@stof Currently I don't mind introducing breaking changes in 3.x

stof:
@ciaranmcnulty it is better to provide an easy upgrade path. Otherwise, the migration is much harder.
Deprecations should at least be considered for things impacting users of phpspec (opposed to extension authors)

ciaranmcnulty:
Are you talking about end users, or extension writers? I think they're different cases.

You have made me realise that `MatcherProviderInterface` and so on might need documenting.

I don't have a problem making small breaks that extension writers have to catch up with. It worked for Behat. We don't have that big an ecosystem that it would cause the problems / require the sort of solutions that a project like Symfony has.

stof:
@ciaranmcnulty I'm OK with BC breaks without BC layers for extension writers (especially given that extension APIs are not really documented).
But anything impacting spec authors (i.e. end users) should introduce the new interfaces and deprecate the old ones in 2.next, before removing the old ones in 3.0.

> You have made me realise that `MatcherProviderInterface` and so on might need documenting.

good point, but should be reported separately (and not waiting 3.0)

ciaranmcnulty:
@stof there is no `2.next`, `3.0` is the next version.

We agree about extension writers. I hadn't considered the impact this PR has on user land.

The options are:

- Say 'implement the new interface' and document it somewhere clear. (similar to the Behat 2->3)
- Leave the naming inconsistent

We should decide this now because it will inform a lot of our other decisions as we move towards 3.0.

I'm leaning towards breaking things in the name of improvement

stof:
@ciaranmcnulty to provide an easy migration path, there should be a 2.next allowing to use new names for user-land APIs.

The Behat 2->3 upgrade is not similar here, as Behat 3 was a rewrite from scratch, and so could not provide such easy migration path.

I'm fine with changing names for consistencies, if we make the migration path easy (and shipping a 2.next for that would not be that hard IMO)

features/extensions/developer_uses_extension.feature
```diff
@@ -14,10 +14,10 @@ Feature: Developer uses extension
 
     namespace Example1\PhpSpec\MatcherExtension;
 
-    use PhpSpec\Extension\ExtensionInterface;
+    use PhpSpec\Extension\PhpSpecExtension;
     use PhpSpec\ServiceContainer;
 
-    class Extension implements ExtensionInterface
+    class Extension implements PhpSpecExtension
```
ciaranmcnulty: TBH this could move up to `PhpSpec\Extension`

src/PhpSpec/Console/IO.php
```diff
@@ -13,15 +13,15 @@
...
+use PhpSpec\IO\IOStream;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use PhpSpec\Config\OptionsConfig;
 
 /**
  * Class IO deals with input and output from command line interaction
  */
-class IO implements IOInterface
+class IO implements IOStream
```
ciaranmcnulty: Not sure about Stream here

ciaranmcnulty:
@shanethehat made a couple of comments and looks like you need to rebase

src/PhpSpec/Console/IO.php
shanethehat: The interface should probably be `IO`, which I guess leaves the question of what is noteworthy about the implementation.

shanethehat:
@ciaranmcnulty rebased and tidied up

ciaranmcnulty:
Thanks for the efforts @shanethehat!

PR #829, opened 2015-12-22
Fix windows failures

shanethehat:


spec/PhpSpec/Formatter/Presenter/Differ/StringEngineSpec.php
```diff
@@ -26,6 +26,6 @@ function it_calculates_strings_diff()
 </code>
 DIFF;
 
-        $this->compare('string1', 'string2')->shouldReturn(str_replace("\n", PHP_EOL, $expected));
+        $this->compare('string1', 'string2')->shouldReturn($expected);
```
stof: why this change ?

spec/PhpSpec/Formatter/Presenter/StringPresenterSpec.php
```diff
@@ -26,7 +26,7 @@ function it_presents_long_string_in_quotes_but_trimmed()
 
     function it_presents_only_first_line_of_multiline_string()
     {
-        $this->presentValue("some\nmultiline\nvalue")->shouldReturn('"some..."');
+        $this->presentValue("some".PHP_EOL."multiline".PHP_EOL."value")->shouldReturn('"some..."');
```
stof: changing this is wrong. We want to test explicitly what happens when the string contains a ``\n``, even on Windows

spec/PhpSpec/Formatter/Presenter/Value/TruncatingStringTypePresenterSpec.php
```diff
@@ -42,6 +42,6 @@ function it_should_return_long_values_truncated(
     function it_presents_only_first_line_of_multiline_string(StringTypePresenter $stringTypePresenter)
     {
         $stringTypePresenter->present('some...')->willReturn('some...');
-        $this->present("some\nmultiline\nvalue")->shouldReturn('some...');
+        $this->present("some".PHP_EOL."multiline".PHP_EOL."value")->shouldReturn('some...');
```
stof: same here

src/PhpSpec/CodeGenerator/Generator/ClassGenerator.php
```diff
@@ -54,7 +54,7 @@ protected function renderTemplate(ResourceInterface $resource, $filepath)
             '%name%'            => $resource->getName(),
             '%namespace%'       => $resource->getSrcNamespace(),
             '%namespace_block%' => '' !== $resource->getSrcNamespace()
-                                ?  sprintf("\n\nnamespace %s;", $resource->getSrcNamespace())
+                                ?  sprintf("%snamespace %s;", PHP_EOL.PHP_EOL, $resource->getSrcNamespace())
```
stof: -1 for this. Almost OSS projects are using LF newlines in their source code, not the OS newlines (as they cannot ship OS-specific source code). this would make things hard for any windows contributor as it would inject CRLF here (I don't know any big PHP project choosing CRLF as their consistent line endings)

src/PhpSpec/Console/IO.php
```diff
@@ -140,6 +140,7 @@ public function getLastWrittenMessage()
      */
     public function writeln($message = '', $indent = null)
     {
+        //var_dump($message);
```
stof: must be reverted

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
```diff
@@ -123,8 +123,8 @@ public function getPriority()
     private function generateCollaborators(CollaboratorManager $collaborators, \ReflectionFunctionAbstract $function, \ReflectionClass $classRefl)
     {
         if ($comment = $function->getDocComment()) {
-            $comment = str_replace("\r\n", "\n", $comment);
-            foreach (explode("\n", trim($comment)) as $line) {
+            //$comment = str_replace("\r\n", "\n", $comment);
+            foreach (explode(PHP_EOL, trim($comment)) as $line) {
```
stof: this means that any testsuite using LF line endings would now break on windows. It does not make any sense.

stof:
I think most of the changes you are doing are related to the fact that you are not running the PhpSpec testsuite itself, but a codebase in which you changed all line endings locally. In this case, the testsuite might indeed not work, but this is quite expected as you are not running the original codebase.
I strongly suggest you to disable the conversion of LF to CRLF on checkout.

src/PhpSpec/CodeGenerator/Generator/ClassGenerator.php
GrahamCampbell: I agree with @stof. There's no way we should be using the OS specific newline character.

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
GrahamCampbell: all these changes are wrong, yeh

GrahamCampbell:
:-1: on this entire PR

shanethehat:
@stof yeah, I take your point. This is for the most part an ill-considered solution to the Windows CI platform cloning the repo with Windows line endings. 

I'll undo all the line ending stuff, although I think the third comment about distinguishing a Windows path from a FQCN is still needed. I'll then look into making Appveyor clone the repo with Unix line endings.

shanethehat:
s/comment/commit/

shanethehat:
All of the test failures related to the PSR0Locator are now fixed, without using system line endings. The remaining test failures appear to be related to the code generation part of the application. I no longer have access to a Windows platform to work on those until the new year.

I do think that we should document somewhere that the tool expects developers on Windows to use Unix style line endings. (I don't use Windows myself, so not sure how much of a convention that is).

src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -372,4 +373,31 @@ private function getQueryPath($query)
 
         return rtrim(realpath($replacedQuery), $sepr);
     }
+
+    /**
+     * @param $query
```
GrahamCampbell: invalid phpdoc

src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -372,4 +373,31 @@ private function getQueryPath($query)
...
+     * @return bool
+     */
+    private function queryContainsQualifiedClassName($query)
+    {
+        return $this->queryContainsBlackslashes($query) && !$this->isWindowsPath($query);
+    }
+
+    /**
+     * @param $query
```
GrahamCampbell: same

src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -372,4 +373,31 @@ private function getQueryPath($query)
...
+     * @return bool
+     */
+    private function queryContainsBlackslashes($query)
+    {
+        return false !== strpos($query, '\\');
+    }
+
+    /**
+     * @param $query
```
GrahamCampbell: etc

ciaranmcnulty:
Thanks @shanethehat this looks much less scary than the original version

PR #830, opened 2015-12-23
Windows re-runner improvements

chrisemerson:
Use escapeshellarg on PHP path in Windows re-runner


ciaranmcnulty:
Thanks for keeping on this

Issue #827, opened 2015-12-21
Windows: first automatically generated method is not inserted in the correct place


garethellis36:
Windows 7 Professional x64
PHP 5.6.13 NTS with Xdebug 2.3.2
PHPSpec 2.4.0

All commands run from Git Bash, but the same problem occurs with cmd.exe.

The first method generated automatically by PhpSpec is placed after the closing brace for the class declaration. This doesn't happen on OS X so I assume it is specific to either Windows or my set-up on my Windows laptop.

How to reproduce:
`phpspec desc Day15/Ingredient`
`phpspec run` - press 'Y' to create `Ingredient` class

Result (one empty line at EOF):
```php 
<?php

namespace Day15;

class Ingredient
{
}

```

Add anything to the spec file which prompts the missing method message - it does this with the constructor or with a method, as long as it's the first method and the Ingredient class is untouched. For example:
```php
public function let()
{
     $this->beConstructedWith('foo', 'bar', 'baz', 'bat');
}
```

Then:
`phpspec run` - press 'Y' to add constructor

Result:
```php
<?php

namespace Day15;

class Ingredient
{
}

    public function __construct($argument1, $argument2, $argument3, $argument4)
    {
        // TODO: write logic here
    }
```

chrisemerson:
This is odd. I get this issue with the 2nd onwards, but the first is fine. I was about to look at this.

garethellis36:
I had a look at the method generator code but to be honest it's a little above my head :-) I wondered if it's something to do with line break characters.

chrisemerson:
The files generated all use LF so there should be no difference there.

chrisemerson:
Following your example, the code generated fine for me (whether it's the constructor or something else). Adding a 2nd method onwards though for me generates these outside the class. It's a problem I've seen quite a few times before, yet on other windows machines it works fine, and you have a different experience again! I'd love to try and get to the bottom of it.

chrisemerson:
I've found (and fixed) the issue - PR to follow. I just want to make sure it always works still. It was indeed to do with line endings, but it looks like it depends how you have your git set up - whether it checks files out as-is or auto-converts them to CRLF (I personally check out as-is, commit as LF, and have my IDE only ever use LF, as per PSR-2 standards).

garethellis36:
I wondered whether it was something to do with Git - I believe I have mine defaulting to CRLF due to legacy stuff. Nice one :+1:

chrisemerson:
[...]

PR #831, opened 2015-12-23
Fix code generation issue on Windows

chrisemerson:
The 'insertStringAfterLine' method first splits the input on the PHP_EOL constant. The trouble with this is that it won't work on systems where git is set up to clone line endings as-is, as PHP_EOL will still evaluate to \r\n and nothing will be split. The method then ends up being inserted at the end of the file, outside the class.

Exploding on "\n" will cover all line-ending cases (though it may leave some extra "\r" characters around in some cases). I'd appreciate some feedback on whether it should then try and trim these extra "\r" characters (if any), or whether it should be joining everything back up with "\n" instead of PHP_EOL or not, but I've just fixed the issue in this PR without consideration to any of that yet.


chrisemerson:
Should have said in my commit message - it should fix #827, though I wasn't able to replicate the precise example.

GrahamCampbell:
:+1:

ciaranmcnulty:
Great stuff, thanks

PR #832, opened 2015-12-27
Make underscores have no special meaning when using PSR4 prefix

muteor:
When using PSR4 I noticed that when trying to spec a class like `MyNs\Foo\Bar_V1` it would split the `_` into a dir separator, this does not match the expected behaviour for PSR4 because the significance of the underscore was removed in that PSR.

This PR makes the `PSR0Resource` ignore (not split) underscores in class names when the `psr4Prefix` is used. This could have some BC implications for anyone relying on the splitting when using `psr4Prefix`. I will let you make the call on that one, to me seemed reasonable to follow the PSR spec.

I am new to PHPSpec so I may have broken rules when updating the spec, happy to fix anything bad in there :)


ciaranmcnulty:
Thanks!

PR #833, opened 2016-01-01
Config/behat yml quoting

ciaranmcnulty:
This was causing a CI failure in the dev build - possibly Symfony YML quoting is stricter?


jakzal:
Indeed, see https://github.com/symfony/symfony/issues/16234 and https://github.com/symfony/symfony/pull/16285 for context.

ciaranmcnulty:
Thanks @jakzal; ping @everzet you're going to get a lot of reports about this ;-)

PR #837, opened 2016-01-04
Fix BootstrapListener to be most priority listener

nightlinus:
There is an issue when bootstrap config option is set and some class described
then class is created and phpspec can't autoload it during rerun because of BootstrapListener priority is less then ReRunListener priority.

I know i need to write test for this case, can somebody give me some tips how i can reproduce this case with behat?

### DIFF
--- a/src/PhpSpec/Listener/BootstrapListener.php
+++ b/src/PhpSpec/Listener/BootstrapListener.php
@@ -19,7 +19,7 @@ class BootstrapListener implements EventSubscriberInterface
 
     public static function getSubscribedEvents()
     {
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Thanks for finding and fixing this!

You will need to cover this in the behat suite; you would add a new scenario and tag it as `@isolated`, then add any missing step definitions to the `IsolatedProcessContext`.

If you're not sure how to do this then perhaps someone else could help.

ciaranmcnulty:
Actually this is probably ok without end to end tests; thanks for the contribution!

PR #838, opened 2016-01-05
specify the filename in psr locator exception message

docteurklein:
visual comparison:

before:
shot

after:
shot

This example uses an extremely long file name, which doesn't help to defend the PR, but anyway :)

### DIFF
--- a/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
+++ b/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
@@ -326,7 +326,7 @@ class PSR0LocatorSpec extends ObjectBehavior
         $fs->getFileContents($filePath)->willReturn('no class definition');
         $file->getRealPath()->willReturn($filePath);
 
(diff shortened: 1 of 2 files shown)


docteurklein:
PS: I just noticed a typo (does not contain**s**), but this was here already, so => other PR?

avant1:
You can amend you last commit with this typo and then push changes to your branch with --force flag. Gihub will detect that PR is changed automatically.

docteurklein:
@avant1 sure, I could :) The question was more about wether I should include it in the same PR, not how to do it.

jon-acker:
its a good idea to show the spec file name, but IMO you don't need to show the absolute path, why not shorten it to path relative to the spec/ root?

docteurklein:
@jon-acker agreed it would be better indeed. I just didn't find an easy way to do it here.
If you know a way  I could easily do that, I'll take it :)

ciaranmcnulty:
I'd like to see the path relative to the root or relative to the CWD but this is good enough right now.

Thanks @docteurklein

Issue #802, opened 2015-11-03
Support grouped use statements


ciaranmcnulty:
For true PHP7 support we need to support grouped use statements: http://php.net/manual/en/migration70.new-features.php#migration70.new-features.group-use-declarations

Because these aren't present in any existing test suites it's lower priority than the rest of the type hint fixes

Thanks @stof for pointing out.

PR #839, opened 2016-01-06
Add php7 group 'use' statements

jon-acker:
This deals with php7 grouped use statements issue reported here #802


ciaranmcnulty:
This looks good to me, thanks @jon-acker 

Closes #802

PR #850, opened 2016-02-05
Grant us the ability to run suites by SuiteName

Saeven:
For example, if I have this config:

```
suites:
    Application:
        namespace: Application
        spec_prefix: Spec
        src_path: module/Application/src/
        spec_path: module/Application/bundle
    CirclicalUser:
        namespace: CirclicalUser
        spec_prefix: Spec
        src_path: module/CirclicalUser/src/
        spec_path: module/CirclicalUser/bundle
```

I would like to be able to run this command:

> ./phpspec run CirclicalUser

In order to execute all examples in CirclicalUser's specs.

This PR achieves this.

Thanks for considering it!

### DIFF
--- a/src/PhpSpec/Locator/Suite/SuiteLocator.php
+++ b/src/PhpSpec/Locator/Suite/SuiteLocator.php
@@ -14,48 +14,47 @@
 namespace PhpSpec\Locator\Suite;
 
 use PhpSpec\Locator\PSR0\PSR0Locator;
-use PhpSpec\Locator\ResourceInterface;
 use PhpSpec\Locator\ResourceLocatorInterface;
 use PhpSpec\Util\Filesystem;
-use InvalidArgumentException;
 
 class SuiteLocator extends PSR0Locator implements ResourceLocatorInterface
 {
-    private $suiteName;
+	private $suiteName;
 
-    /**
-     * @param string     $srcNamespace
-     * @param string     $specNamespacePrefix
-     * @param string     $srcPath
-     * @param string     $specPath
-     * @param Filesystem $filesystem
-     * @param string     $psr4Prefix
-     */
-    public function __construct(
-        $name = '',
-        $srcNamespace = '',
-        $specNamespacePrefix = 'spec',
-        $srcPath = 'src',
-        $specPath = '.',
-        Filesystem $filesystem = null,
-        $psr4Prefix = null
-    ) {
-        $this->suiteName = $name;
-        parent::__construct( $srcNamespace, $specNamespacePrefix, $srcPath, $specPath, $filesystem, $psr4Prefix );
-    }
+	/**
+	 * @param string $srcNamespace
+	 * @param string $specNamespacePrefix
+	 * @param string $srcPath
+	 * @param string $specPath
+	 * @param Filesystem $filesystem
+	 * @param string $psr4Prefix
+	 */
+	public function __construct(
+		$name = '',
+		$srcNamespace = '',
+		$specNamespacePrefix = 'spec',
+		$srcPath = 'src',
+		$specPath = '.',
+		Filesystem $filesystem = null,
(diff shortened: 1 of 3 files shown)


Saeven:
Hm. It marked some stuff as bugs, but they aren't.   Hopefully this is ok.

jakubzapletal:
IMHO this arguments is confusing, it's the same like specifying a path to the file. I would suggest to use a parameter like Behat does `-s` and `--suite=`.

Saeven:
The parameter is distinctly different from a path, you don't feel it's safe to infer that if it matches a suite name, it's a suite?  (e.g., no slashes and so forth)

jakubzapletal:
yeah, I don't think it's obvious enough. Maybe, I'm affected by Behat.
Anyway, you're missing any tests.

ciaranmcnulty:
Thanks for this, it's a good feature! 

I'm wary of making the path too 'fuzzy' though. @jakubzapletal's might be the correct way to do but we'd need to consider what happens when both the `--suite` and a path are given.

You'll need to add some tests to cover the new funtcionality before it can be merged; I think the Travis failure is happening for all PRs so I'll do some investigation.

Saeven:
I'm relieved that you're open to it!  I'm in the very nascent stages of a multi-module app, and running entire test suites, by module, is necessary.  Running the whole lot of tests for all modules when you are working on one of ten or so, can be quite a time sink.

I added some cursory tests.  It piggybacks on existing classes, I'm not sure how exhaustive it truly needs to be; it's really a 'does the suite exist or not?' type of question that it must answer.  The rest is deferred to the PSR0 locator.

Saeven:
I will fedex :beers: if you merge this! :)

src/PhpSpec/Locator/Suite/SuiteLocator.php
```diff
@@ -0,0 +1,59 @@
...
+
+namespace PhpSpec\Locator\Suite;
+
+use PhpSpec\Locator\PSR0\PSR0Locator;
+use PhpSpec\Locator\ResourceLocatorInterface;
+use PhpSpec\Util\Filesystem;
+
+class SuiteLocator extends PSR0Locator implements ResourceLocatorInterface
+{
+	private $suiteName;
```
sroze: Can you use 4 spaces instead of tabs for indentation?

Saeven: Sure!

stof:
> I'm wary of making the path too 'fuzzy' though. @jakubzapletal's might be the correct way to do but we'd need to consider what happens when both the --suite and a path are given.

what Behat does is to run the feature files matching the path, among features files of the specified suite. This is the better way IMO: just do a AND of both conditions given in the command

Saeven:
I swapped the tabs for spaces, looks like Scrutinizer needs a restart though.

src/PhpSpec/Locator/Suite/SuiteLocator.php
```diff
@@ -0,0 +1,60 @@
...
+     * @param string $query
+     *
+     * @return bool
+     */
+    public function supportsQuery($query)
+    {
+        return $query == $this->suiteName;
+    }
+
+    public function findResources($query)
```
mheki: missing doc block here

Saeven:
I'll double the amount of :beer: ! :)

ciaranmcnulty:
@Saeven Can you change to a `--suite` option please?

ciaranmcnulty:
@Saeven If you have time to come back to this with the `--suite` option please reopen, closing for now.

Saeven:
Understood, will do -- just been busy with work these days...

Podbi:
Hi, any updates about this pull request? Running a specific suite by ``--suite`` option could be handy. :+1:

PR #851, opened 2016-02-06
Adds support for PHP7 Error classes

afurculita:
The PHP 7 new Error classes are not supported for now by PhpSpec. This PR adds this support for the ThrowMatcher

### DIFF
--- a/src/PhpSpec/Matcher/ThrowMatcher.php
+++ b/src/PhpSpec/Matcher/ThrowMatcher.php
@@ -28,10 +28,12 @@ class ThrowMatcher implements MatcherInterface
      * @var array
      */
     private static $ignoredProperties = array('file', 'line', 'string', 'trace', 'previous');
+
     /**
      * @var Unwrapper
      */
     private $unwrapper;
+
     /**
      * @var PresenterInterface
      */
@@ -43,9 +45,9 @@ class ThrowMatcher implements MatcherInterface
     private $factory;
 
     /**
-     * @param Unwrapper          $unwrapper
-     * @param PresenterInterface $presenter
-     * @param ReflectionFactory  $factory
+     * @param Unwrapper              $unwrapper
+     * @param PresenterInterface     $presenter
+     * @param ReflectionFactory|null $factory
      */
     public function __construct(Unwrapper $unwrapper, PresenterInterface $presenter, ReflectionFactory $factory = null)
     {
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
It will be cool to have this feature!

I'm uncomfortable with requiring the Symfony polyfill because although I like to use it for projects it is not perfect and could cause confusion.

Is it possible to implement a solution without it?

GrahamCampbell:
> I'm uncomfortable with requiring the Symfony polyfill because although I like to use it for projects it is not perfect and could cause confusion.

Adding that could mess with the code we're testing.

afurculita:
@ciaranmcnulty yep, you're right here. I just wanted to avoid checks for php versions in code. But don't really see how this could mess with the code we're testing.

I will replace the polyfill with checks for version 7 of php.

ciaranmcnulty:
Great; I suspect the version checks will only be needed in the test suite

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -36,22 +36,21 @@ class ThrowMatcher implements MatcherInterface
      * @var PresenterInterface
      */
     private $presenter;
-
```
GrahamCampbell: ?

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -91,114 +90,146 @@ public function negativeMatch($name, $subject, array $arguments)
...
     public function verifyPositive($callable, array $arguments, $exception = null)
     {
-        try {
-            call_user_func_array($callable, $arguments);
-        } catch (\Exception $e) {
-            if (null === $exception) {
-                return;
+        $exceptionThrown = null;
+
+        if (PHP_VERSION_ID < 70000) {
```
GrahamCampbell: No need for this, just try catch both exception and throwable.

GrahamCampbell: eg https://github.com/laravel/framework/blob/v5.2.14/src/Illuminate/View/View.php#L80

ciaranmcnulty: Yes, you can `catch` something that doesn't exist in the current PHP version without an error.

src/PhpSpec/Matcher/ThrowMatcher.php
afurculita: The other properties don't have an empty line between them, so just for consistency I removed this one

GrahamCampbell: do they have phpdoc though?

src/PhpSpec/Matcher/ThrowMatcher.php
afurculita: Yep, you're right

src/PhpSpec/Matcher/ThrowMatcher.php
afurculita: Yep

afurculita:
@ciaranmcnulty any news on this PR?

ciaranmcnulty:
Tests seem to be failing?

afurculita:
@ciaranmcnulty the tests are failing due to problems not related to this PR, it seems

ciaranmcnulty:
There are some coding style issues; you're right about Travis so I'll have to fix that before I can merge.

src/PhpSpec/Matcher/ThrowMatcher.php
ciaranmcnulty: if you want to alter for consistency I'd rather have a line between them

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -91,114 +90,134 @@ public function negativeMatch($name, $subject, array $arguments)
...
-            if (null === $exception) {
-                return;
-            }
+            $exceptionThrown = $e;
+        } catch (\Throwable $e) {
+            $exceptionThrown = $e;
+        }
+
+        if (!$exceptionThrown) {
+            throw new FailureException('Expected to get exception, none got.');
```
ciaranmcnulty: Maybe this message should reflect that it might not be an exception? `Expected to get exception / throwable...`?

afurculita:
The coding styles issues are caused by the `Throwable` interface, that doesn't exist on versions lower than 7

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -261,15 +280,21 @@ private function getException(array $arguments)
             return $arguments[0];
         }
 
-        if (is_object($arguments[0]) && $arguments[0] instanceof \Exception) {
-            return $arguments[0];
+        if (is_object($arguments[0])) {
+            if (PHP_VERSION_ID >= 70000 && $arguments[0] instanceof \Throwable) {
```
ciaranmcnulty: I'm not sure how this will work with the various HHVM version numberings. Might it be better to `class_exists` on the `Throwable` type?

ciaranmcnulty:
OK if you can address my minor comments I'm happy to merge once I figure out WTF travis is doing

afurculita:
@ciaranmcnulty I've made all the changes

ciaranmcnulty:
Great stuff, thanks!

PR #853, opened 2016-02-08
Fixes #852 Handle opening brace of string interpolation

### LINKED ISSUE #852: Presence of curly braces in the specified object causes syntax error during nonexistent method creation.
**phpspec version 2.4.1**
**PHP 5.5.30**

Specified object:
```
<?php

// Specified object
class Markdown
{
    public function toHtmlFromReader(Reader $reader)
    {
        return "<p>{$reader->getMarkdown()}</p>"; // curly braces
    }
}
```
Phpspec for this object:
```
<?php

namespace spec;
use PhpSpec\ObjectBehavior;

class MarkdownSpec extends ObjectBehavior
{
    function it_should_have_second_method()
    {
        $this->second()->shouldReturn(true);
    }
}
```

When you run: `bin/phpspec run`
` Do you want me to create `Markdown::second()` for you? ` appears (because specified object doesn't have the method `second()`)

Press `Y`:
And receive:
```
 Method Markdown::second() has been created.
  
PHP Parse error:  syntax error, unexpected 'public' (T_PUBLIC) in /src/Markdown.php on line 10

Parse error: syntax error, unexpected 'public' (T_PUBLIC) in /src/Markdown.php on line 10

``` 

And in the source code of specified object we can see:
```
<?php

// Specified object
class Markdown
{
    public function toHtmlFromReader(Reader $reader)
    {
        return "<p>{$reader->getMarkdown()}</p>"; // curly braces

    public function second()
    {
        // TODO: write logic here
    }
    }
}
```

shanethehat:
When tokenizing a Class, a `{` character is normally treated as a vanilla string. When it appears within a double quoted string however, it takes on a special meaning and is represented differently in the tokenizer. This was not being handled in the code that detects method and class endings by matching brace counts.

### DIFF
--- a/spec/PhpSpec/CodeGenerator/Writer/TokenizedCodeWriterSpec.php
+++ b/spec/PhpSpec/CodeGenerator/Writer/TokenizedCodeWriterSpec.php
@@ -62,6 +62,15 @@ class TokenizedCodeWriterSpec extends ObjectBehavior
         $this->shouldThrow($exception)->during('insertAfterMethod', array($class, 'methodOne', ''));
     }
 
+    function it_should_generate_a_method_in_a_class_with_a_string_containing_braces()
+    {
+        $class = $this->getClassWithBraceText();
+        $method = $this->getMethod();
+        $result = $this->getClassWithBraceTextAndNewMethod();
+
+        $this->insertMethodLastInClass($class, $method)->shouldReturn($result);
+    }
+
     private function getSingleMethodClass()
     {
         return <<<SINGLE_METHOD_CLASS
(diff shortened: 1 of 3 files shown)


jakzal:
I'm not convinced if we need a new scenario for a bug fix. Examples can easily become unreadable if we do this every time a bug is found.

shanethehat:
Fair point, the spec is probably enough to cover the edgecase found here. I'll update later today.

shanethehat:
@jakzal fixed :)

features/code_generation/developer_generates_method.feature
```diff
@@ -371,3 +371,4 @@ Feature: Developer generates a method
       }
 
       """
+
```
ciaranmcnulty: can you remove this line please?

shanethehat: fixed

ciaranmcnulty:
Thanks @shanethehat!

PR #855, opened 2016-02-11
Fix tests on windows

avant1:
Finishing work started at #829.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,10 +7,6 @@ environment:
     matrix:
         - PHP_DOWNLOAD_FILE: php-5.6.14-nts-Win32-VC11-x86.zip
 
-matrix:
-    allow_failures:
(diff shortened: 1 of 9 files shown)


avant1:
I'm not sure if it is OK to ignore error of temp directories removing.
At the moment of directory removing it is already empty, but Windows keeps saying "Directory not empty".

I cannot understand Scrutinizer errors, it refers to pull request that is not merged yet (#850), and as I can see that PR does not even change `ObjectBehavior.php` file (where coding style errors are found). 

Travis is being failed because `openssl` extension is disabled on *5.3.3* (note [here](https://docs.travis-ci.com/user/languages/php#PHP-installation)). 
The easiest way to fix this error is use *5.3* instead of *5.3.3*, but maybe this extension is installed, but not included in .ini file, so we could enable it in `before_install` script, I'm not sure.

ciaranmcnulty:
Thanks for this; it'll be great to be able to detect Windows breaks

Issue #845, opened 2016-01-25
Send warning before creating a method or class with a php reserved keyword.


petemcfarlane:
Instead of creating a method with a name which is a reserved keyword (which will create a fatal error anyway) that phpspec should instead warn the user.

Example spec.

![screen shot 2016-01-25 at 15 33 48](https://cloud.githubusercontent.com/assets/3472717/12555416/17df6548-c379-11e5-9189-24bef2bef862.png)

![screen shot 2016-01-25 at 15 35 04](https://cloud.githubusercontent.com/assets/3472717/12555449/3b38b652-c379-11e5-8dd6-729319d771fa.png)

PR #856, opened 2016-02-12
Don't allow to generate a method named with a PHP restricted keyword

mheki:
Fixes: https://github.com/phpspec/phpspec/issues/845

Instead of generating method with wrong name and failing it warns user about the problem: 
screen shot 2016-02-12 at 16 14 46


src/PhpSpec/Listener/MethodNotFoundListener.php
```diff
@@ -27,12 +28,28 @@ class MethodNotFoundListener implements EventSubscriberInterface
...
+     * @param IO $io
+     * @param ResourceManagerInterface $resources
+     * @param GeneratorManager $generator
+     * @param VoterInterface $nameChecker
+     */
+    public function __construct(
+        IO $io,
+        ResourceManagerInterface $resources,
+        GeneratorManager $generator,
+        VoterInterface $nameChecker
```
ciaranmcnulty: Changing the constructor in this way is a BC break.

Please make this optional and instantiate a new copy inside the constructor if it's not provided (we can remove this in PhpSpec 3.0)

src/PhpSpec/Util/VoterInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Util;
+
+interface VoterInterface
```
ciaranmcnulty: This naming is far too generic. Where you use it elsewhere it's a `nameChecker` which feels closer to a good name.

Find something that describes how this type is _used_, then for the concrete think about _how_ it does it

src/PhpSpec/Util/PhpNameChecker.php
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Util;
+
+final class PhpNameChecker implements VoterInterface
+{
+    private $reservedWords = array(
```
ciaranmcnulty: Where is this list from?

src/PhpSpec/Listener/MethodNotFoundListener.php
```diff
@@ -82,4 +106,19 @@ public function afterSuite(SuiteEvent $event)
...
+    {
+        if (!$this->nameChecker->supports($methodName)) {
+            $this->wrongMethodNames[] = $methodName;
+        }
+    }
+
+    private function writeErrorMessage()
+    {
+        foreach ($this->wrongMethodNames as $methodName) {
+            $message = sprintf("You cannot use restricted `%s` as a method name", $methodName);
```
ciaranmcnulty: Maybe instead of `restricted` use `the reserved word`

ciaranmcnulty:
I agree with @avant1 about case issues,

I made some naming comments but the overall code is solid

src/PhpSpec/Util/PhpNameChecker.php
mheki: http://php.net/manual/en/reserved.keywords.php

mheki:
done

src/PhpSpec/Listener/MethodNotFoundListener.php
```diff
@@ -27,12 +29,33 @@ class MethodNotFoundListener implements EventSubscriberInterface
...
+        IO $io,
+        ResourceManagerInterface $resources,
+        GeneratorManager $generator,
+        NameCheckerInterface $nameChecker = null
+    ) {
         $this->io        = $io;
         $this->resources = $resources;
         $this->generator = $generator;
+
+        if (null === $nameChecker) {
```
ciaranmcnulty: In other places we've done this with a ternary:

```
$this->nameChecker = $nameChecker ? $nameChecher : new ReservedWordsMethodNameChecker();
```

(one day we can use the `?:` operator when we drop old PHP versions

src/PhpSpec/Listener/MethodNotFoundListener.php
```diff
@@ -63,6 +88,11 @@ public function afterSuite(SuiteEvent $event)
             return;
         }
 
+        if (!empty($this->wrongMethodNames)) {
```
ciaranmcnulty: `empty` is redundant here, empty arrays are always `false` and if it stops being an array I'd prefer a hard break

mheki:
I've just found (and pushed changes) that `CollaboratorMethodNotFoundListener` needs the same functionality, now I'm almost certain this bugfix should be done with additional event listener instead of repeating functionality all over again

src/PhpSpec/Console/IO.php
```diff
@@ -342,4 +342,25 @@ public function getBlockWidth()
...
         return $width;
     }
+
+    /**
+     * @param string $message
+     * @param int|null $indent
+     */
+    public function writeError($message, $indent = null)
+    {
+        $message = wordwrap($message, $this->getBlockWidth() - 4, "\n", true);
```
stof: why ``- 4`` here ?

src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php
```diff
@@ -48,15 +50,31 @@ class CollaboratorMethodNotFoundListener implements EventSubscriberInterface
...
+    public function __construct(
+        IO $io,
+        ResourceManagerInterface $resources,
+        GeneratorManager $generator,
+        NameCheckerInterface $nameChecker = null
+    ) {
         $this->io = $io;
         $this->resources = $resources;
         $this->generator = $generator;
+        $this->nameChecker = $nameChecker ? $nameChecker : new ReservedWordsMethodNameChecker();
```
stof: this can use a shorthand notation

ciaranmcnulty: You're right, for some reason I thought `?:` was 5.4

src/PhpSpec/Console/IO.php
mheki: to be consistent with `askConfirmation()` https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Console/IO.php#L274

mheki: if we want to keep it more generic, it should be `$indent * 2` tbh, then the error message is displayed with the same padding on both sides

ciaranmcnulty: That sounds good.

mheki:
I've changed wording and the block itself, so now corresponds to the `broken code` indications. When the method is wrong, runner will not run the suite again.
screen shot 2016-02-16 at 16 08 39

ciaranmcnulty:
Great, thanks @mheki!

PR #860, opened 2016-02-16
Don't cache vendor folder

ciaranmcnulty:
Seeing if this fixes the appveyor build...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,6 @@ skip_commits:
 cache:
     - c:\php -> appveyor.yml
     - '%LOCALAPPDATA%\Composer'
-    - vendor
 
 init:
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Well this fixes things, a more nuanced fix can come later.

stof:
Caching the composer cache is better than caching the vendor folder: an existing vendor folder impacts the dependency resolution, which can result into issues if it is corrupted.

ciaranmcnulty:
I'm not sure what was breaking, I suspect the `.git` files were not being included in the cache

ciaranmcnulty:
@sroze has pointed out https://github.com/symfony/symfony/pull/17171/files which might be a good option for future

PR #862, opened 2016-02-17
Add the `shouldImplement` method in ObjectBehaviour's PhpDoc.

sroze:
Because everybody love autocompletion and checking that an object implement a given interface.


mheki:
:smile:  :+1:

ciaranmcnulty:
Cool, thanks

PR #863, opened 2016-02-19
write stats once finished

bendavies:
Stats were not printed by the dot formatter on the last line.

before:
screen shot 2016-02-19 at 14 27 24
after:
screen shot 2016-02-19 at 14 27 12

### DIFF
--- a/src/PhpSpec/Formatter/DotFormatter.php
+++ b/src/PhpSpec/Formatter/DotFormatter.php
@@ -61,13 +61,21 @@ class DotFormatter extends ConsoleFormatter
                 break;
         }
 
-        if ($eventsCount % 50 === 0) {
+        $remainder = $eventsCount % 50;
+        $lastRow = $eventsCount === $this->examplesCount;
+
+        if ($remainder === 0 || $lastRow) {
             $length = strlen((string) $this->examplesCount);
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
This is a good idea; is it necessary to have the `$column` though? It looks like it is equal to `$eventsCount % 50`

src/PhpSpec/Formatter/DotFormatter.php
```diff
@@ -61,12 +67,18 @@ public function afterExample(ExampleEvent $event)
                 break;
         }
 
-        if ($eventsCount % 50 === 0) {
+        if ($eventsCount % 50 === 0 || $eventsCount === $this->examplesCount) {
             $length = strlen((string) $this->examplesCount);
             $format = sprintf(' %%%dd / %%%dd', $length, $length);
+
+            if ($eventsCount === $this->examplesCount) {
```
avant1: This condition is being used twice, probably you could create a method which checks this or save result into a variable with meaning name.

bendavies: done, thanks.

bendavies:
@ciaranmcnulty good spot, done.

Issue #868, opened 2016-02-20
Remove dependency on Symfony EventDispatcher


ciaranmcnulty:
Currently we depend on Symphony's EventDispatcher component. Many listeners are implemented as EventSubscribers, including currently Formatters.

I'd prefer to have a PhpSpec abstraction here, even if it's then implemented using the Symfony EventDispatcher.

My own preference would be to move to direct listening with new interfaces like `ExampleListener` that has `beforeExample` and `afterExample`, etc.

Dragonrun1:
I've replaced Symphony's EventDispatcher myself in one of my project and in the process created a low dependence library/new project you might take a look at.
https://github.com/Dragonrun1/event-mediator
It combines many of the better features (IMHO) that I found floating around in other event systems.

PR #871, opened 2016-02-20
Hk issue 868

hanneskaeufler:
I had a few minutes so I thought I made a move on issue #868.
This doesn't remove the dependency yet but kinda isolates it into a smaller corner.
What I did there is

- [x] introduce `ExampleListener`, `SuiteListener`, `SpecificationListener`.
- [x]  have the listenerers in `src/PhpSpec/Listener` conform to one or multiple of these interfaces.
- [x]  have the formatters in `src/Formatter` conform to all of these interfaces (see `BasicFormatter`).
- [x] updated specs to assert on more intention revealing interfaces I just introduced.

Let me know if this is at all what you were thinking of :) It might be a start to the issue at hand at least ...

### DIFF
--- a/spec/PhpSpec/Formatter/HtmlFormatterSpec.php
+++ b/spec/PhpSpec/Formatter/HtmlFormatterSpec.php
@@ -23,7 +23,7 @@ class HtmlFormatterSpec extends ObjectBehavior
 
     function it_is_an_event_subscriber()
     {
-        $this->shouldHaveType('Symfony\Component\EventDispatcher\EventSubscriberInterface');
+        $this->shouldHaveType('PhpSpec\Listener\ExampleListener');
     }
 
     function it_delegates_the_reporting_to_the_event_type_line_reporter(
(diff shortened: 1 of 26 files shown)


mheki:
Thanks for your work @hanneskaeufler !
From the code use cases it looks like we could have slightly different interface segragation:
`afterExample` method usually comes with the `afterSuite`, so having some kind of `AfterExampleAndSuite` interface would not introduce all these empty methods (`beforeExample`, `beforeSuite`). What do you think @ciaranmcnulty ?

hanneskaeufler:
You make a good point, I'm not very fond of those empty methods either. But I guess we will still have one or two of them unless we have an interface for each method.

ciaranmcnulty:
Hm good points about the empty methods - I'm not sure what the best way to proceed is now.

hanneskaeufler:
Lemme know if it's worth rebasing this?

ciaranmcnulty:
Yes sorry for inactivity here. This part of the codebase has fallen off my radar a bit for 3.0..

Rebase if you have time, but I doubt it'll be in the next release. Definitely need to clean up how listeners work

hanneskaeufler:
No worries, rebase done.

ciaranmcnulty:
I'm no longer sure this is the direction to go in, but thank you for the hard work implementing it

hanneskaeufler:
My pleasure, I learned some things 👌🏿

Issue #869, opened 2016-02-20
Remove optional deps


ciaranmcnulty:
As we've worked through 2.x branch we have introduced new dependencies for objects but because of backwards compatibility we've made them optional.

For the `3.0` release these should be made non-optional

PR #872, opened 2016-02-22
Remove optional dependencies

mheki:
This PR removes optional dependency of NameChecker for the 3.0 version. Issue: https://github.com/phpspec/phpspec/issues/869


ciaranmcnulty:
Thanks! Will wait until we have done the 5.5 release before merging

mheki:
I removed optional dependencies from all the classes I found with looking for `?:` in the project.
Also moved Filesystem argument in PSR0Locator before those with default values.

ciaranmcnulty:
There may be some in there that don't use `?:` but use a traditional ternary instead. I guess we need to audit all optional parameters (yawn!)

mheki:
Yeah, found one. Can't see anything else.

ciaranmcnulty:
@mheki can you resolve the merge conflicts please?

src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -63,14 +63,14 @@ class PSR0Locator implements ResourceLocatorInterface
      * @param string     $psr4Prefix
      */
     public function __construct(
+        Filesystem $filesystem,
```
stof: Moving a parameter around breaks BC. this may be fine, as this class is mostly internal, but it should be acknowledged.
@ciaranmcnulty it may be great to add ``@internal`` annotations where appropriate

stof: @mheki in any case, you forgot to reorder the phpdoc

ciaranmcnulty:
@mheki I think merging your other PR has caused conflicts?

mheki:
@ciaranmcnulty branch brought up to date

Issue #873, opened 2016-02-23
Remove deprecated code


mheki:
Approaching version 3.0 gives us an opportunity to remove deprecated classes:
-  `\PhpSpec\Formatter\Html\HtmlPresenter`
-  `\PhpSpec\Formatter\Presenter\PresenterInterface`
-  `\PhpSpec\Formatter\Presenter\StringPresenter`
-  `\PhpSpec\Formatter\Presenter\TaggedPresenter`

And the method:
-  `\PhpSpec\Util\Filesystem::findPhpFilesIn()`

ciaranmcnulty:
Thanks @mheki !

Sam-Burns:
Is this issue fixed by the pull request?

PR #874, opened 2016-02-23
Remove deprecated code

mheki:
This PR removes deprecated code for 3.0 version. Closes issue https://github.com/phpspec/phpspec/issues/873

- [x] `\PhpSpec\Formatter\Html\HtmlPresenter`
- [x] `\PhpSpec\Formatter\Presenter\PresenterInterface`
- [x] `\PhpSpec\Formatter\Presenter\StringPresenter`
- [x] `\PhpSpec\Formatter\Presenter\TaggedPresenter`
- [x] `\PhpSpec\Util\Filesystem::findPhpFilesIn()`


ciaranmcnulty:
@mheki can you resolve the merge conflicts please?

src/PhpSpec/Formatter/JUnitFormatter.php
```diff
@@ -52,7 +52,7 @@ class JUnitFormatter extends BasicFormatter
         ExampleEvent::SKIPPED => 'skipped',
     );
 
-    public function __construct(PresenterInterface $presenter, IO $io, StatisticsCollector $stats)
+    public function __construct(Presenter $presenter, IO $io, StatisticsCollector $stats)
```
stof: technically, all those typehints must be changed in 2.x already. Otherwise, it is impossible to avoid using the BC layers (as everything forces you to rely on the deprecated interface)

stof:
@ciaranmcnulty IMO, the first step before removing deprecated code would be to make usage of deprecated code clear when running specs, to prevent mistakes were we force using a deprecated layer:

- trigger deprecation notices when using deprecated code. I suggest using the Symfony convention, i.e. silencing them, as this prevents them to become exceptions with common error handlers. Not silencing them would mean that common error handlers would turn them  into exceptions, thus breaking BC layers (as the exception would prevent them from running). We made this mistake in Symfony 2.7.0 and then thought about silencing them
- collect triggered deprecations. You can look at how Symfony does it in the Debug component (to log them) and the PHPUnit bridge (to display them and change the exit code when running PHPUnit). As you use both Behat and PhpSpec, this means that both tools should support such feature (I would totally agree about having this feature in the core for Behat, but I haven't had time to implement this yet)

The next step would be to ensure that PhpSpec can run without triggering any deprecation warning (meaning that users can get rid of all deprecations warning too), except in places testing legacy behaviors explicitly (in PHPUnit, we identify them with ``@group legacy``, using the PHPUnit group feature).
This step will show you that ~80% of the changes done in this PR are actually needed in 2.x (this PR should almost only contain removal of code)

src/PhpSpec/Formatter/JUnitFormatter.php
stof: @ciaranmcnulty there is already the new interface and the old one in parallel in the codebase, with the old one being deprecated and inheriting from the new one. But as all typehints are using the child interface, the new interface is actually unusable today. So typehints must be changed in 2.x to allow using the existing ``Presenter`` interface (using the deprecated one will work, as it extends the new one)

stof: the issue is that the PR deprecating PresenterInterface in 2.x did only half of the work (which was not caught because PhpSpec does not have the testing infrastructure in place to detect the usage of deprecated APIs)

mheki:
I've been on holiday. Will take a look into it in a few days.

src/PhpSpec/Formatter/JUnitFormatter.php
ciaranmcnulty: We can detect usage of deprecated APIs by removing them and running the test suite ;-)

ciaranmcnulty:
@stof I put my thoughts in #896  - i'm happy with these changes so long as they're documented.

ciaranmcnulty:
Thanks @mheki !

PR #878, opened 2016-03-01
Only rewrite Spec classes

sroze:
Some other classes can be found in Spec files for testing purposes. It should not remove the typehints of them because if they implement an interface, the contract will be broken.

### DIFF
--- a/src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
+++ b/src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
@@ -74,6 +74,7 @@ final class TokenizedTypeHintRewriter implements TypeHintRewriter
         $this->currentClass = '';
         $this->currentFunction = '';
     }
+
     /**
      * @param array $tokens
      * @return array $tokens
@@ -81,6 +82,13 @@ final class TokenizedTypeHintRewriter implements TypeHintRewriter
     private function stripTypeHints($tokens)
     {
         foreach ($tokens as $index => $token) {
+            if ($this->isToken($token, '{')) {
+                $this->currentBodyLevel++;
+            }
+            elseif ($this->isToken($token, '}')) {
+                $this->currentBodyLevel--;
+            }
+
             switch ($this->state) {
                 case self::STATE_READING_ARGUMENTS:
                     if (')' == $token) {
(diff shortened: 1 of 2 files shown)


src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -62,7 +62,7 @@ public function rewrite($classDefinition)
         $this->namespaceResolver->analyse($classDefinition);
 
         $this->reset();
-        $tokens = $this->stripTypeHints(token_get_all($classDefinition));
+        $tokens = $this->stripTypeHints(token_get_all($classDefinition), strpos($classDefinition, 'ExceptionFactorySpec') !== false);
```
stof: what is this condition exactly ?

sroze: as you can imagine, that wasn't suppose to be here, I just removed it.

sroze:
@ciaranmcnulty FYI, the Scrutinizer failures are not related to this PR, except one duplication that is not really accurate. Obviously that would be better with [the state machine](https://github.com/phpspec/phpspec/compare/master...ciaranmcnulty:feature/php7-typehint-support-object-states) but at the moment, that'll fix this problem with PHP7.

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -192,4 +197,25 @@ private function tokenHasType($token, $type)
...
     }
+
+    /**
+     * @param string $className
+     *
+     * @return bool
+     */
+    private function shouldExtractTokensOfClass($className)
+    {
+        return substr($className, -4) == 'Spec';
```
ciaranmcnulty: This duplicates logic inside the Locator (i.e. it is an assumption that an extension could change). I'm not really happy having it here.

sroze: The best option would be to have the `ResourceInterface` object inside this rewriter, but that's not possible because it come from a stream wrapper that just have the classname. 

The second option can be to use the `ResourceManager` from the classname. If a resource is found, then we extract the tokens from that class. What do you think about that?

ciaranmcnulty:
@sroze After discussing this offline I think the (theoretical) objection about ResourceManager isn't a problem

PR #881, opened 2016-03-06
Drop old PHP/Symfony versions

ciaranmcnulty:
As per the release roadmap, PhpSpec 3.0 will only support PHP 5.6+ and Symfony 2.7+


composer.json
```diff
@@ -18,15 +18,15 @@
     ],
 
     "require": {
-        "php":                      ">=5.3.3",
-        "phpspec/prophecy":         "~1.4",
+        "php":                      "^5.6.0|^7.0.0",
```
GrahamCampbell: why not `~5.6|~7.0` to match every other version constraint in this file?

Issue #876, opened 2016-02-25
Run with spec class fullname


lfbn:
Hi,

I think could be useful be runned a specific spec without defining .php.

When running, if is a folder, runs for every specs as is today, if a file, it runs for the specific spec.

ciaranmcnulty:
I don't like the idea of overloading the resource locator further (this is related to the discussion in #850)

PR #882, opened 2016-03-08
Support to run a spec file omitting file extension

lfbn:
Following [issue 876](https://github.com/phpspec/phpspec/issues/876) I think could be better to run a specific spec without defining the .php file extension. The existent behaviour should stay as it is.

For example, before, we needed to define the file extension:
`bin/phpspec run spec/PhpSpec/CodeAnalysis/MagicAwareAccessInspectorSpec.php`

Now, we don't need to define it:
`bin/phpspec run spec/PhpSpec/CodeAnalysis/MagicAwareAccessInspectorSpec`

Notice that, if file extension is defined, it should work as before.

### DIFF
--- a/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
+++ b/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
@@ -204,6 +204,15 @@ class PSR0LocatorSpec extends ObjectBehavior
         )->shouldReturn(true);
     }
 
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
But the first thing is a real file path, the second is not a FQCN

lfbn:
Ok @ciaranmcnulty . But what you think the issue 876 was?

ciaranmcnulty:
I appreciate the effort you have put into it, but I don't believe the extra complexity is worth saving a few characters (which most shells will autocomplete)

PR #891, opened 2016-03-30
Be consistent with ^ notation

GrahamCampbell:
I've replaced `~` with `^` in all places where it does not change the behaviour.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,14 +19,14 @@
 
     "require": {
         "php":                      "^5.6|^7.0",
-        "phpspec/prophecy":         "~1.5",
+        "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "~1.0.0",
-        "sebastian/exporter":       "~1.0",
-        "symfony/console":          "~2.7|~3.0",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
What's the benefit?

unexge:
@ciaranmcnulty AFAIK, `~1.5` -> `>=1.5 <1.6`, `^1.5` -> `>=1.5.0 <2.0.0`, it also [recommended](https://getcomposer.org/doc/articles/versions.md#caret) by composer.

ciaranmcnulty:
@unexge I don't think you are correct

GrahamCampbell:
> AFAIK, ~1.5 -> >=1.5 <1.6

That's not correct I'm afraid.

GrahamCampbell:
> That's not correct I'm afraid.

That describes what `~1.5.0` would do.

ciaranmcnulty:
I don't see a reason to change these

GrahamCampbell:
The fact you randomly have a `^2.7|~3.0` along with those `~2.7|~3.0` is very odd.

PR #892, opened 2016-03-31
Remove use of expect

shanethehat:
Remove the circular dependency on the expect library, since it keeps causing issues.


ciaranmcnulty:
It saddens me to use the readability but this is a good step.

I'm not sure FailureException is appropriate?

shanethehat:
It's what was in use in the matchers, but quite possibly not. Do you have a suggestion?

ciaranmcnulty:
Oh yeah that was probably me :-)

I guess just `\Exception`

features/bootstrap/ApplicationContext.php
```diff
@@ -1,23 +1,21 @@
...
 use Symfony\Component\Console\Tester\ApplicationTester;
+use PhpSpec\Exception\Example\FailureException;
 
 /**
  * Defines application features from the specific context.
  */
-class ApplicationContext implements Context, MatchersProviderInterface
+class ApplicationContext implements Context
 {
+    const JUNIT_XSD_PATH = '/src/PhpSpec/Resources/schema/junit.xsd';
```
stof: as this constant is used only once, and it does not contain the full path anyway, I suggest removing entirely, keeping the path inline

ciaranmcnulty: I think it adds to clarity to have this here - it's a shame it can't yet contain the full path (unless `.` is now ok in constants)

shanethehat:
@ciaranmcnulty fixed exception class

PR #893, opened 2016-03-31
Fixes #884 Do not duplicate events data for counting

### LINKED ISSUE #884: phpspec not releasing memory/using a lot of memory
We've got ~1300 tests in our suite currently and just ran into an issue where PHP was unable to allocate memory because it had hit the 128M limit we have set. Of course we can bump that up, but it doesn't seem to make sense that it would need to keep all the objects around after the tests have run.

The error that happened that made us aware of this was:

```
Fatal error happened while executing the following 
    it will build the ObjectStatusMap service 
    Allowed memory size of 134217728 bytes exhausted (tried to allocate 72 bytes) in /vagrant_web/vendor/phpspec/phpspec/src/PhpSpec/Listener/StatisticsCollector.php on line 89 
```

Interestingly, if we run with format=pretty, it succeeds. If we comment out some tests, it also works. 

I'm looking at potentially parallelizing the run so fewer tests will run within a single run of phpspec, but it still leaves us in the same state if the individual namespaces grow to a similar number/size of tests as the full suite.

shanethehat:
screen shot 2016-04-01 at 09 09 38 
Blackfire profile of phpspecs own spec suite compared to master

### DIFF
--- a/src/PhpSpec/Listener/StatisticsCollector.php
+++ b/src/PhpSpec/Listener/StatisticsCollector.php
@@ -132,7 +132,7 @@ class StatisticsCollector implements EventSubscriberInterface
 
     public function getEventsCount()
     {
-        return count($this->getAllEvents());
+        return array_sum($this->getCountsHash());
     }
 
     public function getTotalSpecsCount()


ciaranmcnulty:
So it takes slightly more time and only saves 0.3% memory?

I agree with the change from a clarity POV but I'm not convinced it's a huge perf. gain; am I missing something?

shanethehat:
That's only on phpspec's own test suite. I would be interested to see the difference on a bigger suite with thousands of examples.

ciaranmcnulty:
@shanethehat I'd also like to see that before merging

svemir:
Our suite has 108 specs and 523 examples. There is no statistically significant difference in wall time or memory usage with this change.

ciaranmcnulty:
Doesn't seem to make a big diff then?

ciaranmcnulty:
I'll close this on the grounds the original seems to reveal intent better

PR #894, opened 2016-04-04
Use single quotes; add document markers; remove xdebug in non-hhvm en…

shrikeh:
…vironments

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+---
 language: php
 
 sudo: false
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Are the document markers important?

What difference do the single-quotes make?

stof:
-1 for document markers

shrikeh:
Document markers in YAML are important and often overlooked; they are explicit calls that the document begins and ends. For someone who uses YAML all the time, when I see YAML docs without them, it's the same as I'm sure when @stof sees `foo = new stdClass;` - yes, you don't need the brackets after the class name to instantiate the class, because the PHP parser lets you get away with it.

Single quotes are simply the same as per PHP; if something is only a string, it should be wrapped in single quotes to ensure it does not ever evaluate.

stof:
@shrikeh when the file contains a single document (which is always the case for config files, as there is no point putting multiple documents in the config file), I don't see any benefit about using an explicit document rather than a bare document. It only makes things harder to read for people not aware of the full YAML spec.

shrikeh:
*shrugs*

PR #897, opened 2016-04-17
Refactor/scope bootstrap script

ciaranmcnulty:
We create global variables called `$app` and `$autoload`, as well as defining a constant.

There's no actual need to do these things so we can remove them in `3.0.0`

### DIFF
--- a/bin/phpspec
+++ b/bin/phpspec
@@ -1,26 +1,28 @@
 #!/usr/bin/env php
 <?php
 
-define('PHPSPEC_VERSION', '3.0.x@dev');
+call_user_func(function ($version) {
 
-if (is_file($autoload = getcwd() . '/vendor/autoload.php')) {
-    require $autoload;
-} elseif (is_file($autoload = getcwd() . '/../../autoload.php')) {
-    require $autoload;
-}
+    if (is_file($autoload = getcwd() . '/vendor/autoload.php')) {
+        require $autoload;
+    } elseif (is_file($autoload = getcwd() . '/../../autoload.php')) {
+        require $autoload;
+    }
 
-if (is_file($autoload = __DIR__ . '/../vendor/autoload.php')) {
-    require($autoload);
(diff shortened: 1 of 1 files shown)


bin/phpspec
```diff
@@ -1,26 +1,28 @@
 #!/usr/bin/env php
 <?php
 
-define('PHPSPEC_VERSION', '3.0.x@dev');
```
stof: people might be relying on the constant to detect the PhpSpec version in their suite. Is it intended that it is not exposed to the code anymore ?

bin/phpspec
```diff
@@ -1,26 +1,28 @@
...
+        );
+        exit(1);
+    }
 
-$app = new PhpSpec\Console\Application(PHPSPEC_VERSION);
-$app->run();
+    $app = new PhpSpec\Console\Application($version);
+    $app->run();
+
+}, ['3.0.x@dev']);
```
stof: this should not be an array here. Passing an array as version is wrong

ciaranmcnulty: Thanks! I was confusing with `call_user_func_array`.

bin/phpspec
ciaranmcnulty: I couldn't find any usages so would prefer to take it out. It can be added back during beta if extension writers ask - there's probably a better way for us to expose version though I think?

Ocramius: If people need the exact version, there are things like `ocramius/package-versions` instead

bin/phpspec
```diff
@@ -1,26 +1,28 @@
...
+        );
+        exit(1);
+    }
 
-$app = new PhpSpec\Console\Application(PHPSPEC_VERSION);
-$app->run();
+    $app = new PhpSpec\Console\Application($version);
+    $app->run();
+
+}, '3.0.x@dev');
```
Ocramius: I hope you don't have to change this at each release :-P

ciaranmcnulty: Now you mention it, I do do this manually

Issue #899, opened 2016-04-18
Collaborators will throw exception when types in doc comments are not written fully qualified


bestform:
If you are using a PHP IDE, it will probably create a doc comment using imports (with a use statement). Meaning, you define your collaborators in your example method signature and the types are being imported (read: not written fully qualified). When creating a doc comment for the method, the IDE will use those imports and - again - will write the types not fully qualified.

Currently, phpspec will try the type hint in the doc comment first, which it cannot resolve, as it does not have access to the reflection class when parsing the doc comment string. This will result in a ClassNotFound Exception.

If there is no doc comment, it will use the type hint in the method signature, which it can resolve using the imports.

I suggest to switch priorities here and always prefer the type written in the actual method signature, as it will always be possible to resolve it, even if it is being imported.

I have already implemented a solution with a corresponding test case. I will refer to it here.

ciaranmcnulty:
I'm actually considering removing the docblock parsing entirely

bestform:
Yeah, that would be an option, too :)

ciaranmcnulty:
Your change would break backwards compatibility and we have 3.0 coming up but to be honest I don't see a use case for them since we added typehint support (pre-2.0)

ciaranmcnulty:
See #879 by the way

bestform:
I'll leave the PR open if you decide that it would be a good option to keep @param support while ensuring that it won't break if you are using imports

ciaranmcnulty:
This is now closed via #922

PR #900, opened 2016-04-18
Collaborators will take their type from method signature instead of d…

bestform:
…oc comment, if available.

Please see issue https://github.com/phpspec/phpspec/issues/899 for the description of the problem this PR is trying to solve.

### DIFF
--- a/src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
+++ b/src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
@@ -120,8 +120,8 @@ class CollaboratorsMaintainer implements MaintainerInterface
      */
     private function generateCollaborators(CollaboratorManager $collaborators, \ReflectionFunctionAbstract $function, \ReflectionClass $classRefl)
     {
+        $foundClasses = array();
         foreach ($function->getParameters() as $parameter) {
-            $foundClassInSignature = false;
             $collaborator = $this->getOrCreateCollaborator($collaborators, $parameter->getName());
             try {
                 if ($this->isUnsupportedTypeHinting($parameter)) {
@@ -130,7 +130,9 @@ class CollaboratorsMaintainer implements MaintainerInterface
                 if (($indexedClass = $this->getParameterTypeFromIndex($classRefl, $parameter))
                     || ($indexedClass = $this->getParameterTypeFromReflection($parameter))) {
                     $collaborator->beADoubleOf($indexedClass);
-                    $foundClassInSignature = null !== $indexedClass;
+                    if (null !== $indexedClass) {
+                        $foundClasses[] = $parameter->getName();
+                    }
                 }
             }
             catch (ClassNotFoundException $e) {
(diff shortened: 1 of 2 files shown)


bestform:
A check as seen a problem with this approach, so i will close it for the moment until fixed

bestform:
ok, i fixed the issue

bestform:
The php7 build fails. But not because of my code, but with a segfault. This is strange. Any idea what's causing this?

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
```diff
@@ -140,6 +130,9 @@ private function generateCollaborators(CollaboratorManager $collaborators, \Refl
                 if (($indexedClass = $this->getParameterTypeFromIndex($classRefl, $parameter))
                     || ($indexedClass = $this->getParameterTypeFromReflection($parameter))) {
                     $collaborator->beADoubleOf($indexedClass);
+                    if (null !== $indexedClass) {
```
stof: this cannot be ``null`` here, as you are already inside a if checking this

bestform: Oh, right. PHP treats a null assignement as false, right? So, this check is just obsolete. If removed, it will work just as intended.

stof: @bestform the value of an assignment is the value itself. And then, ``null`` is indeed a falsy value when casting it

bestform: @stof This is not strictly related to this PR, but in my opinion, assignments shouldn't be made inside a condition. First of all to avoid mistakes (like doing an assignment when in fact you just want to compare). But also, and this becomes apparent in this code snippet, because of the way, PHP evaluates the value. It is not an identity comparison to false (===) - the 'falsy' part in your description. So this will also evaluate to false for empty strings, or the integer 0 or whatever else PHP is considering 'falsy'. So if you wanted to keep the assignment inside the condition, you should at least do it like that:
`null !== ($foo = $this->bar())`

This is not meant to be a fundamental discussion, I am just interested in your thoughts about that. Is it done like that on purpose here, or would you agree that this is kind of a risky way of writing conditions?

stof: well, I fully agree with you that assignments inside conditions are much harder to read. But this is not the topic here. It does not change the fact that ``$indexedClass`` can never be null here

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
```diff
@@ -120,18 +120,8 @@ public function getPriority()
...
-            $comment = str_replace("\r\n", "\n", $comment);
-            foreach (explode("\n", trim($comment)) as $line) {
-                if (preg_match(self::$docex, $line, $match)) {
-                    $collaborator = $this->getOrCreateCollaborator($collaborators, $match[2]);
-                    $collaborator->beADoubleOf($match[1]);
-                }
-            }
-        }
-
+        $foundClasses = array();
```
stof: this variable name is misleading. It does not contain classes, but collaborator names

bestform: true. I've changed it to $collaboratorsWithAssignedTypes

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
bestform: yes. As i said, just a small off topic note.
I've removed the check. The test is - obviously - still green.

ciaranmcnulty:
It's likely we'll take out the `@param` support but I'll leave this open while we decide

ciaranmcnulty:
See #922

Issue #903, opened 2016-04-20
Separate `--no-interaction` from `--no-code-generation`


kix:
To me it looks like enabling `--no-interaction` should have no effect on `--no-code-generation`. What if I want to always generate the code, without having to answer the interactive questions?

The desired behaviour here is: have PHPSpec always generate code when `--no-interaction` is passed, and skip the code generation when PHPSpec is run with `--no-interaction --no-code-generation`

ciaranmcnulty:
We debated this at the time; the default behaviour of `--no-interaction` is to accept the defaults of every question. 

This is slightly different to 'always choose the option that results in generating code' I guess? Although currently the default answer is to generate in all cases I can think of.

kix:
Hm, perhaps this has been solved in `master`, and I'm using `@stable`. However, even in `master` it looks to me like code generation is always disabled when `--no-interaction` is on: https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Console/ConsoleIO.php#L104

PR #904, opened 2016-04-20
Enable code generation when --no-interaction is on

kix:
This enables code generation even when `--no-interaction` is passed to PHPSpec. The CLI has a separate flag for disabling code generation (that is `--no-code-generation`), so there's no reason to imply that no code should be generated when the input is non-interactive.

Refs #903

### DIFF
--- a/spec/PhpSpec/Console/ConsoleIOSpec.php
+++ b/spec/PhpSpec/Console/ConsoleIOSpec.php
@@ -35,11 +35,11 @@ class ConsoleIOSpec extends ObjectBehavior
         $this->isCodeGenerationEnabled()->shouldReturn(true);
     }
 
-    function it_is_not_code_generation_ready_if_input_is_not_interactive($input)
+    function it_is_code_generation_ready_if_input_is_not_interactive($input)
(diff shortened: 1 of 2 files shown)


stof:
this means that CI server will now run code generation by default, which does not make sense

kix:
@stof, well, yes, this is a BC break, but what are the options? What if I want to save time during local development? I have a continuous runner in the background, and every time I save a spec that references non-existent methods, these methods are automatically generated. Could be a nice timesaver. 
The only other option I see here is that we could add a separate flag to enable code generation explicitly, so that `--no-interaction` does not turn the code generation off when the flag is passed.

pamil:
@kix can't you use `yes 'Y' | phpspec run`?

kix:
@pamil, hm, yes, this could work as well.

ciaranmcnulty:
I'm not going to merge this because I think the behaviour is niche and may be unexpected for some users.

I'd be happy to look at an `--always-generate-code` (or similar phrasing) option that's more explicit about what would happen though

PR #911, opened 2016-05-08
Redirect people to Composer's own instructions on how to install it

adragus-inviqa:


ciaranmcnulty:
Thanks for the contribution!

PR #914, opened 2016-05-09
Add PhpSpec logo in SVG and original Sketch format

everzet:
Original was drawn in [Sketch](https://www.sketchapp.com/) by me, so we should never get any copyright problems.


stof:
does it really make sense to put this in the source code of the tool ? We could create a ``phpspec/assets`` repository instead to store such things

everzet:
@stof my thinking process was that as we already store documentation in the same repo, we might as well store other assets in

stof:
@everzet but the doc is here because it should be versioned together with the code, as it evolves with it.

everzet:
@stof true. @ciaranmcnulty feel free to move this PR to the separate repo if you think it's a better idea.

ciaranmcnulty:
Possibly the static asset can be stored somewhere else? @stof you make a good point.

Would be great to have a page in the docs about the logo and with some sort of specific license.

ciaranmcnulty:
I guess the fact we might make it different in future (albeit no current plans) means it may make sense to version it with the docs?

stof:
@ciaranmcnulty I don't think so. If we change the PhpSpec branding, it should just replace the older logo (we can still find older versions of the logo thanks to the git history of this other repo)

ciaranmcnulty:
I'll make a separate repository..

ciaranmcnulty:
@stof @everzet I finally got around to splitting these commits into a new repo - see phpspec-logo

Issue #908, opened 2016-04-28
No details about certain PHP errors showing in console output


Harrisonbro:
Quite often when using phpspec I see console output similar to `but got [obj:Error]`, particularly when testing that something should throw an exception (so my output is something like `expected exception of class "My\Namespace\...", but got [obj:Error].`)

Is there a way to get phpspec to show me more details about the error that was shown? Currently, I'm having to go and recreate the test in the browser to get the actual error so I can fix it and carry on.

ciaranmcnulty:
Have you tried the different verbosity settings? `-v` or `-vv` ?

Harrisonbro:
Afraid so, it didn't change the output at all.

ciaranmcnulty:
Thanks - I suspect what happens is that we're only riffing them if the exception type is the same - in the cases of `Error` introduced in PHP7 it's probably more important to show the message regardless.

Harrisonbro:
Yep, that makes sense. 
Is that something that needs phpspec itself to be updated (ie. there's no configuration option or anything like that which can make it behave in the way you describe)?
If it does require an update, is there anything I can do to help with that?

ciaranmcnulty:
Yes it's something that will need to be fixed inside PhpSpec - I would start by looking at `PhpSpec\Matcher\ThrowMatcher` (the component that handles `shouldThrow`) and its `verifyPositive` and `verifyNegative` calls.

I'm interested in encouraging new contributors so if you have time and want to help but get stuck please open other issues - I'd love to know where the roadblocks are for new contributors.

Harrisonbro:
Thanks, have done. See https://github.com/phpspec/phpspec/pull/915 and https://github.com/phpspec/phpspec/pull/916

ciaranmcnulty:
Let's leave this open until those PRs are merged

PR #915, opened 2016-05-13
Makes ThrowMatcher show the full error message in the console if the thrown exception is an instance of Error.

Harrisonbro:
Useful since PHP 7's Error class is thrown for a number of problems and it's useful for developers to see the full error message for Error instances for debugging purposes.

Related to https://github.com/phpspec/phpspec/issues/908

### DIFF
--- a/src/PhpSpec/Matcher/ThrowMatcher.php
+++ b/src/PhpSpec/Matcher/ThrowMatcher.php
@@ -121,11 +121,25 @@ class ThrowMatcher implements Matcher
         }
 
         if (!$exceptionThrown instanceof $exception) {
+            $format = 'Expected exception of class %s, but got %s.';
+
+            // Show full error message if the thrown exception is an instance of Error.
+            // See https://github.com/phpspec/phpspec/issues/908
+            if (class_exists('\Error') && $exceptionThrown instanceof \Error) {
+                // TODO: Can we can use $this->presenter->presentValue() rather
+                // than adding the raw Error message in quotes here? Currently
+                // not included since presentValue() truncates the string to 25
+                // characters but we want to show the full Error message for
+                // debugging purposes.
+                $format = 'Expected exception of class %s, but got %s with the message: "%s"';
+            }
+
             throw new FailureException(
                 sprintf(
-                    'Expected exception of class %s, but got %s.',
+                    $format,
                     $this->presenter->presentValue($exception),
-                    $this->presenter->presentValue($exceptionThrown)
+                    $this->presenter->presentValue($exceptionThrown),
+                    $exceptionThrown->getMessage()
                 )
             );
         }
--- a/src/PhpSpec/Matcher/ThrowMatcher.php
+++ b/src/PhpSpec/Matcher/ThrowMatcher.php
@@ -125,7 +125,7 @@ class ThrowMatcher implements Matcher
 
             // Show full error message if the thrown exception is an instance of Error.
             // See https://github.com/phpspec/phpspec/issues/908
-            if (class_exists('\Error') && $exceptionThrown instanceof \Error) {
+            if ($exceptionThrown instanceof \Error) {
                 // TODO: Can we can use $this->presenter->presentValue() rather
                 // than adding the raw Error message in quotes here? Currently
                 // not included since presentValue() truncates the string to 25
--- a/src/PhpSpec/Matcher/ThrowMatcher.php
+++ b/src/PhpSpec/Matcher/ThrowMatcher.php
@@ -123,14 +123,7 @@ class ThrowMatcher implements Matcher
         if (!$exceptionThrown instanceof $exception) {
             $format = 'Expected exception of class %s, but got %s.';
 
-            // Show full error message if the thrown exception is an instance of Error.
-            // See https://github.com/phpspec/phpspec/issues/908
             if ($exceptionThrown instanceof \Error) {
-                // TODO: Can we can use $this->presenter->presentValue() rather
-                // than adding the raw Error message in quotes here? Currently
-                // not included since presentValue() truncates the string to 25
-                // characters but we want to show the full Error message for
-                // debugging purposes.
                 $format = 'Expected exception of class %s, but got %s with the message: "%s"';
             }
 
--- a/spec/PhpSpec/Matcher/ThrowMatcherSpec.php
+++ b/spec/PhpSpec/Matcher/ThrowMatcherSpec.php
@@ -56,4 +56,22 @@ class ThrowMatcherSpec extends ObjectBehavior
 
         $this->negativeMatch('throw', $arr, array('\Error'))->during('ksort', array());
     }
+
+    function it_throws_a_failure_exception_with_the_thrown_exceptions_message_if_a_positive_match_failed(Presenter $presenter)
+    {
+        $actually_thrown_error = new \Error('This is a test Error');
+
+        $callable = function () use ($actually_thrown_error) { throw $actually_thrown_error; };
+
+        $expected_error = new \PhpSpec\Exception\Example\FailureException(
+            'Expected exception of class Exception, but got Error with the message: "This is a test Error"'
+        );
+
+        $incorrectly_matched_exception = new \Exception('This is the exception I expect to be thrown.');
+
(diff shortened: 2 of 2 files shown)


src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -121,11 +121,25 @@ public function verifyPositive($callable, array $arguments, $exception = null)
         }
 
         if (!$exceptionThrown instanceof $exception) {
+            $format = 'Expected exception of class %s, but got %s.';
+
+            // Show full error message if the thrown exception is an instance of Error.
+            // See https://github.com/phpspec/phpspec/issues/908
+            if (class_exists('\Error') && $exceptionThrown instanceof \Error) {
```
stof: ``class_exists``is useless. ``instanceof`` deals fine with non-existent classes (it will obviously return ``false`` in such case)

Harrisonbro: Thanks @stof. I've now removed it (https://github.com/phpspec/phpspec/pull/915/commits/d4485909b120ce3bcde383e3c4c6ef09d13816d1)

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -121,11 +121,25 @@ public function verifyPositive($callable, array $arguments, $exception = null)
         }
 
         if (!$exceptionThrown instanceof $exception) {
+            $format = 'Expected exception of class %s, but got %s.';
+
+            // Show full error message if the thrown exception is an instance of Error.
```
ciaranmcnulty: This is good stuff but you don't need the explanation here IMO, git history's enough.

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -121,11 +121,25 @@ public function verifyPositive($callable, array $arguments, $exception = null)
...
+            // See https://github.com/phpspec/phpspec/issues/908
+            if ($exceptionThrown instanceof \Error) {
+                // TODO: Can we can use $this->presenter->presentValue() rather
```
ciaranmcnulty: Raise an issue for this rather than doing in code ;-)

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -121,11 +121,25 @@ public function verifyPositive($callable, array $arguments, $exception = null)
         }
 
         if (!$exceptionThrown instanceof $exception) {
```
ciaranmcnulty: Maybe now is a good time to rename this variable. It's not really an exception any more right?

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -121,11 +121,25 @@ public function verifyPositive($callable, array $arguments, $exception = null)
...
+            }
+
             throw new FailureException(
                 sprintf(
-                    'Expected exception of class %s, but got %s.',
+                    $format,
                     $this->presenter->presentValue($exception),
-                    $this->presenter->presentValue($exceptionThrown)
+                    $this->presenter->presentValue($exceptionThrown),
+                    $exceptionThrown->getMessage()
```
ciaranmcnulty: I'm not sure how I feel about this silently-ignored parameter... is there a way to make the conditional more explicit?

ciaranmcnulty:
Overall this looks good, I left some comments and would like to see some tests covering this new behaviour.

ciaranmcnulty:
Once you've addressed these comments, please open one PR against the `2.5` branch with the changes (sorry for the confusion, I'm just getting my head around managing this sort of fix)

Harrisonbro:
Sure, will do. Will applying the change to 2.5 not mean it's not included in 3?
Also, could you point me to where I should look for existing tests song can write one for this new code? I did have a look at ThrowMatcherSpec but couldn't make much sense of it so a little guidance to get me started would be handy.

ciaranmcnulty:
No it'll get merged to 2.5 and then merged forwards into 3.0

Harrisonbro:
@ciaranmcnulty Could you point me to where I should look for existing tests song can write one for this new code? I did have a look at `ThrowMatcherSpec` but couldn't make much sense of it so a little guidance to get me started would be handy.

Harrisonbro:
Hi @ciaranmcnulty Just a nudge re. my message above. I want to add tests for this but need a pointer to get started.

Harrisonbro:
@ciaranmcnulty Last nudge on this. I've got time to add these tests if you point out how to get started, but am stuck at the moment. Do let me know if you can.

ciaranmcnulty:
Sorry for lack of response! TBH I don't quite understand the ThrowMatcherSpec either!

I'd trigger the condition and use shouldThrow to check the exception has the right type / message.

if you're not sure about it don't worry - I'll have a go at writing one myself before merging.

ciaranmcnulty:
Reopened so I remember it exists.

src/PhpSpec/Matcher/ThrowMatcher.php
Harrisonbro: Done :) https://github.com/phpspec/phpspec/issues/978

src/PhpSpec/Matcher/ThrowMatcher.php
Harrisonbro: Not in the technical sense, in that it isn't necessarily an `Exception`, but I'd guess most people would consider "exception" to mean that. What's the alternative — `$errorOrException`?!

src/PhpSpec/Matcher/ThrowMatcher.php
Harrisonbro: OK, I've removed.

src/PhpSpec/Matcher/ThrowMatcher.php
Harrisonbro: You mean because the 4th param of `sprintf` might not be used if `$exceptionThrown` is not an instance of `\Error`?

I guess the alternative would be to build the message for `FailureException` within the conditionals and just pass it in (shown below). That does end up with a fair bit of duplication. 

```php
if (!$exceptionThrown instanceof $exception) {
    $failure_message = sprintf(
        'Expected exception of class %s, but got %s.',
        $format,
        $this->presenter->presentValue($exception),
        $this->presenter->presentValue($exceptionThrown)
    );

    if ($exceptionThrown instanceof \Error) {
        $failure_message = sprintf(
            'Expected exception of class %s, but got %s with the message: "%s"',
            $format,
            $this->presenter->presentValue($exception),
            $this->presenter->presentValue($exceptionThrown),
            $exceptionThrown->getMessage()
        );
    }

    throw new FailureException($failure_message);
}
```

An alternative might be to just decided that the error message is always useful (since it's only shown when a test fails and it's prudent to show the developer more information if we can, not less)? That would be as shown below:

```php
if (!$exceptionThrown instanceof $exception) {
    $format = 'Expected exception of class %s, but got %s with the message: "%s".';

    throw new FailureException(
        sprintf(
            $format,
            $this->presenter->presentValue($exception),
            $this->presenter->presentValue($exceptionThrown),
            $exceptionThrown->getMessage()
        )
    );
}
```

Harrisonbro:
Hey @ciaranmcnulty  — OK, finally got my head around it and have added a test. Happy to change/extend it if you have any thoughts. It's late in the day so this is my first draft having only just got it working. Let me know what you think.

Ps. in case anyone else is reading this thread (and for my own future reference) to run phpspec's own tests you need to run `./bin/phpspec run` from the project root.

ciaranmcnulty:
Thanks for the contribution!

Harrisonbro:
Thanks. Do you have a timeline for when this PR will be in a stable release (ie. one that composer will install for me)?

Also, for a few other points we discussed but didn't action and/or other things I might suggest are you happy for me to open a PR for those, even if we just discuss and then reject them, or is there a better way to have discussions around proposed changes?

ciaranmcnulty:
No concrete timeline I'm afraid. you can always tag `dev-master` in Composer.

If you want to address specific refactorings, open a PR.

If there are new features you want to work on, open an issue to discuss.

PR #917, opened 2016-05-16
Updating PHPUnit version

Sam-Burns:
Updates PHPUnit to `5.x` if possible, which will be installed if using PHP 7.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "require-dev": {
         "behat/behat":           "^3.0.11",
         "symfony/filesystem":    "~2.7|~3.0",
-        "phpunit/phpunit":       "~4.4"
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
I'm not sure there's a good reason to disallow 5.0-5.2?

Sam-Burns:
Could change it to `~4.4|~5.0` if you prefer, but it's going to install `5.3` in PHP 7 or `4.8` in PHP 5 anyway, isn't it?

ciaranmcnulty:
I'd prefer that - if someone's tagged phpunit `5.0.4` or something in their composer config they wouldn't be able to install

ciaranmcnulty:
Thanks for stopping this!

PR #918, opened 2016-05-16
Use proc_open rather than passthru to fix console colour problem

### LINKED ISSUE #844: rerun using PassthruRerunner isn't aware the shell supports colours so falls back to non-decorated
Phpspec does a rerun after code generation (using the PassthruRerunner as pcntl isn't enabled). 

This in turn calls phpspec to run again with `passthru(...);` https://github.com/phpspec/phpspec/blob/54b54669a260fda4af471c867baeb680e78c7b1d/src/PhpSpec/Process/ReRunner/PassthruReRunner.php#L54

The problem is that when the Symfony Console app starts again, and the ConsoleOutput is created https://github.com/symfony/console/blob/d09af8f0c981dd381b30522e0939ad7f08c580c0/Application.php#L111

that checks if the stream supports colours
https://github.com/symfony/console/blob/d09af8f0c981dd381b30522e0939ad7f08c580c0/Output/StreamOutput.php#L53-L55

which is _false_ this time around.

We could override this by supplying an `--ansi` flag as an argument to the PassthruReRunner, if the initial ConsoleOutput is decorated, but I don't know how we should pass this to the rerunner.

ciaranmcnulty:
Fixes #844, probably needs testing on Windows

### DIFF
--- a/features/bootstrap/IsolatedProcessContext.php
+++ b/features/bootstrap/IsolatedProcessContext.php
@@ -80,8 +80,8 @@ class IsolatedProcessContext implements Context, SnippetAcceptingContext
      */
     public function iShouldSeeAnErrorAboutTheMissingAutoloader()
     {
-        if (!preg_match('/autoload/', $this->process->getErrorOutput())) {
-            throw new \Exception('There was no error regarding a missing autoloader');
+        if (!preg_match('/autoload/', $this->process->getErrorOutput().$this->process->getOutput())) {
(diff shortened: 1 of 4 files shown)


ciaranmcnulty:
Argh breaks on HHVM

GrahamCampbell:
Keep both implementations, and use the old on on hhvm?

ciaranmcnulty:
The HHVM break seems only partly related, and I can't produce locally :-/

Issue #907, opened 2016-04-26
Improvements to Installation documentation page


ifdattic:
In my opinion the [installation page](http://www.phpspec.net/en/latest/manual/installation.html) is a bit outdated and could be improved/simplified. Here are the changes I would suggest:

* Remove the note about requiring PHP 5.3+. Or change because as far as I can tell the documentation is for the latest version and latest version requires at least 5.6
* Remove *Method #1* as only a single method is available
* Replace composer install with simple "Run the `composer require` command ....."
* Because examples in documentation use the `bin/phpspec` command add the section which tells how to change the bin directory and the documentation uses this directory. That by default the command would be at `vendor/bin/phpspec`
* Instructions for installing composer itself (the link to its installation page) could be left as is.

ciaranmcnulty:
I agree about PHP versions.
I agree about the #1.

I'm unsure about changing all the examples to `vendor/bin/phpspec` as I always set the bin-dir, but we could indeed update the install instructions to use composer commands rather than the JSON

ifdattic:
Might have not translated in my comment, but personally myself I think it's not worth changing all examples to `vendor/bin/phpspec`. That I'm suggesting is just a adding a section explaining that it's possible to change the directory of the executable. If we want to get even fancier we can add that you can make a bash alias to just execute with `phpspec` :)

Sam-Burns:
Fixed in https://github.com/phpspec/phpspec/pull/919

PR #919, opened 2016-05-17
Docs improvements - installation

Sam-Burns:
Changes discussed in https://github.com/phpspec/phpspec/issues/907

### DIFF
--- a/docs/manual/installation.rst
+++ b/docs/manual/installation.rst
@@ -1,23 +1,53 @@
 Installation
 ============
 
-**phpspec** is a php 5.3+ library that you'll have in your project
-development environment. Before you begin, ensure that you have at least
-PHP 5.3.3 installed.
+**phpspec** is a php 5.6+ library that you'll have in your project
+development environment. Before you begin, ensure that you have
+PHP 5.6 or 7 installed.
 
-Method #1 (Composer)
---------------------
+Installation process:
+---------------------
 
-The simplest way to install phpspec with all its dependencies is through
-Composer.
+You can install phpspec with all its dependencies through Composer. Follow
+instructions on `the composer website <https://getcomposer.org/download/>`_ if
+you don't have it installed yet.
 
-First, create a ``composer.json`` file in your project's root directory:
+N.b.: You will need to ensure that your Composer ``autoload`` settings are
+correct. phpspec will not be able to detect classes, even ones it has created,
+unless this is working. This is a common issue which causes confusion when
+installing phpspec.
+
+The ``autoload`` section of your ``composer.json`` file may look something like
+this:
+
+.. code-block:: json
+
+    "autoload": {
+        "psr-0": {
+            "": "src/"
+        }
+    }
+
+Method #1 (Composer command):
+-----------------------------
+
+You can use this Composer command to install phpspec:
+
+.. code-block:: bash
+
+    composer require --dev phpspec/phpspec
+
+Method #2 (Composer config file):
+---------------------------------
+
+If you prefer editing your ``composer.json`` file manually, add phpspec to your
+``require-dev`` section like this:
 
 .. code-block:: js
 
     {
         "require-dev": {
-            "phpspec/phpspec": "~2.0"
+            "phpspec/phpspec": "[your preferred version]"
         },
         "config": {
             "bin-dir": "bin"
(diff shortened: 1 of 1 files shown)


Sam-Burns:
Fixes https://github.com/phpspec/phpspec/issues/907

docs/manual/installation.rst
```diff
@@ -1,23 +1,37 @@
...
 
-First, create a ``composer.json`` file in your project's root directory:
+Method #1 (Composer command):
+-----------------------------
+
+You can use this Composer command to install phpspec:
+
+.. code-block:: bash
+
+    composer require phpspec/phpspec
```
ifdattic: Add the `--dev` flag

docs/manual/installation.rst
```diff
@@ -1,23 +1,37 @@
...
+
+If you prefer editing your ``composer.json`` file manually, add phpspec to your
+``require-dev`` section like this:
 
 .. code-block:: js
 
     {
         "require-dev": {
-            "phpspec/phpspec": "~2.0"
+            "phpspec/phpspec": "*"
```
ifdattic: Star isn't a very good choice for a dependency in my opinion. The better choice would be to make it specific `^2.5` or use a pseudo text `insert-version-requirement`.

But in the end, in my opinion, it would actually be best to remove config file installation completely, too complicated and too much chances for messing something up.

docs/manual/installation.rst
Sam-Burns: Will do

docs/manual/installation.rst
Sam-Burns: I'm pretty sure config file configurers outnumber CLI requirers - it isn't that complicated to test the config file.  I don't want `2.*` though - we're trying to prepare the project for version `3.0` now.  Happy to change it to `"phpspec/phpspec": "[your preferred version]"` if that is preferred though?

ciaranmcnulty:
My concern with this change is that Method #1 will lead to a non-working install in the case of a user not having an autoloader set up.

This is a big FAQ for the project (although it's mitigated recently through the warning when we try and generate the same class twice).

ifdattic:
Wouldn't the same issue happen if the user is using method #2? phpspec not working due to autoloading is an issue, I constantly forget that and it leads to some wasted time before remembering the fix. The solution might be to do like a step based install or add a small section informing about setting up autoloading.

Sam-Burns:
@ifdattic yeah, I will

Sam-Burns:
@ifdattic @ciaranmcnulty Note about autoloading added to installation docs.

ifdattic:
@Sam-Burns I would suggest adding a short snippet of code providing an example of what it means to set autoloading correctly and `phpspec` should be all in lower case

ciaranmcnulty:
@ifdattic we don't actually have a style guide for how to write the name, I should add one

Sam-Burns:
@ifdattic done

docs/manual/installation.rst
```diff
@@ -13,9 +13,16 @@ instructions on `the composer website <https://getcomposer.org/download/>`_ if
...
 unless this is working. This is a common issue which causes confusion when
-installing PHPSpec.
+installing phpspec.
+
+The ``autoload`` section of your ``composer.json`` file may look something like
+this:
+
+.. code-block:: json
+
+    "autoload": {"psr-0": {"": "src"}}
```
ifdattic: This would be better `{"autoload": {"psr-0": {"": "src/"}}}` (and that slash is important)

Pretty print would be even better:

```json
{
  "autoload": {
    "psr-0": {
      "": "src/"
    }
  }
}
```

ifdattic:
@ciaranmcnulty I always just looked at how it's most often used in documentation/website, and assume that it's in lowercase. Think that it would be good to mention somewhere how to write the name, as think had a same comment 1-2 years ago.

@Sam-Burns made some suggestions

ciaranmcnulty:
For now stay lowercase; I'll do an 'official' style guide with maybe the logo src that @everzet  contributed

Sam-Burns:
Ref https://github.com/phpspec/phpspec/pull/933

Sam-Burns:
@ifdattic Reformatting of JSON done :-)

ifdattic:
This is of course personal preference and not worth changing in already current documentation, but think that it might be worth to think about using `#` for headers rather than underlining syntax. Just makes things cleaner (if the header changes no need to modify the line after it).

Now from writing an article about Composer remembered that it has `autoload-dev`which is useful for removing tests from being autoloaded in the production (somehow no testing tools mention that). Might be useful to add a mention about it, but at the same time not sure about adding it as installation page is getting bigger and might start giving a view of being a complex tool (putting new users away).

@Sam-Burns looks good 👍

Sam-Burns:
@ifdattic I'll do a separate PR for you, changing the format of all the headers.

The `autoload-dev` feature of Composer isn't relevant to what we're doing here.  You don't need to use Composer to make PHPSpec spec classes autoloadable - it is doing that itself with reference to the config file, or default settings relating to spec prefixes, folder paths etc.

docs/manual/installation.rst
```diff
@@ -1,23 +1,53 @@
...
+installing phpspec.
+
+The ``autoload`` section of your ``composer.json`` file may look something like
+this:
+
+.. code-block:: json
+
+    "autoload": {
+        "psr-0": {
+            "": "src/"
```
ciaranmcnulty: What's the benefit of the trailing slash? it's inconsistent with the other method below

Sam-Burns: I added it because I was asked to by another contributor (https://github.com/phpspec/phpspec/pull/919#discussion_r64147357), but I don't believe it makes any difference either way.  I'll make sure the two things are consistent.

ciaranmcnulty: @ifdattic What's the benefit of the trailing slash?

ifdattic: No benefit, it works the same either way

Sam-Burns:
@ciaranmcnulty Last commit makes autoloading examples consistent in both places.

ciaranmcnulty:
Thanks

PR #920, opened 2016-05-17
Bump version in composer desc

Sam-Burns:


Sam-Burns:
Does this fix the packagist description automatically, or does someone have to log on and do it?

ciaranmcnulty:
I think packagist reads the composer.json from the specific tag when a release happens

👍

PR #921, opened 2016-05-17
Updating authors

Sam-Burns:
Where credit's due

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,10 @@
         {
             "name":      "Marcello Duarte",
             "homepage":  "http://marcelloduarte.net/"
+        },
+        {
+            "name":      "Ciaran McNulty",
+            "homepage":  "https://ciaranmcnulty.com/"
         }
     ],
 


ciaranmcnulty:
Not sure having multiple authors is correct; there aren't just 3 of us.

Sam-Burns:
Well, there aren't just two... Would you prefer to see the `authors` section removed?  I'd rather it at least mentions the lead maintainer if it's staying.

ciaranmcnulty:
What do other projects do?

henrikbjorn:
I have seen other projects, just listen Contributors as the authors.

shanethehat:
Maybe authors should be anyone with write access?

Sam-Burns:
I think @ciaranmcnulty is just too modest to push the merge button on this one.  Current `authors` section is a bit misleading though, to be fair.  Maybe current and previous leads should be mentioned, as well as 'various contributors'?

stof:
As @ciaranmcnulty is the lead now, I think it is logical to have him in the list here

ciaranmcnulty:
It appears the `authors` key is optional

ciaranmcnulty:
Very well

PR #924, opened 2016-05-18
Minor tidy-up

Sam-Burns:


src/PhpSpec/Listener/CollaboratorMethodNotFoundListener.php
```diff
@@ -184,7 +184,7 @@ private function getRealArguments($prophecyArguments)
 
     /**
      * @param ExampleEvent $event
-     * @return bool
+     * @return bool|\Exception
```
stof: this never returns a boolean. It returns ``\Exception|null``

src/PhpSpec/Locator/ResourceManager.php
```diff
@@ -18,14 +18,14 @@
     /**
      * @param string $query
      *
-     * @return Resource[]
+     * @return \PhpSpec\Locator\Resource[]
```
stof: not needed as the class is in the current namespace and so the short name works already

Issue #926, opened 2016-05-19
Document line number locators


ciaranmcnulty:
Seems to be missing from the docs and also from the Help output:

`phpspec run path/to/MySpec.php:53` - runs spec on line 53

mheki:
I can take care of it

ciaranmcnulty:
@mheki great!

PR #927, opened 2016-05-19
Document line number locators

mheki:
Adds missing docs and `run` help output. Fixes: https://github.com/phpspec/phpspec/issues/926


ciaranmcnulty:
Great! thanks

Issue #928, opened 2016-05-19
Allow to describe classes with leading backslash


mheki:
This issue quite annoys me and can be easily fixed.

My use case:
- I want to describe a new file with a long namespace (other classes of that ns already exist)
- I copy the namespace from the existing file (e.g. `\Lorem\Ipsum\Vehicle\Business\VeryLongNamespace`)
- paste it and add my new class name into `bin/phpspec d "\Lorem\Ipsum\Vehicle\Business\VeryLongNamespace\MySuperCoolClass"`
- I always have to go to the beginning of the namespace to remove the leading backslash

Error message:
```
[InvalidArgumentException]
  String "(...)" is not a valid class name.
  Please see reference document: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-0.md```

The fix is already written, will create a PR if that's OK.

ciaranmcnulty:
Sounds reasonable!

Would be suitable for `2.5` branch too

ciaranmcnulty:
fixed in #931

PR #929, opened 2016-05-19
Allow to describe a class providing a namespace with leading backslash

mheki:
Fixes issue: https://github.com/phpspec/phpspec/issues/928


ciaranmcnulty:
Manually merged, thanks

PR #930, opened 2016-05-19
Bumping dev dependencies

Sam-Burns:


composer.json
```diff
@@ -32,9 +32,9 @@
     },
 
     "require-dev": {
-        "behat/behat":           "^3.0.11",
-        "symfony/filesystem":    "~2.7|~3.0",
-        "phpunit/phpunit":       "~4.4|~5.0"
+        "behat/behat":           "^3.1",
+        "symfony/filesystem":    "^3.0",
+        "phpunit/phpunit":       "^4.8|^5.3"
```
stof: you could even require using PHPUnit 5 all the time, as the min requirement is now PHP 5.6

PR #931, opened 2016-05-19
Allow to describe a class providing a namespace with leading backslash

mheki:
Fixes: #928

Conflicts:
 - spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php

### DIFF
--- a/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
+++ b/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
@@ -455,6 +455,16 @@ class PSR0LocatorSpec extends ObjectBehavior
         $resource->getSpecClassname()->shouldReturn('spec\Console\ApplicationSpec');
     }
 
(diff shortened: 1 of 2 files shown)


stof:
:+1: 

@ciaranmcnulty the 5.3.3 build should be removed from the 2.5 branch IMO. This Travis image is just unusable as it does not have openssl

ciaranmcnulty:
@stof yeah probably if there's no sane solution.

stof:
@ciaranmcnulty the same solution is to stop using the 5.3.3 environment.
Testing against the latest 5.3 is enough guarantee IMO. Other minor versions of PHP are only tested against their latest release (or at least the latest at the time of the Travis environment update...)

ciaranmcnulty:
@stof 5.3.3 is a special case where we also test against the lowest-preferred composer deps. As it's impractical I'm removing it though :-/

ciaranmcnulty:
Thanks @mheki!

PR #935, opened 2016-05-25
DI (option 2, phase 1)

Sam-Burns:
Introducing Container-Interop; renaming container to ServiceLocator

### DIFF
--- a/spec/PhpSpec/Container/ContainerExceptionSpec.php
+++ b/spec/PhpSpec/Container/ContainerExceptionSpec.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace spec\PhpSpec\Container;
+
+use Interop\Container\Exception\ContainerException;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+class ContainerExceptionSpec extends ObjectBehavior
+{
+    function it_is_a_runtime_exception()
+    {
+        $this->shouldHaveType(\RuntimeException::class);
+    }
+
+    function it_is_a_container_interop_compliant_exception()
+    {
+        $this->shouldHaveType(ContainerException::class);
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -18,17 +18,18 @@
     ],
 
     "require": {
-        "php":                      "^5.6|^7.0",
-        "phpspec/prophecy":         "~1.5",
-        "phpspec/php-diff":         "~1.0.0",
-        "sebastian/exporter":       "~1.0",
-        "symfony/console":          "~2.7|~3.0",
-        "symfony/event-dispatcher": "~2.7|~3.0",
-        "symfony/process":          "^2.7|~3.0",
-        "symfony/finder":           "~2.7|~3.0",
-        "symfony/yaml":             "~2.7|~3.0",
-        "doctrine/instantiator":    "^1.0.1",
-        "ext-tokenizer":            "*"
+        "php":                                 "^5.6|^7.0",
+        "phpspec/prophecy":                    "~1.5",
+        "phpspec/php-diff":                    "~1.0.0",
+        "sebastian/exporter":                  "~1.0",
+        "symfony/console":                     "~2.7|~3.0",
+        "symfony/event-dispatcher":            "~2.7|~3.0",
+        "symfony/process":                     "^2.7|~3.0",
+        "symfony/finder":                      "~2.7|~3.0",
+        "symfony/yaml":                        "~2.7|~3.0",
+        "doctrine/instantiator":               "^1.0.1",
+        "container-interop/container-interop": "^1.1",
+        "ext-tokenizer":                       "*"
     },
 
     "require-dev": {
--- a/features/extensions/developer_uses_extension.feature
+++ b/features/extensions/developer_uses_extension.feature
@@ -15,7 +15,7 @@ Feature: Developer uses extension
     namespace Example1\PhpSpec\MatcherExtension;
 
     use PhpSpec\Extension as PhpSpecExtension;
-    use PhpSpec\ServiceContainer;
+    use PhpSpec\Container\ServiceContainer;
 
     class Extension implements PhpSpecExtension
     {
--- a/spec/PhpSpec/Container/ServiceContainerSpec.php
+++ b/spec/PhpSpec/Container/ServiceContainerSpec.php
@@ -1,11 +1,17 @@
 <?php
 
-namespace spec\PhpSpec;
+namespace spec\PhpSpec\Container;
 
+use Interop\Container\ContainerInterface;
 use PhpSpec\ObjectBehavior;
 
 class ServiceContainerSpec extends ObjectBehavior
 {
+    function it_is_standards_compliant()
+    {
+        $this->shouldHaveType(ContainerInterface::class);
+    }
+
     function it_stores_parameters()
     {
         $this->setParam('some_param', 42);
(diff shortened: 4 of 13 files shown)


src/PhpSpec/Container/ServiceLocator.php
```diff
@@ -11,15 +11,16 @@
...
 
+use Interop\Container\ContainerInterface;
 use InvalidArgumentException;
 
 /**
  * The Service Container is a lightweight container based on Pimple to handle
  * object creation of PhpSpec services.
  */
-class ServiceContainer
+class ServiceLocator implements ContainerInterface
```
ciaranmcnulty: 👎 Service Locator is how a container may be used, not a thing that it is

spec/PhpSpec/Container/ContainerExceptionSpec.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace spec\PhpSpec\Container;
+
+use Interop\Container\Exception\ContainerException;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+class ContainerExceptionSpec extends ObjectBehavior
+{
+    function it_is_an_invalid_argument_exception()
```
shanethehat: InvalidArgument exception != RuntimeException

stof:
what is the benefit of implement the container-interop interface on our custom container ? all code is specific to the implementation anyway, because PhpSpec does not need a service locator, but a DIC container which can configured (the container is never injected anywhere as a service locator AFAIK)

ciaranmcnulty:
@stof container is injected into the Application, that's the only place I can think of where it's used as a service locator (there may be more lurking somewhere).

I discussed the following strategy with @Sam-Burns:

  1. make our container implement container-interop
  2. make extensions able to use container-interop delegation mechanism
  3. big swap of container used internally for 'something else' 

For me 2. is the most important, in the sense that it means extension authors don't need to worry about the mess - it's not clear to me whether the container-interop delegation system requires 1. first? If not maybe it can be optional

Anyhow, input welcome

stof:
@ciaranmcnulty the delegation mechanism does not allow to customize existing services easily (it allows to let another extension define the service) and does not support overriding services.
So IMO, it is not the right option to provide extensions. It would make them less powerful than today.

ciaranmcnulty:
I'm happy to see what @Sam-Burns comes up with then mercilessly reject the PR if it doesn't work ;-)

Sam-Burns:
The container delegate mechanism does allow services to be overridden, which is why it is a popular choice for providing an extension point for large libraries.

stof:
@Sam-Burns being able to overwrite services with a different implementation of the same interface is a very useful extension point

ciaranmcnulty:
My understanding is that if you stack services in the right order it can work to replace existing interfaces? I've not delved into container-interop enough to be an authority but rather than discuss whether it's possible or not let's see some working code.

stof:
and btw, container-interop won't work either for the existing use case of collecting services to build an array (of commands for instance) AFAIK, which is currently handled by doing a prefix-based lookup, and could be handled by tags if switching to the Symfony DIC

Sam-Burns:
Container-Interop worked just fine in a previous spike, and does nothing to prevent the use of a service ID to represent an array of other services. The use of the current container as a service locator and as a registry is something we need to think about seriously about reducing, and the delegate container feature will provide a perfect opportunity for extension authors to provide their own implementations of our interfaces.

I'll provide a second PR, based on these changes, writing out the service location and allowing the use of delegate containers as an extension point, to demonstrate what I'm talking about.

stof:
@Sam-Burns Look at any place using ``getByPrefix``, and then try to see how an extension using container-interop would do to hook into them. All these places are important extension points. And PhpSpec itself cannot know the service ids it should retrieve there (this is precisely the point of retrieving all services starting with a given prefix), making delegation unsuitable.

ciaranmcnulty:
@stof wait for the working (or not) code and we can critique then ;-)

ciaranmcnulty:
@stof I think the container-interop approach is to have a meta-service that returns a collection, then extensions can wrap/modify that collection to add new services.

It's not as clean as `getbyPrefix` or a symfony tag + compiler pass but it is simpler

stof:
@ciaranmcnulty I reviewed container-interop (I had several hours-long in-person discussion with one of the people behind it and other PSR-11 experiments) and the delegation-based system relies on the fact that you ask for a precise id.

> then extensions can wrap/modify that collection to add new services.

You cannot, as delegation-based lookup does not allow decorating services (only one of the container can define the service, and others are not considered). This is one of the reason why other ways are explored for PSR-11

ciaranmcnulty:
I admit my knowledge is not great in that area.

I'd really ask though that we let @Sam-Burns have an attempt to solve some of these problems and show if it works.

I'm not committed to container-interop at all BTW, but I want to decouple extensions from our container (and not make them rely on another specific container too) if we can.

stof:
> I'm not committed to container-interop at all BTW, but I want to decouple extensions from our container (and not make them rely on another specific container too) if we can.

Given that the only goal of the extension class is to configure the container, this looks quite hard. You need to give them an API to configure the container (and container-interop is not such an API btw, which is why I think it does not fit here).

On a side note, designing the container for the extension system with just a PR on this repo won't work, as this repo does not dogfood its extension API to configure the container and so won't let you notice places which don't work.
This is the big difference with Behat 3, where core features are actually implemented as a bunch of extensions which are always registered, and so working on Behat itself forced to have enough power in extensions (and this is why Behat extensions access the Symfony DIC API directly rather than abstracting it btw).

src/PhpSpec/Container/ServiceLocator.php
```diff
@@ -69,30 +70,27 @@ public function getParam($id, $default = null)
...
      */
-    public function set($id, $value)
+    public function set($serviceId, $value)
     {
         if (!is_object($value) && !is_callable($value)) {
-            throw new InvalidArgumentException(sprintf(
-                'Service should be callable or object, but %s given.',
-                gettype($value)
-            ));
+            ServiceNotFound::constructFromServiceId($serviceId);
```
ciaranmcnulty: I'd be happier with `fromServiceId` or similar

src/PhpSpec/Container/ServiceLocator.php
```diff
@@ -69,30 +70,27 @@ public function getParam($id, $default = null)
...
      * every time get is called.
      *
-     * @param string          $id
+     * @param string          $serviceId
      * @param object|callable $value
      *
      * @throws \InvalidArgumentException if service is not an object or callable
      */
-    public function set($id, $value)
+    public function set($serviceId, $value)
```
ciaranmcnulty: I don't think it needs renaming to conform to the interface

src/PhpSpec/Container/ServiceLocator.php
```diff
@@ -148,12 +146,12 @@ public function get($id)
     }
 
     /**
-     * @param $id
+     * @param string $serviceId
      * @return bool
      */
-    public function isDefined($id)
+    public function has($serviceId)
```
ciaranmcnulty: Add this for an alias in the first set of changes (until we are sure what happens with extensions)

Sam-Burns: Yes, I'll change it.

src/PhpSpec/Container/ServiceLocator.php
stof: IMO, the name ``PhpSpec\ServiceContainer`` should be kept for now, until we know what extensions should look like, to minimize changes

Sam-Burns:

PR #938, opened 2016-05-30
DI (option 2, phase 2)

Sam-Burns:

Moves config management out of container and application object, and puts it in a separate config manger class.

Though valuable on its own, this is also a prerequisite for other changes to the dependency injection situation, to be proposed shortly.

This PR is based on a previous one, still open at time of writing.  The proposed changes which are unique to this PR can be viewed in the [branch diff](https://github.com/Sam-Burns/phpspec/compare/di-option-2-phase-1...Sam-Burns:di-option-2-phase-2).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,17 +18,18 @@
     ],
 
     "require": {
-        "php":                      "^5.6|^7.0",
-        "phpspec/prophecy":         "~1.5",
-        "phpspec/php-diff":         "~1.0.0",
-        "sebastian/exporter":       "~1.0",
-        "symfony/console":          "~2.7|~3.0",
-        "symfony/event-dispatcher": "~2.7|~3.0",
-        "symfony/process":          "^2.7|~3.0",
-        "symfony/finder":           "~2.7|~3.0",
-        "symfony/yaml":             "~2.7|~3.0",
-        "doctrine/instantiator":    "^1.0.1",
-        "ext-tokenizer":            "*"
+        "php":                                 "^5.6|^7.0",
+        "phpspec/prophecy":                    "~1.5",
+        "phpspec/php-diff":                    "~1.0.0",
+        "sebastian/exporter":                  "~1.0",
+        "symfony/console":                     "~2.7|~3.0",
+        "symfony/event-dispatcher":            "~2.7|~3.0",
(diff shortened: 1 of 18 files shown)


Sam-Burns:
The proposed changes which are unique to this PR can be viewed here: https://github.com/Sam-Burns/phpspec/compare/di-option-2-phase-1...Sam-Burns:di-option-2-phase-2

composer.json
```diff
@@ -18,17 +18,18 @@
...
-        "symfony/yaml":             "~2.7|~3.0",
-        "doctrine/instantiator":    "^1.0.1",
-        "ext-tokenizer":            "*"
+        "php":                                 "^5.6|^7.0",
+        "phpspec/prophecy":                    "~1.5",
+        "phpspec/php-diff":                    "~1.0.0",
+        "sebastian/exporter":                  "~1.0",
+        "symfony/console":                     "~2.7|~3.0",
+        "symfony/event-dispatcher":            "~2.7|~3.0",
+        "symfony/process":                     "^2.7|~3.0",
```
GrahamCampbell: Why is that we have a `^` here, but `~` elsewhere. `~2.7` is the same as `^2.7`.

Sam-Burns: More than happy to make this consistent within the file, but I think it's better if I do that in a separate PR.

GrahamCampbell: Agreed. It was just something I noticed when I read the diff. :)

Sam-Burns: IIRC `^2.7` is the convention, isn't it?  I'll do a PR in a bit.

GrahamCampbell: Yeh, that's the current convention. It's the semver version of `~`.

ciaranmcnulty: Indeed `~2.7` is the same as `^2.7` so what's the point in talking about it?

ciaranmcnulty: I already closed #891

GrahamCampbell: I'm confused as why you're fine having this "typo" in here?

stof:
This proposal looks weird to me. You are basically creating 2 different holders depending on where the params are coming from. The config file should be here to configure the DI container rather than configuring another object needing to be used to retrieve the configuration of services.

Sam-Burns:

PR #939, opened 2016-05-30
DI (option 2, phase 3)

Sam-Burns:

Creates a new Console Manager object.  Using this, removes all usages of `ServiceContainer::set()` from the `Application` class.

Though (possibly) valuable on its own, this is also a prerequisite for other changes to the dependency injection situation, to be proposed shortly.

This PR is based on previous ones, still open at time of writing. The proposed changes which are unique to this PR can be viewed in the [branch diff](https://github.com/Sam-Burns/phpspec/compare/di-option-2-phase-2...Sam-Burns:di-option-2-phase-3)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,17 +18,18 @@
     ],
 
     "require": {
-        "php":                      "^5.6|^7.0",
-        "phpspec/prophecy":         "~1.5",
-        "phpspec/php-diff":         "~1.0.0",
-        "sebastian/exporter":       "~1.0",
-        "symfony/console":          "~2.7|~3.0",
-        "symfony/event-dispatcher": "~2.7|~3.0",
-        "symfony/process":          "^2.7|~3.0",
-        "symfony/finder":           "~2.7|~3.0",
-        "symfony/yaml":             "~2.7|~3.0",
-        "doctrine/instantiator":    "^1.0.1",
-        "ext-tokenizer":            "*"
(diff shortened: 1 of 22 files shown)


stof:
What is the value of this ConsoleManager ? This looks like a separate container for only some of the services instead of using the normal container.

Sam-Burns:
The value is that using the service locator as a registry, and setting things to it in the application object, is a pretty serious code smell.  By using the `ConsoleManager` as a smaller, more manageable registry, we get closer to actual dependency injection.

stof:
@Sam-Burns what we have today is a service container, which is getting configured.
The fact that the Input service is not instantiated by the container itself is an implementation detail today, while your PR forces any service needing to consume it to know about the fact that it is special instead of being available as a service.
I consider the current code cleaner than your proposal in this regard.

Sam-Burns:
This PR removes four uses of `ServiceContainer::set()` from inside the `Application` class.  Do you have an alternative suggestion as to how to get rid of those, so we can stop using the container as a registry pattern and service locator?

stof:
@Sam-Burns this ``set`` calls are part of the container configuration, because we are required to use the existing Input and Output objects rather than instantiating it inside the container.
note that the Symfony DI component aknowledges this use case too. This is exactly what synthetic services are for.

Sam-Burns:

PR #940, opened 2016-05-31
fix printing the last row when remainder is zero

bendavies:
Before:
screen shot 2016-05-31 at 11 43 32

After:
screen shot 2016-05-31 at 11 42 35


ciaranmcnulty:
looks good! Can you reopen the PR against 2.5 please, so it can be a bug fix on that branch?

Thanks

PR #941, opened 2016-05-31
DI (option 2, phase 4)

Sam-Burns:
Depends on: https://github.com/phpspec/phpspec/pull/939

Removes final impediments to using DI for all production code classes, including the `Application` class and the `RunCommand` and `DescribeCommand` classes.

Though (arguably) valuable on its own, this is also a prerequisite for other changes to the dependency injection situation, to be proposed shortly.

This PR is based on previous ones, still open at time of writing. The proposed changes which are unique to this PR can be viewed in the [branch diff](https://github.com/Sam-Burns/phpspec/compare/di-option-2-phase-3...Sam-Burns:di-option-2-phase-4).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,17 +18,18 @@
     ],
 
     "require": {
-        "php":                      "^5.6|^7.0",
-        "phpspec/prophecy":         "~1.5",
-        "phpspec/php-diff":         "~1.0.0",
-        "sebastian/exporter":       "~1.0",
-        "symfony/console":          "~2.7|~3.0",
-        "symfony/event-dispatcher": "~2.7|~3.0",
-        "symfony/process":          "^2.7|~3.0",
-        "symfony/finder":           "~2.7|~3.0",
-        "symfony/yaml":             "~2.7|~3.0",
-        "doctrine/instantiator":    "^1.0.1",
-        "ext-tokenizer":            "*"
+        "php":                                 "^5.6|^7.0",
+        "phpspec/prophecy":                    "~1.5",
+        "phpspec/php-diff":                    "~1.0.0",
+        "sebastian/exporter":                  "~1.0",
+        "symfony/console":                     "~2.7|~3.0",
+        "symfony/event-dispatcher":            "~2.7|~3.0",
+        "symfony/process":                     "^2.7|~3.0",
+        "symfony/finder":                      "~2.7|~3.0",
+        "symfony/yaml":                        "~2.7|~3.0",
+        "doctrine/instantiator":               "^1.0.1",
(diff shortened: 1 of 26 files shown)


stof:
The commit moving the registration of suite locators to PrioritizedResourceManager looks wrong to me:

- it is not the goal of the PrioritizedResourceManager to configure itself. This is the opposite of DI.
- this does not allow extensions to register extra locators from other places.

Removing existing extension points looks weird to me.

the commit https://github.com/Sam-Burns/phpspec/commit/918e797047f4aefaa556e05d97190dad02cc4683 changes the behavior of phpspec, which looks suspicious to me.

and https://github.com/Sam-Burns/phpspec/commit/ad040ae484876a851fc33c37fbf568884aeca01d looks very suspicious to me, as it makes the RunCommand responsible to configure parts of the container instead of just using it.

Sam-Burns:
There is no removal of an extension point, although the way that extensions interact with the application will change during these improvements, to be more standards-compliant.
The commit involving the shutdown process does not change the behaviour of PHPSpec.
The `RunCommand` was already calling `configure()` on the service locator, causing three different steps of service addition and alteration.  Clearly the `configure()` method is a bad thing, and getting rid of it means that the `RunCommand` is now container agnostic, and doesn't alter the configuration of the container.

stof:
> to be more standards-compliant.

container-interop is not a standard. It is a proposal for a standard (and from the discussion about PSR-11, it is unlikely to become the standard btw, as many people have expressed concerns about it being unsuitable for usage in frameworks, defeating the purpose of a FIG standard).

ciaranmcnulty:
`container-interop` is a standard, it's not a FIG standard ;-)

I caught up on the PSR-11 conversation over the weekend - it's worth looking also at the proposed mechanisms there.

@Sam-Burns I am finding the amount of PRs overwhelming, especially as some commits are contained in 5 PRs. Can you reduce it to one please?

stof:
> container-interop is a standard, it's not a FIG standard ;-)

Well OK.

but then, note that it is a service location standard, not a DI container standard (despite what it claims to be).
and PhpSpec needs a DI container to provide extension points (being geared towards a service locator API is what makes container-interop unable to support some use cases, including the most common PhpSpec extension point. See the phase 5 PR failing to support it)

Sam-Burns:
@ciaranmcnulty Fine, I'll close the previous PRs and just leave one open, if that's easier for you.

ciaranmcnulty:
Yeah whichever has most relevant discussion, push to there and drop the others.

Sam-Burns:

PR #942, opened 2016-05-31
Di option 2 phase 5

Sam-Burns:
Depends on: https://github.com/phpspec/phpspec/pull/939
Depends on: https://github.com/phpspec/phpspec/pull/941

The container is now Container-Interop compliant.  This PR removes all implementation-specific method calls on the container.  The extension mechanism also changes a little, as evident from the changes to the relevant Behat feature file.  Further, more significant, changes to the extension mechanism now become possible, and will be implemented shortly.

Though (arguably) valuable on its own, this is also a prerequisite for other changes to the dependency injection situation, to be proposed shortly.

This PR is based on previous ones, still open at time of writing. The proposed changes which are unique to this PR can be viewed in the [branch diff](https://github.com/Sam-Burns/phpspec/compare/di-option-2-phase-4...Sam-Burns:di-option-2-phase-5).


features/extensions/developer_uses_extension.feature
```diff
@@ -24,9 +24,16 @@ Feature: Developer uses extension
          */
         public function load(ServiceContainer $container)
         {
+            // Add new matcher to container
             $container->set('matchers.seven', function (ServiceContainer $c) {
                 return new BeSevenMatcher($c->get('formatter.presenter'));
             });
+
+            // Add new matcher to list of matchers in container
+            $matchers = $container->get('phpspec.matchers');
```
stof: this is not a suitable way to extend an array of matchers. It breaks lazyness of definitions, which will then forbid overriding any dependency of these matchers after this particular extension was registered.

You need to keep a clear distinction between the configuration phase of the container (where you define service definitions) and the runtime phase (where you can retrieve services from the container).
Mixing them together leads to broken behaviors and very hard to debug issues (and the container defined in PhpSpec does not implement any safeguard to warn you, as it was inspired by Pimple 1 which suffered from the same limitations).

stof:
OK, so now you finally have the proof of what I said days ago: container-interop does not work for service decoration (your change triggers an infinite loop).
And btw, this is one of the big issue with container-interop (not the only one btw).

stof:
@Sam-Burns please avoid using a github mention in the commit message, especially when you rebase the branch. It sends me a notification email each time you push a rebased commit (as this is a new commit)

features/bootstrap/ApplicationContext.php
```diff
@@ -58,15 +59,25 @@ public function setupApplication()
 
     private function setupPrompter()
     {
-        $this->prompter = new Prompter();
+        $this->prompter = $prompter = new Prompter();
```
ciaranmcnulty: Instantiating outside the container seems wrong here.

spec/PhpSpec/Console/ManagerSpec.php
```diff
@@ -0,0 +1,31 @@
...
+namespace spec\PhpSpec\Console;
+
+use PhpSpec\Console\Manager;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ManagerSpec extends ObjectBehavior
```
ciaranmcnulty: I'd rather you avoid refactoring related code as part of this change

src/PhpSpec/Extension.php
```diff
@@ -13,14 +13,16 @@
 
 namespace PhpSpec;
 
+use UltraLite\Container\Container;
```
ciaranmcnulty: Presumably we don't actually want to couple to this container

composer.json
```diff
@@ -18,17 +18,19 @@
...
+        "phpspec/php-diff":                    "~1.0.0",
+        "sebastian/exporter":                  "~1.0",
+        "symfony/console":                     "~2.7|~3.0",
+        "symfony/event-dispatcher":            "~2.7|~3.0",
+        "symfony/process":                     "^2.7|~3.0",
+        "symfony/finder":                      "~2.7|~3.0",
+        "symfony/yaml":                        "~2.7|~3.0",
+        "doctrine/instantiator":               "^1.0.1",
+        "container-interop/container-interop": "^1.1",
+        "ultra-lite/container":                "^3.1",
```
ciaranmcnulty: this should only be a dev dependency I think

Sam-Burns: I'm sorry, I'm not sure how that would work, unless there was another Container-Interop compliant container you would like us to use as PhpSpec's own container?  I chose this one because the (obviously unstandardised) configuration process was similar to the internal one it replaces, but I'm open to other suggestions.

src/PhpSpec/Extension.php
Sam-Burns: That's true, although the PR isn't quite at the point where we can remove this interim measure just yet.  This is because I haven't put a composite container.  Hoping to get there very soon though.

spec/PhpSpec/Console/ManagerSpec.php
Sam-Burns: As would I, and I'll think again to see if there is a way of sorting out the container situation without dealing with the existing IO snafu, but I wasn't able to come up with one originally.  Very happy to try again though, to see if the same advantages can be conferred by some slimmer changeset.

Sam-Burns:
Closing in favour of https://github.com/phpspec/phpspec/pull/947 .

PR #943, opened 2016-06-05
Use extension to detect PHP version

ciaranmcnulty:
I wrote an extension with this sort of testing in mind; replaces the existing mechanism for skipping tests based on PHP version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,7 @@ install:
   - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
-before_script:
-  - echo "<?php if (defined('HHVM_VERSION')) { echo ',@hhvm'; } else { if (PHP_VERSION_ID >= 50400) echo ',@php5.4'; if (PHP_VERSION_ID >= 70000) echo ',@php7'; }" > php_version_tags.php
-
 script:
    - bin/phpspec run --format=pretty
    - ./vendor/bin/phpunit --testdox
-   - ./vendor/bin/behat --format=pretty --tags '~@php-version'`php php_version_tags.php`
+   - ./vendor/bin/behat --format=pretty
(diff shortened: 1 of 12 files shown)


ciaranmcnulty:
Anyone know what's up with Scrutiniser?

Sam-Burns:
@ciaranmcnulty Yeah, we need a new version of the `bossa/phpspec2-expects` thing. I'll see what I can do.

mheki:
@Sam-Burns https://github.com/BossaConsulting/phpspec2-expect/pull/25

stof:
@ciaranmcnulty loçoks like it considers the version of the root package as being dev-2.5, which does not make sense. I suspect they have a bug in their code handling cyclic composer dependencies

ciaranmcnulty:
Maybe should report it to Composer :-/

stof:
The issue is not in composer IMO, but in the value which is set by Scrutinizer in the ``COMPOSER_ROOT_VERSION`` constraint. I suspect they just set it as ``dev-$BRANCH``, without taking versionned branches into account.

ciaranmcnulty:
You're right, I get the same error output if I try `COMPOSER_ROOT_VERSION=dev-2.5 composer install
` this will probably be a problem ongoing then :-/

ciaranmcnulty:
Assuming scrutiniser can be ignored, I'd appreciate feedback

stof:
@ciaranmcnulty you should report it to Scrutinizer though.

A workaround might be to add a branch alias from ``dev-2.5`` to ``2.5.x-dev`` in the composer.json (with a comment in ``_comment`` explaining that this is to workaround a scrutinizer bug)

stof:
Looks good to me otherwise

ciaranmcnulty:
Thanks, I wasn't sure about using an egotistically-named extension :-P

Might be useful for similar cases in Behat?

ciaranmcnulty:
I emailed scrutiniser support

stof:
have you tried my suggestion for a workaround ?

ciaranmcnulty:
I backported the fix from master that removes the circular dependency

PR #945, opened 2016-06-05
fix printing the last row when remainder is zero

bendavies:
https://github.com/phpspec/phpspec/pull/940


ciaranmcnulty:
Thanks!

Issue #870, opened 2016-02-20
Easier custom matchers


ciaranmcnulty:
Currently new matchers can be registered inline in a Spec, or by an Extension.

The overhead of writing an extension is too great so we get questions about matcher reuse fairly regularly.

We could introduce a new config option:

```
matchers:
  - MyApp\FooMatcher
  - MyApp\Bar\BazMatcherProvider
```

Aslong as those classes were autoloadable and either `MatcherInterface` or `MatcherProviderInterface` we would auto register them

PR #949, opened 2016-06-08
Registered configured custom matchers

gquemener:
Hello everyone,

I'm starting a piece of contribution related to https://github.com/phpspec/phpspec/issues/870, if that's ok for you.

I like the idea of registering custom matchers that can be reused across specs, which is already possible to achieve with inline matcher and inheritance by the way, but a bit painfull.

The solution provided in this PR is to register the custom matchers in the container using the `matchers` prefix which is then used by the assembler to populate the `MatcherManager`.

I have a few questions:

  - What about overriding PhpSpec core matchers? Is it wishable? I don't really think so as it can lead to confusion.
  - A presenter is injected in all of the core matchers. Is it wishable to do the same with the custom matchers (see below)? 
  - Should it be a 3.0 feature or should it also be implemented in 2.5?

### Reflection
We could infer the ctor parameter and auto-inject the presenter.
```php
$id = strtolower(str_replace('Matcher', '', substr($class, strrpos($class, '\\') + 1)));
$container->set(sprintf('matchers.%s', $id), function (ServiceContainer $c) use ($class) {
    $reflClass = new \ReflectionClass($class);
    $args = [];
    foreach ($reflClass->getConstructor()->getParameters() as $parameter) {
        $class = $parameter->getClass();

        if ($class->implementsInterface(Presenter::class)) {
            $args[] = $c->get('formatter.presenter');
        }
    }

    return $reflClass->newInstanceArgs($args);
});
```

### Be Specific
 Another solution is to be specific in the configuration and define for each custom matcher which service to inject into it like so:
``` yaml
matchers:
    - App\Matcher\FooMatcher:
        formatter.presenter
``` 

Any toughts welcome, of course :wink:

### DIFF
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -147,6 +147,13 @@ class Application extends BaseApplication
 
                     $extension->load($container);
                 }
+            } elseif ('matchers' === $key && is_array($val)) {
+                foreach ($val as $class) {
+                    $id = strtolower(str_replace('Matcher', '', substr($class, strrpos($class, '\\') + 1)));
+                    $container->set(sprintf('matchers.%s', $id), function (ServiceContainer $c) use ($class) {
+                        return new $class();
+                    });
+                }
             } else {
                 $container->setParam($key, $val);
             }
--- a/features/matchers/developer_uses_custom_matcher.feature
+++ b/features/matchers/developer_uses_custom_matcher.feature
@@ -1,0 +1,90 @@
+Feature: Developer uses custom matcher
+  As a Developer
+  I want a custom matcher
+  In order to confirm any custom assertion I need
+
+  Scenario: Succesfully register a custom matcher
+    Given the spec file "spec/Matchers/Custom/MovieSpec.php" contains:
+    """
+    <?php
+
+    namespace spec\Matchers\Custom;
+
+    use PhpSpec\ObjectBehavior;
+    use Prophecy\Argument;
+
+    class MovieSpec extends ObjectBehavior
+    {
+        function it_should_have_some_specific_options_by_default()
+        {
+            $this->getEntries()->shouldTotalize(130);
+        }
+    }
+    """
+
+    And the class file "src/Matchers/Custom/TotalizeMatcher.php" contains:
+    """
+    <?php
+
+    namespace Matchers\Custom;
+
+    use PhpSpec\Matcher\BasicMatcher;
+    use PhpSpec\Exception\Example\FailureException;
+
+    class TotalizeMatcher extends BasicMatcher
+    {
+        public function supports($name, $subject, array $arguments)
+        {
+            return 'totalize' === $name &&
+                is_array($subject) &&
+                isset($arguments[0]) &&
+                is_int($arguments[0])
+            ;
+        }
+
+        protected function matches($subject, array $arguments)
+        {
+            return array_sum($subject) === $arguments[0];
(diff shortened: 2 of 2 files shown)


ciaranmcnulty:
Thanks @gquemener, as this is a new feature it should be in 3.0 (so PR to `master is correct). This also means you have a brief window to not worry about BC as much.

I'll review later - the thing about the presenter will take some thought.

src/PhpSpec/Console/Application.php
```diff
@@ -147,6 +147,13 @@ protected function loadConfigurationFile(InputInterface $input, ServiceContainer
 
                     $extension->load($container);
                 }
+            } elseif ('matchers' === $key && is_array($val)) {
+                foreach ($val as $class) {
+                    $id = strtolower(str_replace('Matcher', '', substr($class, strrpos($class, '\\') + 1)));
```
ciaranmcnulty: Classes don't have to have namespaces, so you need to handle the case where `strpos` here returns `false`.

I don't really see the benefit here of assuming that the string `'Matcher'` will be in the output - it won't hurt if it's in the `$id`

avant1: I'm not sure if removing namespace completely is a good idea.
When you use only matchers defined in your project - it would be OK. 

But if you use third-party matchers - class names (without namespaces) can be equal, and if you use two matchers with the same name the first will be overwritten. That can lead to issues that are hard to understand.

avant1: Also we can check if given classes exist and implement `PhpSpec\Matcher\Matcher` interface.
For now, if given matcher is not `PhpSpec\Matcher\Matcher`, `phpspec` silently exists with non-zero status code.

gquemener: My purpose was to avoid silently overriding matchers.

Removing the `Matcher` suffix is only esthetic, and does not serve any real purpose when I look back at it.

In fact, the id could be the class name provided in the configuration (ie: `matchers.App\Matcher\FooMatcher`), I think it should work (and would speed up the registration process a bit).

stof:
Note that this break BC as it does not allow to define a ``matchers`` parameter anymore.
And as extension are configured by defining global parameters, we have no idea what params are defined.

IMO, we would need a cleaner structure for the config files if we want to start introducing such special config settings

stof:
btw, this answers the question about 2.5 vs 3.0: it cannot go in 2.x

ciaranmcnulty:
@stof there was a suggestion to scope config to the extension, how behat does it. Maybe now is a good time to do that

gquemener:
Indeed, that's a quite valid point @stof.

"Wrapping" the phpspec related configuration into a `phpspec` key would be a better choice IMO.

ciaranmcnulty:
That said we haven't been worried about BC when introducing other configs

gquemener:
What do you think about https://github.com/phpspec/phpspec/pull/955 ?

ciaranmcnulty:
I'd rather something like #189 (which we should do before `3.0.0` if we do at all)

ciaranmcnulty:
See #958

ciaranmcnulty:
As we are in beta for 3.0.0 this may wait for 3.1.0 now... but there's no reason that can't be soon.

toby-griffiths:
Just come across this thread as I'm trying to get the presenter in a custom matcher, but it looks like this isn't possible.

Is there any way to do this currently?

Issue #946, opened 2016-06-06
Not possible to infer available matchers


ciaranmcnulty:
Relates to #809 

It's currently not possible for an IDE or tool to infer from a Matcher what string(s) would be allowed to match. This makes it hard for any sort of autocompletion in an IDE.

It may make sense to remove the `supports` logic from the `Matcher` interface and externalise it (e.g. it could return a regex)

ciaranmcnulty:
@stof @deeky666 this is relevant to the conversation about IDEs doing better autocomplete - currently it's not possible to tell what method names will match

stof:
Another solution would be to add some special phpdoc on matchers.
We have done this in Symfony for events, to allow the PhpStorm plugin and the Drupal doc to identify events, and know the associated event class. See https://github.com/symfony/symfony/blob/24e08e9657f46846d79050af8b6b3d8cdfef0f62/src/Symfony/Component/HttpKernel/KernelEvents.php#L28

We could have a ``@phpspec-matcher haveCount`` on the class (allowing multiple ones to account for matchers supporting multiple naming), to document that this matcher provides support for ``shouldHaveCount`` and ``shouldNotHaveCount`` (the fact that matchers are applied through ``should*`` and ``shouldNot*`` is something the IDE plugin should handle internally IMO).
Inline matchers would not need such phpdoc btw. The IDE just need to look at the keys of the array being returned

Returning a regex would probably make it hard for IDEs, as they would have to reverse-engineer the regex to build matching strings when providing autocompletion.
And I don't thing we need a regex for core matchers. they don't do crazy things requiring it AFAIK.

ciaranmcnulty:
That sounds like a good solution, as it doesn't affect our own BC / APIs.

I guess that means this issue isn't something we need to worry about until someone actually starts working on this functionality.

deeky666:
This sounds like a good idea because that can also support custom matchers then I suppose.

stof:
yeah, matchers provided by extensions will work, as long as the extension author applies the convention too

deeky666:
Hmmm thinking about it again this approach still does not work for PhpStorm as it requires the stub methods in the code. Having a phpspec specific annotation doesn't give us the ability to have PhpStorm recognize the magic methods. We do require `@method`. Please see also [this ticket](https://youtrack.jetbrains.com/issue/WI-22881).

PR #951, opened 2016-06-10
Add missing magic matcher methods

deeky666:
Adds all missing `@method` tags to `ObjectBehaviour` for "static" matcher methods to improve DX (IDE autocompletion).
I'm not sure if this addition is desired but as there are already some present in the class, it should be all or nothing IMO. As long as a [real PhpStorm plugin solution is not possible](https://youtrack.jetbrains.com/issue/WI-22881), this helps at least for most of the matchers. Let me know what you guys think about it ;)

### DIFF
--- a/src/PhpSpec/ObjectBehavior.php
+++ b/src/PhpSpec/ObjectBehavior.php
@@ -31,9 +31,25 @@ use ArrayAccess;
  * @method void beConstructedWith(...$arguments)
  * @method void beConstructedThrough($factoryMethod, array $constructorArguments = array())
  * @method void beAnInstanceOf($class)
+ *
  * @method void shouldHaveType($type)
+ * @method void shouldNotHaveType($type)
+ * @method void shouldBeAnInstanceOf($type)
+ * @method void shouldNotBeAnInstanceOf($type)
  * @method void shouldImplement($interface)
+ * @method void shouldNotImplement($interface)
+ *
  * @method Subject\Expectation\DuringCall shouldThrow($exception = null)
+ * @method Subject\Expectation\DuringCall shouldNotThrow($exception = null)
+ *
+ * @method void shouldHaveCount($count)
+ * @method void shouldNotHaveCount($count)
+ *
+ * @method void shouldHaveKeyWithValue($key, $value)
+ * @method void shouldNotHaveKeyWithValue($key, $value)
+ *
+ * @method void shouldHaveKey($key)
+ * @method void shouldNotHaveKey($key)
  */
 abstract class ObjectBehavior implements
     ArrayAccess,
--- a/src/PhpSpec/Wrapper/Subject.php
+++ b/src/PhpSpec/Wrapper/Subject.php
@@ -20,6 +20,88 @@ use PhpSpec\Wrapper\Subject\ExpectationFactory;
 use PhpSpec\Util\Instantiator;
 use ArrayAccess;
 
+/**
+ * @method void shouldHaveType($type)
+ * @method void shouldNotHaveType($type)
+ * @method void shouldReturnAnInstanceOf($type)
+ * @method void shouldNotReturnAnInstanceOf($type)
+ * @method void shouldBeAnInstanceOf($type)
+ * @method void shouldNotBeAnInstanceOf($type)
+ * @method void shouldImplement($interface)
+ * @method void shouldNotImplement($interface)
+ *
+ * @method void shouldBe($value)
+ * @method void shouldNotBe($value)
+ * @method void shouldBeEqualTo($value)
+ * @method void shouldNotBeEqualTo($value)
+ * @method void shouldReturn($value)
+ * @method void shouldNotReturn($value)
+ * @method void shouldEqual($value)
+ * @method void shouldNotEqual($value)
+ *
+ * @method void shouldBeLike($value)
+ * @method void shouldNotBeLike($value)
+ *
+ * @method void shouldHaveCount($count)
+ * @method void shouldNotHaveCount($count)
+ *
+ * @method void shouldBeArray()
+ * @method void shouldNotBeArray()
+ * @method void shouldBeBool()
+ * @method void shouldNotBeBool()
+ * @method void shouldBeBoolean()
+ * @method void shouldNotBeBoolean()
+ * @method void shouldBeCallable()
+ * @method void shouldNotBeCallable()
+ * @method void shouldBeDouble()
+ * @method void shouldNotBeDouble()
+ * @method void shouldBeFloat()
+ * @method void shouldNotBeFloat()
+ * @method void shouldBeInt()
+ * @method void shouldNotBeInt()
+ * @method void shouldBeInteger()
+ * @method void shouldNotBeInteger()
+ * @method void shouldBeLong()
+ * @method void shouldNotBeLong()
+ * @method void shouldBeNull()
+ * @method void shouldNotBeNull()
+ * @method void shouldBeNumeric()
+ * @method void shouldNotBeNumeric()
+ * @method void shouldBeObject()
+ * @method void shouldNotBeObject()
+ * @method void shouldBeReal()
+ * @method void shouldNotBeReal()
+ * @method void shouldBeResource()
+ * @method void shouldNotBeResource()
+ * @method void shouldBeScalar()
+ * @method void shouldNotBeScalar()
+ * @method void shouldBeString()
+ * @method void shouldNotBeString()
+ * @method void shouldBeNan()
+ * @method void shouldNotBeNan()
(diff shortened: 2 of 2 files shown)


ciaranmcnulty:
I worry if this is going too far, although as you say having half of them in there doesn't make sense.

deeky666:
@ciaranmcnulty yep, I could also live without having the method tags at all in the code and include those in my own code but then we should remove the others, otherwise the topic will likely come up over ans over again and it's hard to argue against adding more then... Think about it and tell me your decision, I will update the PR eventually then :)

src/PhpSpec/ObjectBehavior.php
```diff
@@ -31,9 +31,90 @@
  * @method void beConstructedWith(...$arguments)
  * @method void beConstructedThrough($factoryMethod, array $constructorArguments = array())
  * @method void beAnInstanceOf($class)
+ *
  * @method void shouldHaveType($type)
+ * @method void shouldNotHaveType($type)
+ * @method void shouldReturnAnInstanceOf($type)
```
ciaranmcnulty: Return doesn't make sense to me when talking about `$this`

src/PhpSpec/ObjectBehavior.php
```diff
@@ -31,9 +31,90 @@
...
+ * @method void shouldBeLike($value)
+ * @method void shouldNotBeLike($value)
+ *
  * @method Subject\Expectation\DuringCall shouldThrow($exception = null)
+ * @method Subject\Expectation\DuringCall shouldNotThrow($exception = null)
+ *
+ * @method void shouldHaveCount($count)
+ * @method void shouldNotHaveCount($count)
+ *
+ * @method void shouldBeArray()
```
ciaranmcnulty: These scalar things don't make sense on `$this`, which is always an object

src/PhpSpec/ObjectBehavior.php
```diff
@@ -31,9 +31,90 @@
...
+ * @method void shouldBeString()
+ * @method void shouldNotBeString()
+ * @method void shouldBeNan()
+ * @method void shouldNotBeNan()
+ * @method void shouldBeFinite()
+ * @method void shouldNotBeFinite()
+ * @method void shouldBeInfinite()
+ * @method void shouldNotBeInfinite()
+ *
+ * @method void shouldContain($value)
```
ciaranmcnulty: Don't work on objects...

ciaranmcnulty:
OK so having thought about it it's good to add all the variations, but all the stuff that refers to scalar types (the straight shouldBeArray stuff, the shouldContain stuff) doesn't make sense when referring to an object under test.

src/PhpSpec/ObjectBehavior.php
deeky666: @ciaranmcnulty right so. Can you give me a hint then where to add the magic methods for method/property expectations? Because basically all of those calls are delegated to `Subject` and from there to `Matcher` objects no matter if it is an expectation on the described class, one of its methods or properties (if I understand the voodoo correctly).

ciaranmcnulty: There's nowhere you can add them that will work.

deeky666: I see two options how this could work:

1. Through `@method` tags for methods of the described class in the particular specification

```php
class Foo
{
    public function bar()
    {
        return 'foobar';
    }
}

/**
 * @method \PhpSpec\Wrapper\Subject bar()
 */
class FooSpec extends ObjectBehaviour
{
    function it_should_bar()
    {
        $this->bar()->shouldBe('foobar');
    }
}
```

2. Through an IDE plugin that already knows the method delegations of the described class and knows that those methods return `Subject` (I already have such a plugin for PhpStorm that supports this). So if we had the matcher magic methods in the `Subject` class then, autocompletion could work for most cases. Get what I mean?

ciaranmcnulty: In your case a) shouldBe() will never autocomplete, is the point

option b) is the only workable solution IMO

deeky666: Case a) would autocomplete if we had `shouldBe()` as `@method` tag in `Subject` or am I missing something here? Also case b) would also need `shouldBe()` as `@method` tag in `Subject`.

ciaranmcnulty: It's not a method on subject though - it's added dynamically at runtime by a matcher.

deeky666: Sure I know that but that's exactly the whole point of having `@method` annotations for magic calls through `__call()`...

src/PhpSpec/ObjectBehavior.php
stof: These things should be on the Subject class indeed, not here

src/PhpSpec/ObjectBehavior.php
ciaranmcnulty: If I wanted to hard-code a list of matchers into the Subject, we'd add methods! We're never going to be able to annotate matchers that extensions add, or that users add in getMatchers...

deeky666: That is also true. It won't work for custom matchers and it won't work for matchers that depend on method names of the described class. What I wanted to achieve with this PR is to at least add those that are officially bundled anyways (and will most probably be the most used). But if you do not like this idea then feel free to close this PR :) I can totally understand your standpoint... But currently the only possible way to get some autocompletion with a PhpStorm plugin is to have those magic method annotation in the `Subject` class, because currently we cannot "add" those dynamically.

stof: @deeky666 but most builtin matchers make no sense on the ObjectBehavior, as they are about matching expectations on method return values, not on the SUT instance itself. Most of the things you are adding make no sense here.

On ObjectBehavior, we should add only these ones (on top of the existing ones):

- ``shouldNotHaveType``
- ``shouldBeAnInstanceOf``
- ``shouldNotBeAnInstanceOf``
- ``shouldNotImplement``
- ``shouldNotThrow``

and maybe these ones to account for Countable or ArrayAccess SUT:
- ``shouldHaveCount``
- ``shouldNotHaveCount``
- ``shouldHaveKeyWithValue``
- ``shouldNotHaveKeyWithValue``
- ``shouldHaveKey``
- ``shouldNotHaveKey``

Everything else should be removed.

deeky666: @stof agreed on that already. That was indeed a mistake of mine. But I would still like to wait for further opinions about having the other magic methods in `Subject` making expectations on SUT methods/properties. IMO it does not make sense having some in `ObjectBehaviour` but none in `Subject` (this seems weird). So as I stated before it should be all or nothing. The argument to say matchers are added at runtime and therefore should not be hard-coded / annotated also applies to `ObjectBehaviour` then...

ciaranmcnulty: I have no objection to adding them all to `Subject` in that case, it makes life slightly easier for users..

stof: Adding them in Subject would not help at all, because your IDE will not know that ``$this->testedMethod()`` will return a Subject. We cannot add magic phpdoc for that. Knowing it in the IDE would require to build a phpspec integration in the IDE to know it. And then, such an integration could support matchers properly rather than relying on ``@method`` in the Subject phpdoc, ensuring support for non-core matchers

stof: matchers added at runtime are much more like to apply on return values of methods btw

deeky666:
@ciaranmcnulty @stof okay redid the PR according to the discussion.

src/PhpSpec/ObjectBehavior.php
deeky666: @stof I am aware of that and those magic methods on `Subject` are intended to help building IDE support via a plugin. As I told before already, I already have a WIP plugin for PhpStorm which collects the SUT methods and makes them available via `$this`. Those methods then are typed to return `Subject`. The current problem is that we are stuck at this place because we do not have the available matcher methods as those are not present in the parse tree (missing `@method` annotations). And there currently is no way to add stub elements dynamically without having them somewhere in the code. There it is mandatory to have those. Also I am not sure how it would be possible to "support matchers properly rather than relying on `@mathod`" as we do not the matcher method names, arguments and return types. How would you infer that dynamically?

stof: discussion about finding matchers for IDEs should happen in https://github.com/phpspec/phpspec/issues/946

ciaranmcnulty:
Cool, thanks

deeky666:
@ciaranmcnulty thank you for merging!

PR #953, opened 2016-06-13
[3.0 migration guide] Fixed `PhpSpec\Extension`

gnugat:
removed duplicated line, and fixed the actual new FQCN.


ciaranmcnulty:
Hm this actually looks wrong at http://www.phpspec.net/en/latest/manual/upgrading-to-phpspec-3.html

Any idea how to properly escape these?

gnugat:
Probably using back ticks instead of asterisks

gnugat:
Actually, it's double backticks (because we use RST). There's a working example in [Prophet Objects](https://raw.githubusercontent.com/phpspec/phpspec/master/docs/manual/prophet-objects.rst)

ciaranmcnulty:
Great, thanks

PR #955, opened 2016-06-14
[WIP] Wrapped PhpSpec config in a phpspec keyword

gquemener:


ciaranmcnulty:
I don't think having 'non-phpspefc' stuff in here makes sense; I'll take a look at whether #189 is fixable to merge

PR #958, opened 2016-06-19
Scope extension parameters to relevant extension

ciaranmcnulty:
This is based on #189 by @docteurklein  - I've left his commit in the history for posterity but could squash.

Flips the extension format from:

```yaml
some_extension_config: foo

extensions:
    - SomeExtension
```

To:

```yaml
extensions:
    SomeExtension:
        some_config:foo
```

This will give us more freedom in future to define new config parameters in the root config, as it's clear they aren't for Extension usage


pamil:
:+1:, but I guess we can unify the naming - there are both `$config` and `$params` now.

ciaranmcnulty:
@pamil Good point, `$config` makes most sense to distinguish from container parameters

PR #960, opened 2016-06-20
Possibility to test if a php error is triggered

### LINKED ISSUE #957: Support for testing triggered error
This is kind of related to #366, but not completely. IT would be nice I think to test that a call on a method triggers error messages (such as deprecations), or on construct methods (with a `duringInstanciation`).

Something along the lines of `$this->shouldTrigger(E_USER_DEPRECATED, $message)->during()` ? Unless something equivalent is already done (I've been searching but found nothing so far...) ?

Taluu:
Should fix #957, and add helpers such as `$this->shouldTrigger()`.

If nothing is provided as arguments, it validates if any error at all is triggered (or none at all, if `shouldNotTrigger` is used). If a `$level` and / or `$message` is provided, it should filter them and check that the triggered error tested against these parameters matches before validating or not the match.

- [x] Add TriggerMatcher
- [x] Add support for `trigger` methods in `ExpectationFactory`
- [x] Add `PositiveTrigger` and `NegativeTrigger`... even though it looks a lot like the `throw` counterpart.
- [x] Add some specs (I can think of some more to add, such as testing the filters parts or going against the predictions, making them fail ?)
- [x] Add some behat tests ?
- [x] Add some documentation

I was first going for the handling of phpspec's `ErrorException` and use the `ThrowMatcher` (so just creating a `ErrorException` exception, and then feed it to `PositiveThrow` and `NegativeThrow`), but as the `ErrorException` is currently made (only creating a string, with filename and line number), it looked difficult to do so and build a correct exception.

So I went with the long way, with more or less duplicating a lot of stuff from the `Throw` counterparts, registering a temporary error handler to do the job.

As this is a kind of a new feature, I targetted the master branch. Open for discussion :}

### DIFF
--- a/features/matchers/developer_uses_trigger_matcher.feature
+++ b/features/matchers/developer_uses_trigger_matcher.feature
@@ -1,0 +1,119 @@
+Feature: Developer uses trigger matcher
+  As a Developer
+  I want a trigger matcher
+  In order to validate triggered exceptions against my expectations
+
+  Scenario: Checking if a deprecated error has been triggered
+    Given the spec file "spec/Matchers/TriggerExample1/FooSpec.php" contains:
+    """
+    <?php
+
+    namespace spec\Matchers\TriggerExample1;
+
+    use PhpSpec\ObjectBehavior;
+    use Prophecy\Argument;
+
+    class FooSpec extends ObjectBehavior
+    {
+        function it_triggers_an_error_when_calling_something_deprecated()
+        {
+            $this->shouldTrigger(E_USER_DEPRECATED)->duringDoDeprecatedStuff();
+        }
+    }
+    """
+    And the class file "src/Matchers/TriggerExample1/Foo.php" contains:
+    """
+    <?php
+
+    namespace Matchers\TriggerExample1;
+
+    class Foo
+    {
+        public function doDeprecatedStuff()
+        {
+            trigger_error('Foo', E_USER_DEPRECATED);
+        }
+    }
+    """
+    When I run phpspec
+    Then the suite should pass
+
+  Scenario: Checking that a deprecated error has the right message
+    Given the spec file "spec/Matchers/TriggerExample2/FooSpec.php" contains:
+    """
+    <?php
+
+    namespace spec\Matchers\TriggerExample3;
+
+    use PhpSpec\ObjectBehavior;
+    use Prophecy\Argument;
+
+    class FooSpec extends ObjectBehavior
+    {
+        function it_triggers_a_specific_deprecated_error_when_calling_deprecated_method()
+        {
+            $this->shouldTrigger(E_USER_DEPRECATED, 'This is deprecated')->duringDoDeprecatedStuff();
+        }
+    }
+    """
+    And the class file "src/Matchers/TriggerExample2/Foo.php" contains:
+    """
+    <?php
+
+    namespace Matchers\TriggerExample2;
+
+    class Foo
+    {
+        public function doDeprecatedStuff()
+        {
+            trigger_error('This is deprecated', E_USER_DEPRECATED);
+        }
+    }
+    """
+    When I run phpspec
+    Then the suite should pass
+
+  Scenario: "Trigger" alias matches using the trigger matcher and let the code continue afterwards
+    Given the spec file "spec/Matchers/TriggerExample3/FooSpec.php" contains:
+    """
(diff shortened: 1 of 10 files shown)


ciaranmcnulty:
I've been thinking about your use case - this only really works for execution-halting errors, right?

If we're talking about testing things like deprecation warnings, I'm not sure this will work out because execution inside the method being tested will halt?

Taluu:
Shouldn't be, as as i am not stopping it. And in my uss case too it
shouldn't be stopping either, just verifying that this error has been
triggered

Taluu:
@ciaranmcnulty would it help if I added in the spec something that tells me "catching the triggered error does *not* the execution" ?

I'll add behat tests if everything is alright (or should I drop this ?)

ciaranmcnulty:
This would definitely need tests to be merged; I'm also not confident that it works ;-)

Taluu:
Added behat tests, and I had to add a `unpackArgs` method. Sorry for the delay :}

/cc @ciaranmcnulty

features/matchers/developer_uses_trigger_matcher.feature
```diff
@@ -0,0 +1,163 @@
...
+    <?php
+
+    namespace spec\Matchers\TriggerExample2;
+
+    use PhpSpec\ObjectBehavior;
+    use Prophecy\Argument;
+
+    class FooSpec extends ObjectBehavior
+    {
+        function it_triggers_an_errpr_when_calling_something_deprecated()
```
ciaranmcnulty: Typo in `error` here

features/matchers/developer_uses_trigger_matcher.feature
```diff
@@ -0,0 +1,163 @@
...
+        {
+            $this->shouldTrigger(E_USER_DEPRECATED)->duringDoDeprecatedStuff();
+        }
+    }
+    """
+
```
ciaranmcnulty: Please don't have whitespace inbetween steps (here and other places in the feature files)

features/matchers/developer_uses_trigger_matcher.feature
```diff
@@ -0,0 +1,163 @@
+Feature: Developer uses trigger matcher
+  As a Developer
+  I want a trigger matcher
+  In order to validate triggered exceptions against my expectations
+
+  Scenario: "Trigger" alias matches using the trigger matcher with explicit method name
```
ciaranmcnulty: Can you please rephrase the titles to describe more what this lets the developer do?

Example here: "Checking a deprecated error is triggered"

features/matchers/developer_uses_trigger_matcher.feature
```diff
@@ -0,0 +1,163 @@
...
+        {
+            trigger_error('Foo', E_USER_DEPRECATED);
+        }
+    }
+    """
+
+    When I run phpspec
+    Then the suite should pass
+
+  Scenario: "Trigger" alias matches using the trigger matcher with implicit method name
```
ciaranmcnulty: I don't think this scenario's necessary - you're testing that the `during*` magic works, which is the responsibility of `DuringCall` (which I don't think you've modified).

You can just use one syntax in the scenarios.

src/PhpSpec/Matcher/TriggerMatcher.php
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @param callable    $callable
+     * @param array       $arguments
+     * @param int|null    $level
+     * @param string|null $message
+     *
+     * @throws \PhpSpec\Exception\Example\FailureException
+     */
+    public function verifyPositive($callable, array $arguments, $level = null, $message = null)
```
ciaranmcnulty: You can typehint `callable` here - it's supported in all versions we support now (older code may be missing it)

features/matchers/developer_uses_trigger_matcher.feature
ciaranmcnulty: Probably overtested in the other place ;-)

Taluu: Removed and reworded the scenario. :}

Taluu:
I added some basic documentation and updated the PR with your comments.

ciaranmcnulty:
Awesome stuff, thanks @Taluu!

PR #961, opened 2016-06-20
Refactor/simplify container

ciaranmcnulty:
Bearing in mind there's likely to be some changes in how extensions operate, I've taken the opportunity to simplify the container usage, chiefly to introduce an interface but also to mark areas likely to be volatile as `@internal`.

Having the interface should enable us to change the internally-used container, as well as support more container types in extensions in future via adapters.

The only part of the interface I particularly dislike is the `getByPrefix` which always felt hacky, but it's behaviour should be simple enough to emulate in other containers.

### DIFF
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -26,8 +26,10 @@ use RuntimeException;
 
 /**
  * The command line application entry point
+ *
+ * @Internal
  */
-class Application extends BaseApplication
+final class Application extends BaseApplication
 {
     /**
      * @var ServiceContainer
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -20,7 +20,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Yaml\Yaml;
-use PhpSpec\ServiceContainer;
+use PhpSpec\ServiceContainer\IndexedServiceContainer;
 use PhpSpec\Extension;
 use RuntimeException;
 
@@ -32,7 +32,7 @@ use RuntimeException;
 final class Application extends BaseApplication
 {
     /**
-     * @var ServiceContainer
+     * @var IndexedServiceContainer
      */
     private $container;
 
@@ -41,12 +41,12 @@ final class Application extends BaseApplication
      */
     public function __construct($version)
     {
-        $this->container = new ServiceContainer();
+        $this->container = new IndexedServiceContainer();
         parent::__construct('phpspec', $version);
     }
 
     /**
-     * @return ServiceContainer
+     * @return IndexedServiceContainer
      */
     public function getContainer()
     {
@@ -127,11 +127,11 @@ final class Application extends BaseApplication
 
     /**
      * @param InputInterface   $input
-     * @param ServiceContainer $container
+     * @param IndexedServiceContainer $container
      *
      * @throws \RuntimeException
      */
-    protected function loadConfigurationFile(InputInterface $input, ServiceContainer $container)
+    protected function loadConfigurationFile(InputInterface $input, IndexedServiceContainer $container)
     {
         $config = $this->parseConfigurationFile($input);
 
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -15,6 +15,7 @@ namespace PhpSpec\Console;
 
 use PhpSpec\Loader\StreamWrapper;
 use PhpSpec\Process\Context\JsonExecutionContext;
+use PhpSpec\ServiceContainer;
 use Symfony\Component\Console\Application as BaseApplication;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -27,7 +28,7 @@ use RuntimeException;
 /**
  * The command line application entry point
  *
- * @Internal
+ * @internal
  */
 final class Application extends BaseApplication
 {
@@ -46,7 +47,7 @@ final class Application extends BaseApplication
     }
 
     /**
-     * @return IndexedServiceContainer
+     * @return ServiceContainer
      */
     public function getContainer()
     {
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -75,7 +75,7 @@ final class Application extends BaseApplication
 
         $this->loadConfigurationFile($input, $this->container);
 
-        foreach ($this->container->getByPrefix('console.commands') as $command) {
+        foreach ($this->container->getByTag('console.commands') as $command) {
             $this->add($command);
         }
 
@@ -84,7 +84,7 @@ final class Application extends BaseApplication
         $this->container->get('console.io')->setConsoleWidth($this->getTerminalWidth());
 
         StreamWrapper::reset();
-        foreach ($this->container->getByPrefix('loader.resource_loader.spec_transformer') as $transformer) {
+        foreach ($this->container->getByTag('loader.resource_loader.spec_transformer') as $transformer) {
             StreamWrapper::addTransformer($transformer);
         }
         StreamWrapper::register();
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -66,7 +66,7 @@ final class Application extends BaseApplication
         $this->container->set('console.output', $output);
         $this->container->set('console.helper_set', $helperSet);
 
-        $this->container->setShared('process.executioncontext', function () {
+        $this->container->define('process.executioncontext', function () {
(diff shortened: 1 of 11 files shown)


Sam-Burns:
Scrutinizer's a bit miffed about coding standards; otherwise, definitely a :+1:

everzet:
1. Now as your set/get doesn't transform values when fetching them, you don't need separate setParameter/getParameter
2. define() method name is a bit confusing. I'd prefer factory(), factoryFor(), initialiseVia() or something like that

ciaranmcnulty:
@everzet I'm checking that services are objects at the moment... I would find mixing parameters and services a bit weird...

Factory might be better than Definition but there isn't a handy verb for it.

src/PhpSpec/Console/Application.php
```diff
@@ -20,17 +20,19 @@
...
 use Symfony\Component\Yaml\Yaml;
-use PhpSpec\ServiceContainer;
+use PhpSpec\ServiceContainer\IndexedServiceContainer;
 use PhpSpec\Extension;
 use RuntimeException;
 
 /**
  * The command line application entry point
+ *
+ * @Internal
```
stof: should be ``@internal`` in lowercase to match phpdoc

stof: same everywhere

src/PhpSpec/Console/Application.php
```diff
@@ -39,12 +41,12 @@ class Application extends BaseApplication
...
     public function __construct($version)
     {
-        $this->container = new ServiceContainer();
+        $this->container = new IndexedServiceContainer();
         parent::__construct('phpspec', $version);
     }
 
     /**
-     * @return ServiceContainer
+     * @return IndexedServiceContainer
```
stof: shouldn't it keep document the interface ?

src/PhpSpec/ServiceContainer.php
```diff
@@ -1,128 +1,51 @@
...
-     * Sets a object or a callable for the object creation. A callable will be invoked
-     * every time get is called.
+     * Sets a object to be used as a service
      *
-     * @param string          $id
-     * @param object|callable $value
+     * @param string $id
+     * @param object $value
      *
      * @throws \InvalidArgumentException if service is not an object or callable
```
stof: it should not allow callables anymore, as definitions are handled by a separate method

src/PhpSpec/Console/Application.php
ciaranmcnulty: Yep, good catch - I'm considering removing this BTW, it's currently only used for testing? It should be `@internal` anyhow

ciaranmcnulty:
I've noticed a performance issue now due to too much `call_user_func` so I'll simplify the container implementation later.

src/PhpSpec/ServiceContainer/IndexedServiceContainer.php
```diff
@@ -0,0 +1,238 @@
...
+     *
+     * @param string $id
+     * @param mixed  $value
+     * @param array  $tags
+     */
+    private function addToIndex($id, $value, array $tags)
+    {
+        foreach ($tags as $tag) {
+            if (!isset($this->tags[$tag])) {
+                $this->tags[$tag] = [];
```
stof: this is useless, due to the way PHP handle nested writes in arrays

ciaranmcnulty: It is? Hm I thought it triggered a warning. that's good then!

stof: reading triggers a warning. but ``$this->tags[$tag][] = $id;`` does not need reading the current value

ciaranmcnulty: Cool, today I learned :-)

src/PhpSpec/ServiceContainer/IndexedServiceContainer.php
```diff
@@ -70,61 +75,42 @@ public function getParam($id, $default = null)
...
-        $this->addToIndex(
-            $id,
-            function ($container) use ($callable) {
-                static $instance;
-
-                if (null === $instance) {
-                    $instance = call_user_func($callable, $container);
-                }
+        $this->definitions[$id] = $value;
+        unset($this->services[$id]);
```
stof: should we allow overwriting a service definition once we have instantiated it ? this can only cause issues (other services may be referencing the old instance as a dependency).
For reference, Pimple 3 forbids this

src/PhpSpec/ServiceContainer/IndexedServiceContainer.php
```diff
@@ -70,61 +75,42 @@ public function getParam($id, $default = null)
...
-     * @param callable $callable
+     * @param callable $value
      * @param array    $tags
      *
      * @throws \InvalidArgumentException if service is not a callable
      */
-    public function define($id, $callable, $tags = [])
+    public function define($id, $value, $tags = [])
     {
-        if (!is_callable($callable)) {
```
stof: this check should not be removed IMO. Definitions must be callables

src/PhpSpec/ServiceContainer/IndexedServiceContainer.php
```diff
@@ -0,0 +1,238 @@
...
+    /**
+     * Adds a configurator, that can configure many services in one callable
+     *
+     * @internal
+     *
+     * @param callable $configurator
+     *
+     * @throws \InvalidArgumentException if configurator is not a callable
+     */
+    public function addConfigurator($configurator)
```
stof: As you started typehinting the definition callable, you could use a typehint here too

stof:
@ciaranmcnulty please look at https://github.com/phpspec/phpspec/pull/961#discussion_r67895606 even if Github collapses it by default due to other changes.
It makes sense to add such restriction to avoid hard-to-debug issues due to a misuse

src/PhpSpec/ServiceContainer/IndexedServiceContainer.php
ciaranmcnulty: That's a very good point. Current services *are* redefined in the behat application suite (which uses Symfony application tester). 

Specifically this happens in scenarios that do describe followed by run.

stof:
@ciaranmcnulty your PR changes the behavior of the ``set`` method. I don't know whether it is intended. The old method allows to do 2 things:

- set a service instance directly, which will be reused all the time until replaced. In the Symfony DIC, this is a shared (container-scoped in 2.x) synthetic service
- set a service definition which will rebuild the service each time. In Symfony DIC, this is a non-shared (prototyped-scoped in 2.x) service

``setShared`` was used to created shared services, which is what ``define`` does know, and is fine.

You changed most places using ``set`` to use ``define``, but this changes the behavior. Most changes are fine (it makes sense to keep it shared in most cases, as using set for many of them was just a performance optimization relying on the fact that they were used by a single shared service, but optimizing the way services are shared instead is way better), but changes should be reviewed carefully.
And we need to decide whether all services should be shared or no (knowing whether we need a replacement feature or no)

src/PhpSpec/ServiceContainer/IndexedServiceContainer.php
ciaranmcnulty: BTW this is allowed in the current version, but there's no test covering it.

ciaranmcnulty:
Yes indeed, there were three methods:

* `set` with object - object storage - now `set`
* `setShared` with callable - factory - now `define`
* `set` with callable - prototype-style factory - now removed

Why remove the third? As I went through the internal uses of `set` with a callable, I didn't find any that needed to be prototype-style, so I dropped support for it on the basis that it kept things simple.

I don't think non-shared services are, in general, a good idea* but I could add a third method (definePrototype?) that supports it if there's a good case for extensions using it? I definitely don't want to keep the overloaded `set` because of the ambiguity when you `set` an object that has an `__invoke`.

* Actually if they're stateless it's fine, but that's not the style PhpSpec is currently using.

src/PhpSpec/ServiceContainer/IndexedServiceContainer.php
stof: but are they redefined after instantiation ? If yes, this is fragile (the object graph could use the non-overwritten service). If no, this is not an issue

stof: if we do describe+run, we should probably isolate these 2 calls by recreating the application, because they would not run in the same PHP process in real usage and so they cannot share state

ciaranmcnulty: Yep could be, there'd be some work to refactor the behat suite, which I'll do if there's a clear benefit? Maybe there just needs a `reset` method on the container?

stof:
@ciaranmcnulty I'm totally fine with removing support for the prototype scope, as long as we don't need it in Phpspec. We just need to make sure we don't need it

ciaranmcnulty:
I'm 95% confident all the places it changed in this PR, `describe` was appropriate. I haven't audited extensions but presumably when they add `3.0` support they'll come and tell us and we'll need some BC-compatible way to add stuff.

PR #962, opened 2016-06-20
Reducing reliance on reflection, for performance reasons

Sam-Burns:


Sam-Burns:
On a side note, the same thing could be done in (Prophecy)[https://github.com/phpspec/prophecy], with implications for supported language versions.

PR #963, opened 2016-06-24
Updating dependencies

gnugat:
Hi there!

By running `composer outdated` on some projects with phpspec, I've realised `phpspec/php-diff` was outdated (installed version 1.0, when 1.1 was available). This PR fixes it (I've run phpspec, phpunit and behat tests and all went well).

I've also taken the opportunity to make composer configuration more consistent (using `^` instead of `~`) and update it (using `||` instead of `|`).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,15 +22,15 @@
     ],
 
     "require": {
-        "php":                      "^5.6|^7.0",
-        "phpspec/prophecy":         "~1.5",
+        "php":                      "^5.6 || ^7.0",
+        "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "~1.0",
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I doubt the other composer changes will be accepted. I've tried minor fixes before and had them rejected.

ciaranmcnulty:
Seems to be broken in apple yor?

ciaranmcnulty:
There seems to be a difference in the diff output on windows, so I can't merge this right now - it may need an upstream change in php-diff?

stof:
@ciaranmcnulty php-diff 1.1.0 uses the EOL of the platform: https://github.com/phpspec/php-diff/releases/tag/v1.1.0. If it makes tests fail on windows, it is probably because the testsuite does not normalize line endings before comparing the value

PR #964, opened 2016-06-25
Compare diff output using system line endings

ciaranmcnulty:
This test fails on windows when php-diff is upgraded (see #963)


ciaranmcnulty:
It's very frustrating not being able to test this on a windows box.

PR #967, opened 2016-06-28
Normalized end of line

gnugat:
Using `\n` should be safe to used for display purpose on any environment, however for read / comparison purpose we'll need to rely on `PHP_EOL`.

> **Note**: Tests pass on GNU/Linux. Unfortunately I don't have any Windows environment to test it so I'll rely on appveyor.

### DIFF
--- a/src/PhpSpec/CodeGenerator/Writer/TokenizedCodeWriter.php
+++ b/src/PhpSpec/CodeGenerator/Writer/TokenizedCodeWriter.php
@@ -83,9 +83,9 @@ final class TokenizedCodeWriter implements CodeWriter
      */
     private function insertStringAfterLine($target, $toInsert, $line, $leadingNewline = true)
     {
-        $lines = explode("\n", $target);
+        $lines = explode(PHP_EOL, $target);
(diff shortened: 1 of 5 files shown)


gnugat:
Looks like that wasn't the issue

src/PhpSpec/CodeGenerator/Writer/TokenizedCodeWriter.php
```diff
@@ -83,9 +83,9 @@ public function insertAfterMethod($class, $methodName, $method)
      */
     private function insertStringAfterLine($target, $toInsert, $line, $leadingNewline = true)
     {
-        $lines = explode("\n", $target);
+        $lines = explode(PHP_EOL, $target);
```
stof: if you explode on ``PHP_EOL`` it breaks things when the source code using ``\n`` EOL and you run on windows.

stof:
@gnugat the place where you should normalize line endings is when comparing the output in tests. Forcing to use PHP_EOL in the code is breaking things (most project force using ``\n`` in source code)

PR #968, opened 2016-06-30
Update dependencies (with fix for tests)

avant1:
This PR is based on #963, so I've left existing commits untouched.

Maybe it would be better to open separate PR with changes made in fb160040, merge it and then rebase #963 on top of it, I'm not sure. If it is - I'll do it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,15 +22,15 @@
     ],
 
     "require": {
-        "php":                      "^5.6|^7.0",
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
Great, thanks @avant1 and @gnugat

gnugat:
Thank you @avant1, you're amazing!

PR #970, opened 2016-07-04
Documented migration changes from 3.0@beta2

gnugat:
Here are the newly documented changes:

* phpspec extension configuration
* `PhpSpec\ServiceContainer` now being an interface
* `PhpSpec\ServiceContainer#define` method replacing `set` and `setShared`
* `PhpSpec\Extension#load` new argument

### DIFF
--- a/docs/manual/upgrading-to-phpspec-3.rst
+++ b/docs/manual/upgrading-to-phpspec-3.rst
@@ -46,7 +46,7 @@ To:
 
     extensions:
         SomeExtension:
-            some_config:foo
+            some_config: foo
 
         SomeOtherExtension: ~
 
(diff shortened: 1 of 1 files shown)


docs/manual/upgrading-to-phpspec-3.rst
```diff
@@ -30,6 +30,26 @@ Change to:
...
+        - SomeExtension
+        - SomeOtherExtension
+
+To:
+
+.. code-block:: yml
+
+    extensions:
+        SomeExtension:
+            some_config:foo
```
stof: missing space after the colon

gnugat: Thanks, it should be fixed now

PR #973, opened 2016-07-19
Improve the default specification template

everzet:
Original `it_is_initializable` was used to address one particular
problem in the PhpSpec 2.0 design - only one generator was being
executed during a single `phpspec run`. So basically, if back then
you described a new spec and then immediately wrote an example for
it, you were forced to run `phpspec run` twice - once to generate
the object and once to generate the method itself. To avoid this
problem I added `it_is_initializable` example into the
specification template. So that when you `describe` and then
immediately execute ` run`, PhpSpec generates an object for you.
That limited situations when you needed to rerun specs just
because of code generation. That's it.

@ciaranmcnulty solved the real problem in 2.1 by introducing
automated specification rerun on code generation. This made this
functionality redundant.

Redundancy is a good enough reason to change template on its own,
but I have a bigger reason - I see way too many specs with
`it_is_initializable` still in them. This particular "example" is
not a real example and was never supposed to be used for anything
but the use case described above.

This change both removes meaningless example and provides a
better starting point for PhpSpec newcomers by giving them a
better example template to start.

### DIFF
--- a/src/PhpSpec/CodeGenerator/Generator/templates/specification.template
+++ b/src/PhpSpec/CodeGenerator/Generator/templates/specification.template
@@ -8,8 +8,8 @@ use Prophecy\Argument;
 
 class %name% extends ObjectBehavior
 {
-    function it_is_initializable()
+    function it_()
     {
-        $this->shouldHaveType(%subject_class%::class);
+        // $this->new%name%Method($arg1, $arg2)->shouldReturn($result);
     }
 }


everzet:
Forgot about the integration tests. Fixing

everzet:
@ciaranmcnulty before I fix test suite (which is bigger change than I thought). 👍 or 👎 ?

everzet:
I assume you perceive my current proposal as less useful. Right?

ciaranmcnulty:
Having a commented out assertion rather than having a type check, yeah

everzet:
I see 4 options:

1. Keep type assertion
2. Provide a dummy example of example (like I did currently)
3. Empty example body
4. No starting example

I'm ok with all options except 1, which is a motivator for the change. Thoughts?

ciaranmcnulty:
I accept it's possible I'm having an emotional reaction beause we only just changed it to use `::class`

What's the reason for the starting spec? We should come at it from a workflow POV, what should happen when I `describe` then `run`?

a) Nothing, spec runs with no examples, no call to action + no code generation prompt (option 4)
b) Pending example prompt, no code generation (option 3 or your option 2)
c) Broken result, code generation prompt (option 1)

We should also consider the motivations for this skeleton example

i) Provide starting point for new users
ii) Provide starting skeleton for experienced users

If it's ii) then deleting that example of an example would get annoying

mheki:
Personally, I'm 👍  for option number 3
```
function it_()
      {      
      }
```
pending example indicates there is an action for me to do. 
This approach also saves me from deleting commented out line again and again.

everzet:
```
function it_()
{
}
```

avant1:
Does it makes sense to add a possibility to override used spec template via configuration?
Some thing like 
```yml
templates:
    specification: path/to/spec-template.tpl
    class: path/to/class-template.tpl
```
in `phpspec.yml`.

There could be a set of predefined templates (at least for specification templates) with type assertion, dummy example or example template without name and body.

It would allow developers to choose templates they need, and even introduce their own templates if it would be necessary for them. For now we should edit templates in `vendor` directory to use custom templates.

However, the default template should be chosen anyway.

ciaranmcnulty:
@avant1 this is already possible by convention by placing templates in `.tpl` in the project folder, and in the `.phpspec` folder in your home directory

avant1:
Ahh, I see, thanks!

ciaranmcnulty:
@avant1 see this repo, in fact ;-)

avant1:
Yeah, I've even discovered separate page in docs related to templates.

everzet:
@ciaranmcnulty so what's it gonna be?

jon-acker:
IMO, the example it_is_initializable() is very useful for newcomers, it makes it clear immediately how to write a spec, so even though I automatically delete this example I can see its value in the context of helping people to learn. I would leave it *and* in addition including a new function it_() {} underneath it, that way you can have the best of both worlds: and example of an example *plus* a `todo` spec.

PR #976, opened 2016-07-25
Phpspec 3 has been released

stof:


ciaranmcnulty:
This reveals the issues we have with docs being in the same repo :-/ I can merge this but it won't make it into the `stable` docs until `3.0.1`

stof:
@ciaranmcnulty isn't RTD building the master branch on each commit ?

ciaranmcnulty:
It is, currently the default is `3.0.0`, before that it was `stable` :-/

I mean that errata like this are naturally going to lag behind the release tags of the software the docs are advertised as being relevant to!

stof:
ah, only for ``latest``, not for ``stable``.

The big advantage of having the doc in the same repo is that it allows to force people to keep it in sync with code changes.
The issue was to write a message saying ``this version is in development``, which has 100% chances to be forgotten at release time.

ciaranmcnulty:
Maybe I should be stricter about docs changes in the PR?

PR #985, opened 2016-08-23
Removed ConsoleIO dependency from *most* Generators

guiwoda:
Generators could report back to the caller instead of writing directly to output, so now the `generate` method returns the string that used to be written directly.

Output used to be void, so there shouldn't be any risk of BC.
I think I've been thorough with usage lookup, but let me know if there's any other case I can cover.

The PromptingGenerators are pretty much impossible to decouple. I would consider switching out `ConsoleIO` for `IO` instead, that way those classes can be used with a simple `NullIO` object in non-console scenarios.

### DIFF
--- a/spec/PhpSpec/CodeGenerator/Generator/MethodGeneratorSpec.php
+++ b/spec/PhpSpec/CodeGenerator/Generator/MethodGeneratorSpec.php
@@ -13,9 +13,9 @@ use PhpSpec\Locator\Resource;
 
 class MethodGeneratorSpec extends ObjectBehavior
 {
-    function let(ConsoleIO $io, TemplateRenderer $tpl, Filesystem $fs, CodeWriter $codeWriter)
+    function let(TemplateRenderer $tpl, Filesystem $fs, CodeWriter $codeWriter)
     {
-        $this->beConstructedWith($io, $tpl, $fs, $codeWriter);
+        $this->beConstructedWith($tpl, $fs, $codeWriter);
     }
 
     function it_is_a_generator()
@@ -38,7 +38,7 @@ class MethodGeneratorSpec extends ObjectBehavior
         $this->getPriority()->shouldReturn(0);
     }
 
-    function it_generates_class_method_from_resource($io, $tpl, $fs, Resource $resource, CodeWriter $codeWriter)
+    function it_generates_class_method_from_resource($tpl, $fs, Resource $resource, CodeWriter $codeWriter)
     {
         $codeWithoutMethod = <<<CODE
 <?php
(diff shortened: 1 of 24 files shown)


guiwoda:
In specs where the generator test wasn't there, I also added that test case separately from the output expectation tests.

ciaranmcnulty:
Changing the dependencies here seems like a good idea, but I'm concerned that changing the return type is a break of the interface (although not a technical PHP break).

stof:
having a generator method returning a string meant to be a feedback message to the user looks weird to me. If I see a signature ``generate(...): string``, I expect the return value to be the generated string.

ciaranmcnulty:
I've been thinking about how we approach output and there are really three interfaces:

1) Generic output for printing (junit, dot formatter)
2) Fancy output for coloured printing (progress, nyan cat(!))
3) Asking questions to the user

Arguably 3 should go to stderr, and certainly there could be a different interface for each. 2) would probably remain framework-specific (i.e. coupled with Symfony\Console in our case)

stof:
@ciaranmcnulty simple echo statements are a bad idea IMO. If someone wants to write a UI running phpspec in the background (I know someone did it for Behat in the past for instance, so this is not an insane idea), being able to pass a different implementation of OutputInterface (for instance the Symfony BufferOutput) is an easy way to collect the output. If you switch to echo, it becomes much more painful (especially if part of the output uses echo and the other one uses the Symfony OutputInterface, as you would still be forced to use the OutputInterface doing echo as well, to keep the output in the right order)

ciaranmcnulty:
Longer term I'd like to move some of our stuff internally into Extensions similar to Behat (this is why I marked a lot of the container stuff internal BTW).

So phpspec without Symfony could just echo stuff, i.e. have some output interface with a naive implementation, and a limited set of formatters, maybe just dot.

Symfony Output extension could add an implementation of the interface that is friendlier and allows more complex (coloured) formatting.

stof:
@ciaranmcnulty but then, you will still have your own output layer here, not straight echo everywhere needing to display something (and the IO interface can be this layer btw).

ciaranmcnulty:
Yeah I'm talking about an interface with a default straight-echo implementation - sorry if that wasn't clear. Not `echo` hardcoded everywhere.

ciaranmcnulty:
@guiwoda Thanks for your efforts here, but for the reasons mentioned above I won't be merging this.

PRs like this one help us come up with new ideas, so please don't think of it as a wasted effort!

guiwoda:
Hey @ciaranmcnulty! I figured that, after your brainstorming, you'd close it, so don't worry about it. 
I was trying to reuse the generators outside of the framework, so if you don't mind, here's my point of view on the subject:

- A side-effect free `generate(...): string` method as @stof mentioned would be ideal. 
- IO could be handled outside the generator, and framework could decorate over each of them if knowing what failed or happened is a concern. 
- Prompting could be reworked as a try - catch - prompt and retry cycle so generators don't really care about getting feedback in the middle of the process. 

And, of course, splitting them to a standalone package would be awesome! :-) 

Thanks for all the great work you do!

PR #987, opened 2016-08-24
Disallow doubling of iterable

### LINKED ISSUE #986: Disallow doubling of `iterable`
PHP 7.1 introduces a new iterable type

Because we don't have a sane way to deal with this yet, PhpSpec should complain if a user typehints an example's parameter with this pseudotype.

```
function it_does_something_iterable(iterable $iterable) {
     // we should error here
}
```

avant1:
Closes #986

### DIFF
--- a/src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
+++ b/src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
@@ -43,7 +43,15 @@ final class StaticRejectingNamespaceResolver implements NamespaceResolver
      */
     private function guardScalarTypeHints($typeAlias)
     {
-        if (in_array($typeAlias, array('int', 'float', 'string', 'bool'))) {
+        $scalarTypes = [
+            'int',
+            'float',
+            'string',
+            'bool',
+            'iterable',
+        ];
+
+        if (in_array($typeAlias, $scalarTypes, true)) {
             throw new DisallowedScalarTypehintException("Scalar type $typeAlias cannot be resolved within a namespace");
         }
     }
--- a/src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
+++ b/src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
@@ -32,7 +32,7 @@ final class StaticRejectingNamespaceResolver implements NamespaceResolver
 
     public function resolve($typeAlias)
     {
-        $this->guardScalarTypeHints($typeAlias);
+        $this->guardNonObjectTypeHints($typeAlias);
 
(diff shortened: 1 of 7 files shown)


src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
```diff
@@ -43,7 +43,15 @@ public function resolve($typeAlias)
      */
     private function guardScalarTypeHints($typeAlias)
     {
-        if (in_array($typeAlias, array('int', 'float', 'string', 'bool'))) {
+        $scalarTypes = [
```
ciaranmcnulty: These aren't really scalar types any more, I guess? Can you rename the private method and this local variable please?

ciaranmcnulty:
It occurs to me that `callable` shouldn't be allowed either.

src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
```diff
@@ -43,7 +43,15 @@ public function resolve($typeAlias)
...
+            'int',
+            'float',
+            'string',
+            'bool',
+            'iterable',
+        ];
+
+        if (in_array($typeAlias, $scalarTypes, true)) {
             throw new DisallowedScalarTypehintException("Scalar type $typeAlias cannot be resolved within a namespace");
```
ciaranmcnulty: Rephrase instead of `scalar` here, or handle `iterable` separately

avant1:
Good catch about scalars!

> It occurs to me that callable shouldn't be allowed either.

`callable` and `array` are handled here: https://github.com/phpspec/phpspec/blob/32d0c02f0be297e2639682677adbcced5b7538a5/src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php#L146

And the initial version of typehints protection was implemented there too, but it was moved from there during implementing PHP7 support.

Probably things can be simplifed in `CollaboratorsMaintainer`, it makes sense to move all collaborators typehints checks to one place. Will do later.

ciaranmcnulty:
Yep, consolidating in one place makes sense. Perhaps scalars and callable/iterable can have different messages in the error though, something user-friendly.

avant1:
Seems like we cannot remove typehint checks from `CollaboratorsMaintainer` as it is a guard for PHP5.6 since typehint rewriter is being used only for PHP7.

> Perhaps scalars and callable/iterable can have different messages in the error though, something user-friendly.

Developers would see exception message of the following format:
```
Collaborator must be an object: argument 0 defined in spec\Path\To\ClassSpec::it_can_do
```

`DisallowedNonObjectTypehintException` is being used only internally.

ciaranmcnulty:
Thanks ;-)

PR #992, opened 2016-08-29
Handle fatal errors in SUS and Spec

ciaranmcnulty:
Rather than try and handle Errors directly, this wraps any Errors that occur in an Exception.

This allows PhpSpec to continue in the case of runtime Errors

### DIFF
--- a/spec/PhpSpec/Formatter/DotFormatterSpec.php
+++ b/spec/PhpSpec/Formatter/DotFormatterSpec.php
@@ -108,8 +108,7 @@ class DotFormatterSpec extends ObjectBehavior
         ConsoleIO $io,
         StatisticsCollector $stats,
         SpecificationNode $specification,
-        ExampleNode $example,
-        ReflectionFunctionAbstract $reflectionFunction
+        ExampleNode $example
     ) {
         $example->getLineNumber()->willReturn(37);
         $example->getTitle()->willReturn('it tests something');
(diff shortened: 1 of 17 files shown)


src/PhpSpec/Exception/ErrorException.php
```diff
@@ -0,0 +1,22 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Exception;
+
+final class ErrorException extends \Exception
+{
+    public function __construct(\Error $error)
+    {
+        parent::__construct($error->getMessage());
```
stof: you should get the original file and line too

ciaranmcnulty: Very good point! Thanks

ciaranmcnulty:
In the last commit I added support for recovering from parse errors in specs, but it's messy. Any feedback on doing it more cleanly would be welcome.

spec/PhpSpec/Formatter/DotFormatterSpec.php
```diff
@@ -111,8 +111,7 @@ function it_outputs_exceptions_for_failed_examples(
         ExampleNode $example,
         ReflectionFunctionAbstract $reflectionFunction
```
avant1: This double seems to be unused in example.

ciaranmcnulty:
Last commit is cleaner but introduces an issue - I need to debug before it could be merged

Issue #993, opened 2016-08-30
Exceptions not displayed correctly


ciaranmcnulty:
When presenting an exception type it's wrapped in <label> tags that aren't then parsed into colours, making them harder to read.

This also happens in non-decorated output which is definitely incorrect.

ciaranmcnulty:
Example:
![screen shot 2016-09-01 at 12 46 43](https://cloud.githubusercontent.com/assets/237866/18166288/31e017d2-7042-11e6-8f01-0ea257587c9e.png)

ciaranmcnulty:
The colouring is addressed by #994 but there is still the issue of the tagged version being used for the undecorated output:

```
PhpSpec/Foo                                                                                    
  11  - it is initializable
      exception [exc:Exception("exception")] has been thrown.

/  skipped: 0%  /  pending: 0%  /  passed: 99%  /  failed: 0%   /  broken: 1%   /  771 examples
131 specs
771 examples (770 passed, 1 broken)
5218ms
```

PR #994, opened 2016-09-01
Define a style for labels

ciaranmcnulty:
This partially addresses #993 

screen shot 2016-09-01 at 13 32 35

### DIFF
--- a/src/PhpSpec/Console/Formatter.php
+++ b/src/PhpSpec/Console/Formatter.php
@@ -49,6 +49,7 @@ class Formatter extends OutputFormatter
         $this->setStyle('value', new OutputFormatterStyle('yellow'));
         $this->setStyle('lineno', new OutputFormatterStyle(null, 'black'));
         $this->setStyle('code', new OutputFormatterStyle('white'));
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
The travis failure seems unrelated but needs investigation

src/PhpSpec/Console/Formatter.php
```diff
@@ -49,6 +49,7 @@ public function __construct($decorated = false, array $styles = array())
         $this->setStyle('value', new OutputFormatterStyle('yellow'));
         $this->setStyle('lineno', new OutputFormatterStyle(null, 'black'));
         $this->setStyle('code', new OutputFormatterStyle('white'));
+        $this->setStyle('label', new OutputFormatterStyle('white', null, array('bold')));
```
avant1: I'm wondering if white colors are readable on terminals with light colored background. Or we should not care about this (since there is white colored style for `code` and nobody complained about this)?

ciaranmcnulty: It's a good question, the answer is there are several things hard to read right now and it should be addressed in a different set of changes.

PR #995, opened 2016-09-01
Stop newline characters from accidentally being added when describing…

randompixel:
So one day when writing a description on my Macbook I accidentally hit Shift-Enter in the middle of the line (It's near " - what can I say?). 

I tried to fix it by carrying on....
screen shot 2016-09-01 at 21 41 43

However PHPSpec continues and you end up with something very magical and a bit broken:
screen shot 2016-09-01 at 21 35 10

This patch is a simple one that will strip the OS default newline character out of the name. There should never be an instance when you'd want a new line in the middle of a describe string. This gives you:
screen shot 2016-09-01 at 21 34 40

I did have a look around the tests to see if there was something to update but couldn't find anything. If you have any suggestions I'm happy to update this PR.

### DIFF
--- a/src/PhpSpec/Locator/PSR0/PSR0Locator.php
+++ b/src/PhpSpec/Locator/PSR0/PSR0Locator.php
@@ -357,7 +357,7 @@ class PSR0Locator implements ResourceLocator
      */
     private function validatePsr0Classname($classname)
     {
-        $pattern = '/^([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*[\/\\\\]?)*[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/';
+        $pattern = '/\A([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*[\/\\\\]?)*[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*\z/';
 
         if (!preg_match($pattern, $classname)) {
             throw new InvalidArgumentException(
--- a/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
+++ b/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
@@ -503,6 +503,18 @@ class PSR0LocatorSpec extends ObjectBehavior
         $this->shouldThrow($exception)->duringCreateResource('Namespace/');
     }
 
+    function it_throws_an_exception_on_PSR0_resource_with_line_breaks_at_end(Filesystem $fs)
+    {
+        $this->beConstructedWith($fs, '', 'spec', $this->srcPath, $this->specPath);
+
+        $exception = new \InvalidArgumentException(
+            'String "Namespace\Classname'.PHP_EOL.'" is not a valid class name.'.PHP_EOL.
+            'Please see reference document: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-0.md'
+        );
+
+        $this->shouldThrow($exception)->duringCreateResource('Namespace\Classname'.PHP_EOL);
+    }
+
     function it_throws_an_exception_on_PSR4_prefix_not_matching_namespace(Filesystem $fs)
     {
         $exception = new \InvalidArgumentException(


ciaranmcnulty:
This is definitely a bug, but I'd rather there was an error than a silent correction.

We currently check the class name is valid in PSR-0: https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Locator/PSR0/PSR0Locator.php#L360 so it'd rather catch here. It looks like the regex isn't multiline.

Would you be able to add the checks here perhaps?

randompixel:
I've updated the PR based on your suggestions. The regex seemed like the most sensible place to put the change (although _what_ a regex!). I also added a spec to cover the use-case.

I've tested on my Mac with a number of combinations of namespaces and line-breaks. Is this worth testing on a Windows machine too?

screen shot 2016-09-13 at 21 34 43

src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -357,7 +357,7 @@ private function createResourceFromSpecFile($path)
      */
     private function validatePsr0Classname($classname)
     {
-        $pattern = '/^([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*[\/\\\\]?)*[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/';
+        $pattern = '/^[^\r\n]([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*[\/\\\\]?)*[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/';
```
ciaranmcnulty: This isn't correct, you're allowing any char that isn't `\r` or `\n` to be the first character of the class name, including a bunch of invalid ones. I\'m not really clear why this makes your test pass, to be honest.

randompixel: OK. Happy to rework.

randompixel:
After rebase and a bit more investigation it seems
screen shot 2016-09-17 at 21 15 27
was being blocked successfully but
screen shot 2016-09-17 at 21 13 52
was being let through. This updated regexp I hope should fix that issue.
screen shot 2016-09-17 at 21 11 24

src/PhpSpec/Locator/PSR0/PSR0Locator.php
ciaranmcnulty: Ah ok it might be that `preg_match` is not in multiline mode so `^` and `$` are matching 'start of line' and 'end of line', that means extra linebreaks are allowed before and after.

You could try using `\A` and `\z` instead (note not `\Z` as that would allow a linebreak at the end), or try the `/m` or `/D` switches at the end of the pattern to affect the default behaviour of `^` and `$`.

randompixel: Yup. I went with \z and updated the spec to better capture that.

src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -357,7 +357,7 @@ private function createResourceFromSpecFile($path)
      */
     private function validatePsr0Classname($classname)
     {
-        $pattern = '/^([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*[\/\\\\]?)*[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/';
+        $pattern = '/^([a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*[\/\\\\]?)*[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*\z$/';
```
ciaranmcnulty: You don't need the dollar at the end if you've added `\z`. You can replace the `^` at the start with `\A` too

randompixel: Updated as requested :)

ciaranmcnulty:
Great, thanks for this!

ciaranmcnulty:
Ah as this is a bug fix I should have merged this to the `2.5` branch - will fix

randompixel:
Thanks for the mentoring too 🙂

PR #997, opened 2016-09-06
Support non-standard vendor and bin dirs

ciaranmcnulty:
Current binary searches well-known paths for the composer autoloader - unfortunately these are configurable by users.

It breaks in the case of the user configuring `vendor-dir` or a deep `bin-dir`

This change introduces a post-update hook to composer that modifies the path in the binary file to point to the actual autoloader location.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ behat.yml
 vendor
 composer.lock
 phpspec.phar
+bin/*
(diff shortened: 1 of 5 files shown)


ciaranmcnulty:
In the back of my head I suspect this might be a TERRIBLE idea so feedback welcome

Issue #998, opened 2016-09-07
[phpspec 3.1 beta] Method not implemented fatal errors not caught


mheki:
I pretty much like new phpspec feature resulting with fatal errors marked as broken examples. I've found one issue - for some reason, method not implemented fatal errors are still red.

How to replicate?

```
<?php

interface Foo
{
    public function foo();
}
```

```
<?php

class Bar implements Foo
{
}
```

<img width="971" alt="screen shot 2016-09-07 at 08 00 23" src="https://cloud.githubusercontent.com/assets/2435655/18302735/445b4e66-74d1-11e6-8908-391d85b93cad.png">

ciaranmcnulty:
Thanks @mheki - we currently catch parse errors when including the class and runtime errors when running the methods. I guess this is another case that happens somewhere inbetween!

ciaranmcnulty:
It seems this type of error is not catchable https://gist.github.com/ciaranmcnulty/b1b8379fbafbdd76a50381a7382944ac

jameshalsall:
We could possibly get around this by using reflection to ensure that the subject implements all of the methods in the interface and, if not, then offer to generate those stubs. If this sounds like something you want to pursue then I don't mind having a run at it, just let me know.

ciaranmcnulty:
If you want to prototype it go for it. It's possible reflection will crap out first though.

jameshalsall:
Yeah you're right, it's obviously just going to try and autoload the class by name using reflection. So the only other way would be manually parsing the content of the class file, which would be horrendous.

ciaranmcnulty:
Possibly we could use https://github.com/Roave/BetterReflection but TBH there's probably a whole class of errors that aren't catchable.

Will have to improve the experience when an uncatchable error happens.

jameshalsall:
That looks promising, I will take a look at that.

Currently if I have an interface `Foo`, and a class `Bar` that does not yet implement that interface, and my spec looks like this...

```
<?php

namespace spec;

use Bar;
use Foo;
use PhpSpec\ObjectBehavior;

class BarSpec extends ObjectBehavior
{
    function it_is_initializable()
    {
        $this->shouldHaveType(Bar::class);
    }

    function it_is_a_foo()
    {
        $this->shouldHaveType(Foo::class);
    }
}
```

[...]

PR #1000, opened 2016-09-12
Interface method generation

jameshalsall:
This is an implementation for the discussion here: https://github.com/phpspec/phpspec/issues/998#issuecomment-246125287. It aims to cover the following scenario:
1. Interface `Foo` already exists
2. We create a new concrete type called `Bar`
3. PhpSpec generates `Bar` for us
4. We update our `BarSpec` so that it specifies it should be a `Foo` type
5. PhpSpec then adds the `implements` based on whether `Foo` is an interface, and any method stubs.

TODO:
- [x] Create new generator for interfaces
- [x] Add specs to cover edge cases (e.g. classes with existing interfaces, interfaces from other namespaces)
- [x] Add more scenarios
- [x] Add scenario to ensure that answering "N" to code generation does not generate interface methods
- [x] Refactor `ClassFileAnalyser` methods (`findEndOfImplementsDeclaration` and `getClassNamespace`)
- [x] Refactor `InvalidTypeListener::afterSuite()` method to improve grade
- [x] Add support for type hinted arguments in interface methods
- [x] Update docs
- [ ] Fix method return type hint generation


jameshalsall:
I've written a basic implementation of this, it's not all encompassing and there are edge cases I need to write specs for, but there is enough there to get some feedback on.

Example, code before:

```php
<?php

namespace PhpSpec;

final class Person
{
}
```

The spec:

```php
<?php

namespace spec\PhpSpec;

use PhpSpec\Person;
use PhpSpec\ObjectBehavior;
use PhpSpec\SaysHello;

class PersonSpec extends ObjectBehavior
{
    {
        $this->shouldHaveType(Person::class);
    }

    function it_has_type()
    {
        $this->shouldHaveType(SaysHello::class);
    }
}
```

The interface `SaysHello`:

```php
<?php

namespace PhpSpec;

interface SaysHello
{
    public function saysHello();
}
```

The prompt after running the suite:

prompt

And the code after...

```php
<?php

namespace PhpSpec;

final class Person implements SaysHello
{
    public function saysHello()
    {
        // TODO: write logic here
    }
}
```

Interested to here your thoughts @ciaranmcnulty

features/code_generation/developer_generates_abstract_methods.feature
```diff
@@ -0,0 +1,62 @@
+Feature: Developer generates abstract type methods
```
ciaranmcnulty: Something like 'Developer implements interface' is more clear

jameshalsall:
Yeah absolutely, this is just the happy path at the moment. I'll now start adding tests for all of the edge cases.

For the `ImplementsGeneratorSpec` I followed the `MethodGeneratorSpec` which uses a mock Filesystem object. I'll leave this as is but also add in integrations tests.

ciaranmcnulty:
I think `MethodGeneratorSpec` predates us introducing PHPUnit (and specs like that are partly what drove the introduction)

stof:
And btw, we should also have much more behat scenario covering this:

- case of extending a class
- case of being asked to extend a class while the subject already extends a class
- case of adding a new interface in a class extending a parent class
- case of adding a new interface in a class already implementing some interface

ciaranmcnulty:
I'd rather we didn't encourage inheritance

jameshalsall:
I agree with Ciaran, but I am in the process of writing scenarios for all of those Stof mentioned to cover any situations where people are using inheritance (and it's not always avoidable, eg. with Exceptions).

On Thu, 06 Oct 2016 at 09:12 Ciaran McNulty

<
mailto:Ciaran McNulty <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

I'd rather we didn't encourage inheritance

—

You are receiving this because you modified the open/close state.

Reply to this email directly,
https://github.com/phpspec/phpspec/pull/1000#issuecomment-251895483
, or

ciaranmcnulty:
In terms of limiting your effort for this increment I'd suggest covering those scenarios with 'Then I should not get promoted' steps

jameshalsall:
I think most of those scenarios will already work anyway, but I'll bear that in mind. 

On Thu, 06 Oct 2016 at 13:20 Ciaran McNulty

<
mailto:Ciaran McNulty <notifications@github.com>
> wrote:

a, pre, code, a:link, body { word-wrap: break-word !important; }

In terms of limiting your effort for this increment I'd suggest covering those scenarios with 'Then I should not get promoted' steps

—

You are receiving this because you modified the open/close state.

Reply to this email directly,
https://github.com/phpspec/phpspec/pull/1000#issuecomment-251945390
, or

jameshalsall:
I've added more scenarios now that cover the following:

- generating methods from an interface in both an empty and a non-empty class, where the class and interface are in the same namespace
- generating methods from an interface in an empty class, where the class and interface are from different namespaces
- generating methods from an interface in a class that has a parent class
- generating methods from an interface in a class that already implements another interface
- specifying an abstract class for a concrete class (no code generation should take place and the suite should fail)
- specifying an interface for a class that has a parent class already implementing that interface (no code generation should take place)

I've removed the aim of adding abstract class support in this PR as there is a lot more involved than adding interface method generation.

jameshalsall:
@ciaranmcnulty be good to get some more feedback on this - the failure in Travis is on PHP 7.1 and doesn't look related to this, unable to replicate it locally on the same PHP version.

ciaranmcnulty:
Overall code looks good - there are probably small errors a closer review will catch :-)

One thing that's missing is typehinted arguments in the interface - I suspect this is going to generate tonnes of edge cases...

jameshalsall:
Wow, I can't believe I missed type hinted arguments in my scenarios. I'll get working on that.

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,618 @@
...
+      }
+      """
+    When I run phpspec and answer "y" when asked if I want to generate the code
+    Then the class in "src/CodeGeneration/AbstractTypeMethods/Person.php" should contain:
+      """
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods;
+
+      class Person implements CanSpeak
```
roukmoute: Is it possible to add a `final` class ?

jameshalsall: I will add a `final` class in one my scenarios :)

jameshalsall:
@ciaranmcnulty - I think it would make sense for me to open another PR to bring your work here (https://github.com/ciaranmcnulty/phpspec-typehintedmethods/blob/master/src/Generator/TypeHintedMethodGenerator.php) into phpspec core.

If you're happy with that then I'll open the PR and commit with you as the author to maintain attribution.

ciaranmcnulty:
@jameshalsall If that code will help out, feel free to merge it into phpspec core.

I'm going to raise an issue to resurrect the idea of bringing that functionality back into core either via my extension, @gnugat's, or a third option

jameshalsall:
Ok, I'll open a separate PR to bring the typehinting into core because this PR is blocked now - maybe as a future consideration we can look at @gnugat's extension

ciaranmcnulty:
I mean maybe pull in the specific infrastructure you need to do the typehint generation for this PR, but not enable it for the other cases

jameshalsall:
This will mean adding a new generator specifically for interface method generation, or modifying the current one which I'm less inclined to do.

gnugat:
Maybe we should worry about typehinted arguments later? In my humble opinion it deserves a whole Pull Request on its own.

ciaranmcnulty:
@gnugat Most interfaces have type hints in their arguments, so to implement the interface methods in the class need to have matching type hints

ciaranmcnulty:
@jameshalsall what's the status of this, any idea why the failure?

jameshalsall:
I have added scenarios to cover type hinted arguments which is why the build is failing. I will pick this up again, it's been updated with latest from `master`.

jameshalsall:
I've added basic support for type hinted argument support, to do so I've added a new optional key in the `MethodGenerator`'s `$data` argument, `reflection_parameters`. This is currently populated by the `InvalidTypeListener::promptUserToImplementMethods()` method with an instance of `\ReflectionParameters` (we can always guarantee that the interface is "reflectable" (unless it doesn't exist).

For now, I haven't added code to manipulate the `use` statements because of the work involved, this PR is already quite big. However it will shorten the type hint to just the class short name if the type hinted class shares the same namespace as the class we are updating.

Let me know if there are any edge cases you think I've missed outside of what is in the feature file.

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+
+      class ManagerSpec extends ObjectBehavior
+      {
+          function it_can_speak()
+          {
+              $this->shouldHaveType(CanManage::class);
+          }
+      }
+      """
+    And the class file "src/CodeGeneration/AbstractTypeMethods/CanManage.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+
+      class PersonSpec extends ObjectBehavior
+      {
+          function it_can_speak()
+          {
+              $this->shouldHaveType(CanSpeak::class);
+          }
+      }
+      """
+    And the class file "src/CodeGeneration/AbstractTypeMethods/CanSpeak.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+
+      class PersonSpec extends ObjectBehavior
+      {
+          function it_can_speak()
+          {
+              $this->shouldHaveType(\CanSpeak::class);
+          }
+      }
+      """
+    And the class file "src/CanSpeak.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      """
+    And the class file "src/Person.php" contains:
+      """
+      <?php
+
+      class Person
+      {
+      }
+      """
+    And the class file "src/Phrase.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      class EngineerSpec extends ObjectBehavior
+      {
+          function it_can_write_code()
+          {
+              $this->shouldHaveType(CanWriteCode::class);
+          }
+      }
+     """
+
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Engineering/CanWriteCode.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+
+      class EngineerSpec extends ObjectBehavior
+      {
+          function it_can_write_code()
+          {
+              $this->shouldHaveType(CanWriteCode::class);
+          }
+      }
+     """
+
```
ciaranmcnulty: shouldn't be a gap here

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods\Engineering;
+
+      interface CanWriteCode
+      {
+          public function writeCode($code);
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      """
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods;
+
+      class Engineer
+      {
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      class UserFriendlyExceptionSpec extends ObjectBehavior
+      {
+          function it_is_a_friendly_exception()
+          {
+              $this->shouldHaveType(FriendlyMessageException::class);
+          }
+      }
+     """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Exceptions/FriendlyMessageException.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods\Exceptions;
+
+      interface FriendlyMessageException
+      {
+          public function getFriendlyMessage();
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      """
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods;
+
+      class UserFriendlyException extends \RuntimeException
+      {
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+          }
+      }
+     """
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Type.php" contains:
+      """
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods;
+
+      interface Type
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+
+      class UserActionExceptionSpec extends ObjectBehavior
+      {
+          function it_is_an_authorised_user_exception()
+          {
+              $this->shouldHaveType(AuthorisedUserException::class);
+          }
+      }
+     """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Exceptions/UserException.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      namespace CodeGeneration\AbstractTypeMethods\Auth;
+
+      interface AuthorisedUserException extends UserException
+      {
+          public function getToken();
+
+          public function setToken(Token $token);
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      """
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods;
+
+      class UserActionException
+      {
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Auth/Token.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      """
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods\Auth;
+
+      interface Token
+      {
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods\Auth;
+
+      interface UserException
+      {
+          public function getUser();
+      }
+      """
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Exceptions/AuthorisedUserException.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      <?php
+
+      namespace CodeGeneration\AbstractTypeMethods\Auth;
+
+      interface TokenAware
+      {
+          public function getToken();
+      }
+      """
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Exceptions/UsernameAware.php" contains:
```
ciaranmcnulty: this is an "interface file"

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      use CodeGeneration\AbstractTypeMethods\Auth\UsernameAware;
+
+      class User implements UsernameAware
+      {
+          public function getUsername()
+          {
+          }
+      }
+      """
+
```
ciaranmcnulty: remove gap

features/code_generation/developer_implements_interface.feature
```diff
@@ -0,0 +1,644 @@
...
+      class UserSpec extends ObjectBehavior
+      {
+          function it_is_aware_of_tokens()
+          {
+              $this->shouldHaveType(TokenAware::class);
+          }
+      }
+     """
+    And the class file "src/CodeGeneration/AbstractTypeMethods/Exceptions/TokenAware.php" contains:
```
ciaranmcnulty: this is an "interface file"

src/PhpSpec/CodeGenerator/Generator/Argument/StringBuilder.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+namespace PhpSpec\CodeGenerator\Generator\Argument;
+
+class StringBuilder
```
ciaranmcnulty: I'd rather not use the `Builder` suffix when not implementing the builder pattern. The responsibility of this class seems to be something around converting reflection parameters into an arguments string, right?

(this will potentially be useful if we re-introduce argument type hints in future)

src/PhpSpec/CodeGenerator/Generator/Argument/StringBuilder.php
```diff
@@ -0,0 +1,56 @@
...
+    public function buildFromReflectionParameters(array $parameters, string $targetClassNamespace) : string
+    {
+        return implode(', ', array_map(function (\ReflectionParameter $parameter) use ($targetClassNamespace) {
+            return $this->buildArgument($parameter, $targetClassNamespace);
+        }, $parameters));
+    }
+
+    private function buildArgument(\ReflectionParameter $parameter, string $targetClassNamespace) : string
+    {
+        $parameterName = '$' . $parameter->getName();
```
ciaranmcnulty: Even though this is trivial at the moment it might be better to extract this to a method, to make the level of abstraction in this method consistent

src/PhpSpec/CodeGenerator/Generator/Argument/StringBuilder.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $type = $parameter->getType();
+
+        switch (true) {
```
ciaranmcnulty: This type logic can be extracted to a method, and possibly replaced with if/else

src/PhpSpec/CodeGenerator/Generator/Argument/StringBuilder.php
```diff
@@ -0,0 +1,56 @@
...
+            case (class_exists($type) || interface_exists($type)):
```
ciaranmcnulty: Why does the class have to exist before we resolve its type? I'd expect this logic to also apply to not existing classes too?

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
```diff
@@ -45,17 +46,29 @@
...
+        TemplateRenderer $templates,
+        Filesystem $filesystem,
+        CodeWriter $codeWriter,
+        StringBuilder $argumentBuilder
+    ) {
+        $this->io = $io;
+        $this->templates = $templates;
         $this->filesystem = $filesystem;
         $this->codeWriter = $codeWriter;
+        $this->argumentBuilder = $argumentBuilder;
```
ciaranmcnulty: The fact that you've called it `argumentBuilder` makes me think `StringBuilder` is too generic (see other comment about using `*Builder`) - reconsider this class's name

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
```diff
@@ -65,18 +78,21 @@ public function supports(Resource $resource, string $generation, array $data) :
...
-        $arguments = $data['arguments'];
-
-        $argString = \count($arguments)
-            ? '$argument'.implode(', $argument', range(1, \count($arguments)))
-            : ''
-        ;
+        $filepath = $resource->getSrcFilename();
+        $name = $data['name'];
+
+        if (isset($data['reflection_parameters'])) {
```
ciaranmcnulty: I'm not happy with this conditional - calling the method with 'reflection arguments' instead, when we want type hints, is not an obvious API.

Perhaps we should have an `arguments` array in all cases, and a separate optional `argument_types`? This would require the caller doing the reflection before calling the generator

It makes sense to me that the caller 'knows' the types come from analysing a method, but the generator is just using a list it's passed - will clean up the generator code / make tests less reliant on example classes

ciaranmcnulty: Or possibly we should have an `Argument` value object - might be overkill

src/PhpSpec/CodeGenerator/Writer/TokenizedCodeWriter.php
```diff
@@ -58,6 +58,53 @@ public function insertAfterMethod(string $class, string $methodName, string $met
         return $this->insertStringAfterLine($class, $method, $line);
     }
 
+    public function insertImplementsInClass(string $class, string $interface) : string
```
ciaranmcnulty: This needs cleaning up - the logic is hard to follow - I'd suggest extracting some private methods

src/PhpSpec/CodeGenerator/Writer/TokenizedCodeWriter.php
```diff
@@ -58,6 +58,53 @@ public function insertAfterMethod(string $class, string $methodName, string $met
...
+    {
+        $classLines = explode("\n", $class);
+        $interfaceNamespace = '';
+
+        if (false !== strpos($interface, '\\')) {
+            $parts = explode('\\', $interface);
```
ciaranmcnulty: this explode/join might be better with string functions (regex)

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -301,7 +308,8 @@ private function setupGenerators(IndexedServiceContainer $container)
                 $c->get('console.io'),
                 $c->get('code_generator.templates'),
                 $c->get('util.filesystem'),
-                $c->get('code_generator.writers.tokenized')
+                $c->get('code_generator.writers.tokenized'),
+                new CodeGenerator\Generator\Argument\StringBuilder()
```
ciaranmcnulty: define this as a service so we can reuse it elsewhere (and so extensions can decorate it)

src/PhpSpec/CodeGenerator/Generator/Argument/StringBuilder.php
jameshalsall: The interface method wouldn't be type-hinting for a non-existent class or interface at the stage we're speccing the implementor, so we should check it exists which means we can also use reflection to get its argument list

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
jameshalsall: I'm not happy with it either - I mentioned in a comment further up that it needs a refactor as part of bringing in type hints for all method generation. I have tried to keep all of type hinted method signature generator out of this class and in `PhpSpec\CodeGenerator\Generator\Argument\StringBuilder`, so it can be re-used easily after refactoring.

It's not as straightforward as just adding a new `argument_types` key, as we would need to map nullable and default values along with the types - also I've tried to avoid the `$argument1, $argument2 ...` style of naming with this generation as we know the interface method argument names.

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
jameshalsall: I've called it `argumentBuilder` to reflect it's a builder for arguments, (`PhpSpec\CodeGenerator\Generator\Argument\StringBuilder` makes more sense when you read the FQCN) - I'll change the builder suffix.

src/PhpSpec/CodeGenerator/Generator/Argument/StringBuilder.php
jameshalsall: I usually use `switch` for > 2 cases, just personal preference - I will change it to an if/else and extract it

src/PhpSpec/CodeGenerator/Generator/MethodGenerator.php
ciaranmcnulty: It sounds then like `arguments` should contain value objects that capture this complexity (and can have `__toString` to support BC)

src/PhpSpec/CodeGenerator/Generator/Argument/StringBuilder.php
ciaranmcnulty: It's very possible to have type hints for types that aren't defined _in the current project yet_

ciaranmcnulty: i it's completely impossible then you don't need the other case anyhow ;-0

ciaranmcnulty:
@jameshalsall are you able to fix the test failure?

jameshalsall:
There are some things still left to finish. 
My MacBook has died so I’m unable to work on this right now, although I’ll be trying to repair it tonight. If it works then I’ll finish this PR.

jameshalsall:
I'm working on the argument modelling currently so that we don't have to rely on the check for the `reflection_parameters` key in `MethodGenerator`

jameshalsall:
I've updated this now so that there is an `Argument` value object, that can be created from `ReflectionParam` instances. The `InvalidTypeListener` now sets an array of `Argument` instances in the `argument` key, just like the `MethodNotFoundListener`  does, except the latter generates numbered arguments from call time values (see https://github.com/phpspec/phpspec/pull/1000/files#diff-8637c505497ce19a95ac11e456ada6b0R106). This has allowed me to clean up the implementation of `MethodGenerator` slightly by having the method argument string generation done entirely by `PhpSpec\CodeGenerator\Generator\Argument\StringConverter` and should make it easier down the line to add true type hinted method support / nullable types support.

ciaranmcnulty:
This looks good to me, @jameshalsall - can you update the docs?

docs/manual/getting-started.rst
```diff
@@ -189,6 +189,109 @@ In the example here you specified the value the ``toHtml`` method should
...
+
+If we take a look at our ``Knight`` class should contain method stubs for the 2 methods in the ``GamePiece`` interface.
+
+.. code-block:: php
+
+    <?php
+
+    class Knight implements GamePiece
+    {
+        public function getName(): string
```
jameshalsall: writing the docs I found an issue with the method stub generation, it doesn't add the return type hint on method stubs if the interface has them, leaving out the return type hint doesn't break anything but it is still incomplete - I can add this to the PR or do it later after it's merged, up to you @ciaranmcnulty

ciaranmcnulty: Makes sense to fix in this PR, if you can

MarcelloDuarte:
I just had a thought. Have you considered generating examples for the interface methods in the `PersonSpec` instead:

```php
<?php

namespace spec\PhpSpec;

use PhpSpec\Person;
use PhpSpec\ObjectBehavior;
use PhpSpec\SaysHello;

class PersonSpec extends ObjectBehavior
{
    {
        $this->shouldHaveType(Person::class);
    }

    function it_has_type()
    {
        $this->shouldHaveType(SaysHello::class);
    }

    function it_implements_saysHello()
    {
        $this->saysHello()->shouldReturn(NULL); // <- not sure whether to have a matcher or not
    }
}
```

I just think that generating the implementation without a test is less aligned with the goals of the tool.

jameshalsall:
I'm not sure about stubbing interface methods just for their return type given that PHP's return types will do this for us anyway. 

I think having examples for each method in the interface offers little to no value.

MarcelloDuarte:
That's not the case though. The spec is not for describing the `SayHello` interface, which I agree would deliver no value. But it is a `Person` spec, the example should describe the method in this particular concrete implementation.

ciaranmcnulty:
@MarcelloDuarte I see your point but I'm not sure I agree:

**1. Testing the type system**

We won't be able to generate specs with meaningful behaviour, we'll only be able to generate things like:

```php
function it_takes_foo_and_returns_bar(Foo $foo)
{
 $this->baz($foo)->shouldHaveType(Bar::class)`
}
```

This is kind of only making assertions about types - the interface contains that already:

```php
public baz(Foo $foo) : Bar
```

So for me in this case, asserting that the class implements the interface is sufficient; we will write specs when we have some actual behaviour to describe aside from the type system

**2. Technical difficulty**

Currently we don't generate type hints for method arguments or return types, therefore

```php
 $this->baz($foo)->shouldHaveType(Bar::class)`
```

Will lead to generation of:

```php
public baz($foo) { /* Todo message */ }
```

Which will fail to fulfil the interface's contract and cause an error

(I want to revisit type hint generation at some point BTW)

jameshalsall:
This is still on my list :)

ciaranmcnulty:
@jameshalsall good to know!

ciaranmcnulty:
I'm going to close this, please reopen if you get time to work on it

PR #1006, opened 2016-09-26
Allow \Error instances in ThrowMatcher

jameshalsall:
Helps to fix https://github.com/phpspec/prophecy/issues/281.

`ThrowMatcher` was checking for a class `\Throwable` instead of an interface.

I've opened this against 2.5 as it is a bug that should probably be backported (prophecy is still using PHPSpec 2.x).

The build will fail until https://github.com/phpspec/prophecy/pull/288 is merged into a stable release.

### DIFF
--- a/spec/PhpSpec/Matcher/ThrowMatcherSpec.php
+++ b/spec/PhpSpec/Matcher/ThrowMatcherSpec.php
@@ -42,6 +42,10 @@ class ThrowMatcherSpec extends ObjectBehavior
         $arr->ksort()->willThrow('\Error');
 
         $this->positiveMatch('throw', $arr, array('\Error'))->during('ksort', array());
+
+        $arr->ksort()->willThrow(new \Error());
+
+        $this->positiveMatch('throw', $arr, array('\Error'))->during('ksort', array());
     }
 
     function it_accepts_a_method_during_which_an_exception_should_not_be_thrown(ArrayObject $arr)
--- a/spec/PhpSpec/Matcher/ThrowMatcherSpec.php
+++ b/spec/PhpSpec/Matcher/ThrowMatcherSpec.php
@@ -33,7 +33,7 @@ class ThrowMatcherSpec extends ObjectBehavior
         $this->positiveMatch('throw', $arr, array('\Exception'))->during('ksort', array());
     }
 
-    function it_accepts_a_method_during_which_an_error_should_be_thrown(ArrayObject $arr)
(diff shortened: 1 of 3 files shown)


spec/PhpSpec/Matcher/ThrowMatcherSpec.php
```diff
@@ -42,6 +42,10 @@ function it_accepts_a_method_during_which_an_error_should_be_thrown(ArrayObject
         $arr->ksort()->willThrow('\Error');
 
         $this->positiveMatch('throw', $arr, array('\Error'))->during('ksort', array());
+
+        $arr->ksort()->willThrow(new \Error());
```
ciaranmcnulty: This is a second case so should be a separate example

src/PhpSpec/Matcher/ThrowMatcher.php
```diff
@@ -284,7 +284,7 @@ private function getException(array $arguments)
         }
 
         if (is_object($arguments[0])) {
-            if (class_exists('\Throwable') && $arguments[0] instanceof \Throwable) {
+            if (interface_exists('\Throwable') && $arguments[0] instanceof \Throwable) {
```
jameshalsall: Removed.

spec/PhpSpec/Matcher/ThrowMatcherSpec.php
jameshalsall: I've split that up now.

spec/PhpSpec/Matcher/ThrowMatcherSpec.php
```diff
@@ -44,6 +44,17 @@ function it_accepts_a_method_during_which_an_error_should_be_thrown(ArrayObject
         $this->positiveMatch('throw', $arr, array('\Error'))->during('ksort', array());
     }
 
+    function it_accepts_a_method_during_which_an_error_specified_by_instance_should_be_thrown(ArrayObject $arr)
+    {
+        if (!class_exists('\Error')) {
+            throw new SkippingException('The class Error, introduced in PHP 7, does not exist');
+        }
+
+        $arr->ksort()->willThrow(new \Error());
```
ciaranmcnulty: See if this works...

`$arr->ksort()->will(function(){ throw new \Error(); });`

This might keep us with a passing build until the corresponding Prophecy changes can be released

jameshalsall: That works... some satanic black magic.

ciaranmcnulty:
Thanks for this! Good catch

PR #1010, opened 2016-10-11
Remove the make dependency

unfunco:
This commit removes the make dependency, since it seemed a roundabout
way of building the phar file, since the box.json finder ignore array
is also case-sensitive, I have added additional ignores from the
vendor directory, this reduces the size of the phar by about 20%.


ciaranmcnulty:
Probably a good idea - Make was only really useful for building locally.

.travis.yml
```diff
@@ -36,10 +36,10 @@ script:
 
 before_deploy:
   - curl -LSs https://box-project.github.io/box2/installer.php | php
-  - chmod +x box.phar
```
unfunco: I think the installer does it automatically when piped to `php`, when I downloaded it and ran through the steps yesterday, `box.phar` was already executable. I could add it back in just incase box-project changes that in the future though?

unfunco:
```
🍍  curl -LSs https://box-project.github.io/box2/installer.php | grep chmod
    @chmod($item->name, 0755);
```

ciaranmcnulty: Ah! Good stuff

ciaranmcnulty:
@unfunco can you rebase this against 2.5 please?

ciaranmcnulty:
Good stuff, thanks!

PR #1011, opened 2016-10-11
Don't fail the build when building against pre-releases of PHP

unfunco:
I'm not sure how I feel about this yet, my other open PR is failing
because RC2 is failing the integration tests, that is also the cause
of the failing test on #1000 which @jameshalsall is working on.

I think it makes sense to test against the nightly builds since it
reveals issues earlier (for example, the RFC for disallowing null
parameters into `get_class` that passed 15/3 on October 9th has
already revealed some changes that are required in PhpSpec, I
have configured Travis to not panic too much about failures in
nightly builds because it is also likely to cause some false
positives.) – I think versions should only be added once they're
ready for production.

If this is merged, the other two open PRs should start passing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,15 +13,15 @@ matrix:
     - php: 5.6
     - php: hhvm
     - php: 7.0
-    - php: 7.1
-    - php: 7.1
+    - php: nightly
+    - php: nightly
       env: DEPENDENCIES='dev'
   allow_failures:
-    - env: DEPENDENCIES='dev'
+    - php: nightly
   fast_finish: true
 
 before_install:
-  - if [ ${TRAVIS_PHP_VERSION:0:4} != "hhvm" ] && [ ${TRAVIS_PHP_VERSION:0:3} != "7.1" ]; then phpenv config-rm xdebug.ini; fi
+  - if [ ${TRAVIS_PHP_VERSION:0:4} != "hhvm" ] && [ ${TRAVIS_PHP_VERSION:0:7} != "nightly" ]; then phpenv config-rm xdebug.ini; fi
 
 install:
   - export COMPOSER_ROOT_VERSION=dev-master
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,15 +13,16 @@ matrix:
     - php: 5.6
     - php: hhvm
     - php: 7.0
-    - php: nightly
-    - php: nightly
       env: DEPENDENCIES='dev'
+    - php: 7.1
+    - php: nightly
   allow_failures:
+    - php: 7.1
     - php: nightly
   fast_finish: true
 
 before_install:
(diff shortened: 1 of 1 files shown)


unfunco:
This is accepted for 7.2 and the allowed failures in the change to the travis configuration are from the merge of this RFC: https://wiki.php.net/rfc/get_class_disallow_null_parameter – the nightly build doesn't fail the integration tests like RC2 of 7.1 is doing at the moment though.

GrahamCampbell:
👎

unfunco:
Anything constructive to add other than an emoji?

My reasoning for the above changes are:

* PHP 7.1 isn't production ready, and likely won't be for another month or two, the release candidates currently used to build and continuously integrate PhpSpec are segfaulting, causing failed builds.
* Using nightly builds keeps the project closer to the edge of the latest developments to the language itself, there will perhaps be some additional failures testing against nightlies, but that's why I think it's perfectly reasonable to allow those failures to happen and not obstruct the continued development of the project.

Feel free to improve upon the PR.

GrahamCampbell:
7.1 is fine.

GrahamCampbell:
> Using nightly builds keeps the project closer to the edge of the latest developments

Not really significant. A new RC comes out every two weeks.

.travis.yml
```diff
@@ -13,15 +13,15 @@ matrix:
     - php: 5.6
     - php: hhvm
     - php: 7.0
-    - php: 7.1
-    - php: 7.1
+    - php: nightly
+    - php: nightly
       env: DEPENDENCIES='dev'
```
stof: dev deps should be tested on a stable PHP version IMO, to avoid having this job failing due to PHP issues

ciaranmcnulty:
Removing 7.1 from the allow-failure section was possibly an oversight on my part - I had assumed that during the RC phase it would be fairly stable. Lesson learned!

There's still value in testing against 7.1RCs as well as nightlies, but probably both should be allowed to fail.

I agree with @stof - we should test `DEPENDENCIES='dev' `agains stable

Also... should we be tagging against php `~7.0.0` and add 7.1 support when we're sure it works? Otherwise we're doomed to repeat this dance again when 7.2 is released (due to the way composer works this won't protect our users from broken versions until we bump to 4.0.0)

unfunco:
@stof @ciaranmcnulty So should I move `env: DEPENDENCIES='dev'` to `7.0`, add back in 7.1 as well as nightlies, and allow both 7.1 and nightlies to fail without failing the entire build?

(Personally, I think this is reasonable and would avoid holdups with other outstanding PRs)

ciaranmcnulty:
That sounds good to me

jameshalsall:
@ciaranmcnulty can this be merged?

ciaranmcnulty:
Thanks @unfunco

PR #1012, opened 2016-10-11
Compatibility for PHP 7.2 get_class() changes

mbeccati:
The RFC https://wiki.php.net/rfc/get_class_disallow_null_parameter has been implemented yesterday and phpSpec isn't compatible with it. The PR fixes the issue, while retaining backwards compatibility.

### DIFF
--- a/src/PhpSpec/Listener/MethodNotFoundListener.php
+++ b/src/PhpSpec/Listener/MethodNotFoundListener.php
@@ -70,7 +70,8 @@ final class MethodNotFoundListener implements EventSubscriberInterface
             return;
         }
 
-        $classname = get_class($exception->getSubject());
+        $subject = $exception->getSubject();
+        $classname = null === $subject ? get_class() : get_class($subject);
         $methodName = $exception->getMethodName();
         $this->methods[$classname .'::'.$methodName] = $exception->getArguments();
         $this->checkIfMethodNameAllowed($methodName);
--- a/src/PhpSpec/Listener/MethodReturnedNullListener.php
+++ b/src/PhpSpec/Listener/MethodReturnedNullListener.php
@@ -110,7 +110,8 @@ final class MethodReturnedNullListener implements EventSubscriberInterface
             return;
         }
 
-        $class = get_class($this->lastMethodCallEvent->getSubject());
+        $subject = $this->lastMethodCallEvent->getSubject();
+        $class = null === $subject ? get_class() : get_class($subject);
         $method = $this->lastMethodCallEvent->getMethod();
 
         if (!$this->methodAnalyser->methodIsEmpty($class, $method)) {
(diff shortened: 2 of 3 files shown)


mbeccati:
Not sure why the travis build failed for 7.1, as it passes locally.

mbeccati:
Not sure why the travis build failed for 7.1, as it passes locally.

stloyd:
@mbeccati It looks like `./vendor/bin/phpunit --testdox` fails on PHP 7.1

jameshalsall:
That RFC is for 7.2 - not 7.1.

mbeccati:
@jameshalsall indeed. I shouldn't be doing PRs this early in the morning... I've seen the [failure with latest 7.1](https://revive.beccati.com/bamboo/browse/PHP-PHPSPEC-861) and I rushed the fix. I've sent an email to the committer to get the change reverted form 7.1, but in any case the fix will guarantee forward compatibility ;)

src/PhpSpec/Listener/MethodNotFoundListener.php
```diff
@@ -70,7 +70,8 @@ public function afterExample(ExampleEvent $event)
             return;
         }
 
-        $classname = get_class($exception->getSubject());
+        $subject = $exception->getSubject();
+        $classname = null === $subject ? get_class() : get_class($subject);
```
stof: getting the class of the MethodNotFoundListener does not make any sense here. We should rather handle the case of not having the subject in a different way

mbeccati: @stof sure, although it's unrelated to the actual get_class fix ;)

ciaranmcnulty: Yes it seems odd that we use the current class - in what case is that even unset?

Sam-Burns:
Am I right in thinking that it's only when our own specs run that `$exception->getSubject()` returns `null`?

src/PhpSpec/Listener/MethodNotFoundListener.php
MarcelloDuarte: No use case. It should never happen as the method is guarded to only exceptions of method not found, which means subject is an object.

mbeccati:
@Sam-Burns I believe so.

Sam-Burns:
@mbeccati Maybe another solution could be to make `$exception->getSubject()` return something in the tests failing in [the PHP nightly build](https://travis-ci.org/phpspec/phpspec/jobs/208612691)?

mbeccati:
@Sam-Burns Possibly a better idea, yes. The point of my PR was just to quickly adapt the existing behaviour, which however was buggy, it seems ;)

My suggestion would be to close the PR and fix the real cause of the issue.

stof:
Closing in favor of #1106 which fixed the cause of the issue rather than making workarounds

PR #1014, opened 2016-10-12
Add basic support for array objects in ArrayContainMatcher

jameshalsall:
Fixes issue raised in https://github.com/phpspec/prophecy/issues/206.

This allows the `ArrayContainMatcher` to work with `ArrayObject` instances as well as arrays, by basically using an array copy.

There may also be an argument for supporting `SplFixedArray`

EDIT: If this is of value then I will add the same support to the other Array matchers

### DIFF
--- a/spec/PhpSpec/Matcher/ArrayContainMatcherSpec.php
+++ b/spec/PhpSpec/Matcher/ArrayContainMatcherSpec.php
@@ -2,6 +2,7 @@
 
 namespace spec\PhpSpec\Matcher;
 
+use PhpSpec\Exception\Example\FailureException;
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
This is quite specific to `ArrayObject`, but presumably would be useful for any other `Traversable` objects.  Similarly `shouldHaveKey` ought to support `Traversable` and `ArrayAccess`.

I'd rather we support these more generic cases than do something specific to `ArrayObject`, so don't want to merge this on that basis.

On another note, rather than adding cases to existing matchers it'll be better to add additional matchers (it's fine for matches to support the same method invocation but act on different types)

PR #1016, opened 2016-10-17
Implement matchers for iterables

### LINKED ISSUE #379: Add support for php 5.5 generators
At the moment, it is not possible to spec php 5.5 generators:
http://www.php.net/manual/en/language.generators.php

Here is a simplified sample spec that I cannot get working:

**_The subject**_

``` php
namespace Sample;

final class Generator
{

    public function generate()
    {
        foreach(range(1,3) as $index => $value) {
            yield $index => $value;
        }
    }

}
```

**_The Spec**_

``` php
namespace spec\Sample;

use PhpSpec\ObjectBehavior;
use Prophecy\Argument;

final class GeneratorSpec extends ObjectBehavior
{
    public function it_should_loop_over_generator_values_attempt1()
    {
        $generated = $this->generate();
        $generated->shouldBeAnInstanceOf('Generator');
        foreach ($generated as $index => $value) {
            die('not getting here');
        }
    }

    public function it_should_loop_over_generated_values_attempt2()
    {
        foreach ($this->generate() as $index => $value) {
            die('not getting here');
        }

    }

    public function it_should_loop_over_generated_values_attempt3()
    {
        foreach ($this->generate()->getWrappedObject() as $index => $value) {
           // This works but $index and $value are just integers, so nothing can be tested in a phpspec way. 
        }

    }   

    public function it_should_loop_over_generated_values_attempt4()
    {
        $generated = $this->generate();
        $next = $generated->next();
        $next->shouldBe(1);
    }
}

```

### LINKED ISSUE #954: [RFC] Support for iterators (incl. generators)
Hello!

Due to lack of generators support (#379), I wrote a [small extension](https://github.com/pamil/phpspec-generator-extension). However, it would be awesome to have it included in the core (looking forward to 3.0 release) :)

Iterators and generators are quite different than arrays:
 - they can be infinite
 - they can return keys that aren't just string or integer
 - keys do not have to be unique
 - generators cannot be rewinded if they were run

My use-cases have brought up the following matchers:
 - `shouldGenerateKeys(...$keys)` - I was generating a map `object => array` and only cared, if objects are in the correct order
 - `shouldGenerateValues(...$values)` - it's like `shouldReturn(['foo', 'bar'])`, but also works for infinite iterator
 - `shouldGenerate` - to assert both keys and values, two APIs:
    - `shouldGenerate(['key' => 'value1'], ['key' => 'value2'])` - for string/integer keys
    - `shouldGenerate([$object, 'value1'], [$anotherObject, 'value2'])` - for more complicated keys

The following generators assume that we want to check only the first elements, ie. `shouldGenerateValues('foo')` matches `yield 'foo'; yield 'bar';`.

Is there a need to check if given iterator has no more elements than it is matched against? `shouldExactlyGenerate`?

Is there a need to check if given iterator has no elements? `shouldGenerate*()` with no arguments?

That's all, I hope :tada: Let me know what can be improved.

pamil:
*Fixes #379, fixes #954.*

It's high time to finally get iterators support in PhpSpec (#954).

I've implemented six matchers for `\Traversable` objects (mostly useful for finite iterators):
 - `shouldHaveCount($count)`
 - `shouldHaveKey($key)` *supports infinite iterators*
 - `shouldHaveKeyWithValue($key, $value)`
 - `shouldContain($value)`
 - `shouldIterate(iterable $iterable)`
 - `shouldStartIterating(iterable $iterable)` *supports infinite interators*

Though `\Traversable` can by only implemented by implementing either `\Iterator` or `\IteratorAggregate` and we can always get `\Iterator` from there, using `iterator_to_array` is not the solution as keys may be duplicated, made of something else than string or integer, it can be infinite and it would use more memory. Therefore, `foreach` was used to make those matchers working.

TODO:
 - before PR review:
   - [x] Better exception messages for `IterateMatcher`
   - [x] Support for infinite iterators in `StartIteratingMatcher`
 - after PR review:
   - [x] Update documentation for existing matchers
   - [x] Add documentation for new matchers


pamil:
I've named the new matchers `shouldIterate` and `shouldStartIterating`, but I'm open for any other names propositions.

src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        return is_array($variable) || $variable instanceof \Traversable;
+    }
+
+    /**
+     * @param array|\Traversable $expected
+     *
+     * @return \Iterator
+     */
+    private function createIteratorFromIterable($expected)
```
dantleech: `$expected`? should this be `$iterable` ?

spec/PhpSpec/Matcher/StartIteratingMatcherSpec.php
```diff
@@ -0,0 +1,158 @@
...
+            ->shouldNotThrow()
+            ->during('positiveMatch', [
+                'startIterating',
+                $this->createGeneratorReturningArray(['a' => 'b', 'c' => 'd']),
+                [$this->createGeneratorReturningArray(['a' => 'b'])],
+            ])
+        ;
+    }
+
+    function it_positive_matches_infitite_generator_while_starting_iterating_the_same()
```
ciaranmcnulty: typo here in `infinite` (and a few other places)

src/PhpSpec/Matcher/Iterate/SubjectHasLessElementsException.php
```diff
@@ -0,0 +1,22 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Matcher\Iterate;
+
+class SubjectHasLessElementsException extends \LengthException
+{
+    public function __construct()
+    {
+        parent::__construct('Subject has less elements than expected.');
```
ciaranmcnulty: This should be "fewer" rather than "less"

src/PhpSpec/Matcher/IterateMatcher.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function positiveMatch($name, $subject, array $arguments)
+    {
+        try {
+            $this->iterablesMatcher->match($subject, $arguments[0]);
+        } catch (Iterate\SubjectHasLessElementsException $exception) {
+            throw new FailureException('Expected subject to have the same number of elements than matched value, but it has less.', 0, $exception);
```
ciaranmcnulty: "fewer" here rather than "less'

pamil:
@ciaranmcnulty all comments applied, but I've only changed `SubjectElementDoesNotMatchException` to extend `FailureException`, the other ones behave differently in each matcher.

ciaranmcnulty:
Great, thanks for adding this!

PR #1017, opened 2016-10-19
Use an easier to maintain way of disabling xdebug

unfunco:
Originally had it as `phpenv config-rm xdebug.ini 2>/dev/null || :` but Travis didn't like the colon (null command that returns `0`) and `true` is probably more obvious.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
   fast_finish: true
 
 before_install:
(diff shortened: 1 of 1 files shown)


pamil:
I'd remove `2>/dev/null` part, in case of `phpenv` failing to disable XDebug unexpectedly.

unfunco:
Good point, it might be good to know the reason rather than have Travis send the error to `/dev/null` and having no knowledge of why... I will update.

ciaranmcnulty:
Thanks, that simplifies things a lot

PR #1018, opened 2016-10-20
Fix for custom templates breaking the behat tests

### LINKED ISSUE #1008: Personal templates break the test suite
Not sure how to resolve this but opening an issue as an FAQ

If a user has personal templates in `~/.phpspec` the Behat features relating to code generation will fail, because those templates will be used

unfunco:
Fixes #1008 based on the idea by @stof, I have also implemented a scenario that tests custom templates. I have only implemented a scenario for a class so far, but it's generic enough to be used for alternative templates.

### DIFF
--- a/features/bootstrap/FilesystemContext.php
+++ b/features/bootstrap/FilesystemContext.php
@@ -35,6 +35,15 @@ class FilesystemContext implements Context
         $this->filesystem->mkdir($this->workingDirectory);
         chdir($this->workingDirectory);
 
+        $fakeHomeDirectory = sprintf('%s/fake-home/', $this->workingDirectory);
+        $this->filesystem->mkdir($fakeHomeDirectory . '.phpspec');
+
+        if (!empty($_SERVER['HOMEDRIVE']) && !empty($_SERVER['HOMEPATH'])) {
+            $_SERVER['HOMEPATH'] = substr($fakeHomeDirectory, 2);
+        } else {
+            putenv(sprintf('HOME=%s', $fakeHomeDirectory));
+        }
+
         $this->filesystem->mkdir($this->workingDirectory . '/vendor');
         $this->filesystem->copy(
             __DIR__ . '/autoloader/autoload.php',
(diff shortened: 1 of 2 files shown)


unfunco:
It's failing on Windows, and I have no idea how Windows works... I've done a quick search and have found something called `C:\`, would this be where Windows users would be expected to have their personal `.phpspec` directory?

avant1:
On Windows home directory is read here: https://github.com/phpspec/phpspec/blob/b7dde5f5bf4aa79ec18f00c91f59bf2d81ff80da/src/PhpSpec/Console/ContainerAssembler.php#L341

`HOMEDRIVE` variable should be set to something like 'C:', and `HOMEPATH` to full path to fake home.

pamil:
Nice, didn't think there is some function to set an environmental variable and restore it at the end of script execution.

stof:
@unfunco next time, please avoid mentioning me in your commit message. Each time you push such a commit to github, they send a notification. And each time you amend your commit, it creates a new commit and so retriggers the notif

ciaranmcnulty:
Looks good, if you can get it passing on Windows. Thanks!

unfunco:
I thought that might have worked but it was a 🙈 guess... I'll actually try to get use a Windows machine this week before pushing to this branch again. If anyone else in this thread is a Windows user and knows the fix then feel free...

ciaranmcnulty:
Take a look at how https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Process/ReRunner/WindowsPassthruReRunner.php sets environment / detects Windows is being used

unfunco:
@pamil Are you a Windows user? I didn't get a chance to look at last weekend and this weekend I'm at a wedding this weekend – if you can complete the work you're welcome to squash into a single fix – if not; I did get a chance to launch a "Windows Amazon Workspace" on AWS and that looked promising for testing on Windows, but didn't manage to install PHP or Xdebug, so until I get some time for that this is stuck.

features/bootstrap/FilesystemContext.php
```diff
@@ -35,6 +35,15 @@ public function prepWorkingDirectory()
         $this->filesystem->mkdir($this->workingDirectory);
         chdir($this->workingDirectory);
 
+        $fakeHomeDirectory = sprintf('%s/fake-home/', $this->workingDirectory);
+        $this->filesystem->mkdir($fakeHomeDirectory . '.phpspec');
+
+        if (!empty($_SERVER['HOMEDRIVE']) && !empty($_SERVER['HOMEPATH'])) {
+            $_SERVER['HOMEPATH'] = substr($fakeHomeDirectory, 2);
```
unfunco: This is removing the drive letter, which I have been told is always one letter. It's a bit crude but the `pathinfo` function isn't able to do this, an alternative option might be to replace the first instance of `HOMEDRIVE` within `HOMEPATH` with an empty string, if that's preferred.

Issue #581, opened 2014-12-12
Fails to assert float values properly.


bgallagh3r:
I recently came across an issue when attempting to test a body fat calculator, I set up the tests and ran them only to get: 

```
expected [double:0.995089], but got [double:0.995089]
```

Here is a repo I added with the latest stable build of PHPSpec with a use case: https://github.com/bgallagh3r/PhpspecTest

Essentially when I run `$this->calculateBodyDensity()->shouldReturn(0.995089)`  it fails, even trying ->shouldBeLike() fails.

Casting the return value & expected value to **int** or **string** allows the tests to pass.

I'm not certain if I'm just being a complete halfwit or what.  But any help, and confirmation that I'm not completely out of my mind would be appreciated.

MarioBlazek:
Hi all, I would like to start contributing to phpspec project. This feature looks very interesting, can I go for approach 1 and implement "shouldBeApproximately" matcher ?

ciaranmcnulty:
Sure, the name might not be the best and it's unclear what level of precision to aim for (it should probably be a parameter) so if you want to think about those things and do a PR it'd be great.

2. would be something like changing the shouldBe failure message to include 'did you mean shouldBeApproximately?' or some similar nudge, when both values are float.

MarioBlazek:
Level of precision should definitely be a parameter. I will implement it, and do PR.

ciaranmcnulty:
Actually in terms of priority, a message when they're both floats would add a lot of value and be simpler to implement. It depends how much time you can devote to it.

nocive:
:+1:

PR #1020, opened 2016-10-30
ApproximatelyMatcher to match two float number using a delta parameter

brainrepo:
This pull request try to create a matcher that checks float value is close to given (with a certain precision). See #581.

### DIFF
--- a/src/PhpSpec/Matcher/CloseFloatMatcher.php
+++ b/src/PhpSpec/Matcher/CloseFloatMatcher.php
@@ -1,0 +1,89 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Matcher;
+
+use PhpSpec\Exception\Example\FailureException;
+use PhpSpec\Exception\Example\NotEqualException;
+use PhpSpec\Formatter\Presenter\Presenter;
+
+final class CloseFloatMatcher extends BasicMatcher
+{
+
+    /**
+     * @var array
+     */
+    private static $keywords = array(
+        'beACloseFloat'
+    );
+
+    /**
+     * @var Presenter
+     */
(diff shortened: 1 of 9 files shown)


ciaranmcnulty:
This looks really useful - I'm not sure about the matcher name though? maybe `shouldBeApproximately` instead / as well?

src/PhpSpec/Matcher/CloseFloatMatcher.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @param mixed $subject
+     * @param array $arguments
+     *
+     * @return bool
+     */
+    protected function matches($subject, array $arguments)
+    {
+        $value = (double)$arguments[0];
```
unfunco: `(double)` should be `(float)` here so that it conforms with PSR-2.

src/PhpSpec/Matcher/CloseFloatMatcher.php
```diff
@@ -0,0 +1,89 @@
...
+        $precisionFactor = pow(10, -1 * (int)$arguments[1]);
```
unfunco: You could use the `**` operator here since phpspec supports 5.6+

unfunco:
Also, what is the behaviour currently if `$precision` is greater than the precision defined within the INI file, e.g. `ini_get('precision');`?

unfunco:
Having checked the source, `PHP_FLOAT_EPSILON` is just a copy of `DBL_EPSILON` from `float.h` – which on my machine is `2.2204460492503131e-16`, and in PHP yields `2.2204460492503e-16`, which you could use as the default, but that assumes phpspec is running on 64-bit architecture, so until phpspec supports 7.2+ and above, it would probably be better to default to `1.0e-10` if you were to go down the delta route instead.

brainrepo:
Sorry, my english is not good.
I dont like **CloseFloatMatcher** name but i choose that because **shouldBeApproximately** can be referred to strings. What you think about?

unfunco:
Approximately is almost exclusively used to refer to numbers in this context, I don't think `shouldBeApproximately` would cause confusion with which data type should be passed to the method.

I don't mind assisting with the additional changes (depending on the consensus, of-course) and using the `--author` flag in Git to mark the changes as being yours, since you've introduced the feature.

brainrepo:
There is no problem for me :)

ciaranmcnulty:
this looks great, thanks!

PR #1022, opened 2016-11-02
Implements Best Match Locator

mahalay:
* Replaces PrioritizedResourceManager
* Adds `suite` option to describe command
* Potentially renders ResourceLocator::getPriority() deprecated

============================================================
Composer allows for classes of the same namespace to be located on multiple paths. Phpspec makes the decision which locator to use based on `priority`. It would be better if the developer is informed that a class being described matches multiple suites and thus be allowed to specify a suite to use. This will make Phpspec easier to work with when used on projects using old frameworks (e.g., Code Igniter).

### DIFF
--- a/spec/PhpSpec/Locator/BestMatchResourceManagerSpec.php
+++ b/spec/PhpSpec/Locator/BestMatchResourceManagerSpec.php
@@ -15,8 +15,8 @@ class BestMatchResourceManagerSpec extends ObjectBehavior
         $locator2->getPriority()->willReturn(10);
     }
 
-    function it_locates_resources_using_all_registered_locators($locator1, $locator2,
-        Resource $resource1, Resource $resource2, Resource $resource3
+    function it_locates_resources_using_all_registered_locators(
+        $locator1, $locator2, Resource $resource1, Resource $resource2, Resource $resource3
     ) {
         $this->registerLocator($locator1);
         $this->registerLocator($locator2);
@@ -33,8 +33,8 @@ class BestMatchResourceManagerSpec extends ObjectBehavior
         $this->locateResources('s:query')->shouldReturn(array($resource1, $resource3, $resource2));
     }
 
-    function it_locates_all_locators_resources_if_query_string_is_empty($locator1, $locator2,
-                                                                        Resource $resource1, Resource $resource2, Resource $resource3
+    function it_locates_all_locators_resources_if_query_string_is_empty(
+        $locator1, $locator2, Resource $resource1, Resource $resource2, Resource $resource3
     ) {
         $this->registerLocator($locator1);
         $this->registerLocator($locator2);
--- a/spec/PhpSpec/Locator/BestMatchResourceManagerSpec.php
+++ b/spec/PhpSpec/Locator/BestMatchResourceManagerSpec.php
@@ -1,0 +1,115 @@
+<?php
+
+namespace spec\PhpSpec\Locator;
+
+use PhpSpec\Exception\Locator\ResourceCreationException;
+use PhpSpec\Locator\Resource;
+use PhpSpec\Locator\ResourceLocator;
+use PhpSpec\ObjectBehavior;
+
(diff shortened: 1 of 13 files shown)


src/PhpSpec/Locator/BestMatchResourceManager.php
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * @param  string $classname
+     *
+     * @return array
+     */
+    private function sortAndGroupLocatorsByMatchScore($classname)
+    {
+        $matchedLocators = [];
+        foreach ($this->locators as $locator) {
+            if (($matchScore = $matchScore = $locator->calculateMatchScore($classname)) > 0) {
```
mheki: `$matchScore = $matchScore = ` mistake here

src/PhpSpec/Locator/BestMatchResourceManager.php
```diff
@@ -0,0 +1,123 @@
...
+    private $locators = array();
+
+    /**
+     * @param ResourceLocator $locator
+     */
+    public function registerLocator(ResourceLocator $locator)
+    {
+        $this->locators[] = $locator;
+
+        @usort($this->locators, function (ResourceLocator $locator1, ResourceLocator $locator2) {
```
mheki: out of curiosity - why `@` here?

spec/PhpSpec/Locator/BestMatchResourceManagerSpec.php
```diff
@@ -0,0 +1,115 @@
...
+
+        $resource1->getSpecClassname()->willReturn('Some\Spec1');
+        $resource2->getSpecClassname()->willReturn('Some\Spec2');
+        $resource3->getSpecClassname()->willReturn('Some\Spec3');
+
+        $this->locateResources('s:query')->shouldReturn(array($resource1, $resource3, $resource2));
+    }
+
+    function it_locates_all_locators_resources_if_query_string_is_empty($locator1, $locator2,
+                                                                        Resource $resource1, Resource $resource2, Resource $resource3
```
mheki: something is messed up with whitespaces here, could you make it more PSR-2 compliant, please?

src/PhpSpec/Locator/BestMatchResourceManager.php
ciaranmcnulty: There's some historical issue that needs it - I think PHP believes the array has changed while being sorted and issues a warning?

mahalay:
Thanks for the feedback. Corrections were made as suggested.

mheki:
thanks @mahalay! Would be great if new features were in sync with the docs. Can you describe your changes in docs, please?

mahalay:
Will work on it after office... Why have I forgotten to update the docs. ;)

mahalay:
hey @mheki, I've uncovered a scenario where the current implementation will fail. I'll fix that first then finish the docs.

mahalay:
Updated the docs and added the scenario where the implementation was failing.

Just curious, what do you guys use for coding standard testing? I tried using squizlabs/php_codesniffer and specified PSR2 as the coding standard and there are a lot of codes with warnings.

ciaranmcnulty:
We're using the scrutiniser PRS-2 rules

ciaranmcnulty:
@mahalay can you resolve the conflicts please?

ciaranmcnulty:
Something now seems to be breaking on PHP 7.2

mahalay:
Hey @ciaranmcnulty, installed PHP 7.2 from my end but did not get any errors when I run the tests. Will try to figure out the issue.

ciaranmcnulty:
Ah it looks like it's the dev dependencies build that's failing https://travis-ci.org/phpspec/phpspec/jobs/288129896

mahalay:
The error was due to the Exception class not being displayed along with the exception message.

ciaranmcnulty:
Can you rebase this please?

mahalay:
Rebase done, but fails tests on Windows - will setup up a Win VM later

PR #1023, opened 2016-11-08
Docblocks

Sam-Burns:


src/PhpSpec/Wrapper/Subject/ExpectationFactory.php
```diff
@@ -115,7 +115,7 @@ private function createNegative($name, $subject, array $arguments = array())
      * @param mixed  $subject
      * @param array  $arguments
      *
-     * @return Expectation
+     * @return Expectation|ThrowExpectation
```
stof: ThrowExpectation is an Expectation, so is it necessary ?

ciaranmcnulty: You seem to be right but that's a worry - from what I can see all instances of `ThrowExpectation` have a method compatible with `Expectation` which is called as if they're an expectation.

Seems to be a bug - the most straightforward fix would be to make the one interface extend the other. Arguably that would break BC because there's a chance someone will have used `ThrowExpectation` without implementing a `match()` method, but I would expect that code to be non-functional.

Sam-Burns: Ref https://github.com/phpspec/phpspec/pull/1024

ciaranmcnulty: Can you revert this in your PR please?

Sam-Burns:
@ciaranmcnulty Ready for merging?

ciaranmcnulty:
Thanks!

PR #1024, opened 2016-11-08
Alterations to interface inheritance to reflect the fact of a thrown …

Sam-Burns:
…expectation being a kind of expectation

### DIFF
--- a/src/PhpSpec/Wrapper/Subject/Expectation/ThrowExpectation.php
+++ b/src/PhpSpec/Wrapper/Subject/Expectation/ThrowExpectation.php
@@ -13,7 +13,7 @@
 
 namespace PhpSpec\Wrapper\Subject\Expectation;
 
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
From what I can see all classes that implement `ThrowExpectation` are going to get their `match()` method called, so I can't see a situation where this change would cause a BC break (i.e. the code will already be broken).

I'm nervous though - @stof what do you think?

ciaranmcnulty:
A further look shows that anything constructed by the ExpectationFactory will immediately have `match()` called on it:
https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Wrapper/Subject.php#L340-L346

So we only really need to worry about people using the interface _outside_ of PhpSpec - something I don't think we need to support.

ciaranmcnulty:
Thanks @Sam-Burns

PR #1025, opened 2016-11-09
[WIP] Fix errors on PHP 7

### LINKED ISSUE #999: Infinite loop silently fails
Given a spec like:

```php
    {
        $this->shouldHaveType(TaskCollection::class);
    }

    function it_is_countable()
    {
        $this->shouldImplement(Countable::class);
    }

    function it_counts_elements_in_collection(Task $task)
    {
        $this->count()->shouldBe(0);

        $this->add($task);
        $this->count()->shouldBe(1);
    }
```

And a class like:

```php
class TaskCollection implements Countable
{
    protected $tasks;

    public function add(Task $task)
    {
        $this->tasks[] = $task;
    }

    public function count()
    {
        return $this->count();
    }

    public function at($index)
    {
        if ($index >= $this->count()) {
            throw new OutOfRangeException;
        }

        return $this->tasks[$index];
    }
}
```

When running phpspec you'll get *GREEN*:

```
      TodoBundle\TaskCollection

  13  ✔ is initializable
  27  ✔ is countable
```

Now, I would expect to see also:

```
  32  ✘ counts elements in collection
        <some error message about the infinite loop?>
```

It took me a while to notice, as a was doing a bit refactoring. Only when I added new spec about `remove()` I wasn't asked to create the method (which made me look read carefully console output).

Also a lesson for myself `GREEN !== OK`

unfunco:
Sort of resolves issue #999 – when removing the `@php:~5.6` annotation, errors are not being output in PHP 7 since most fatals are now an instance of `\Error` – this initial work fixes that but I kind of think the `Shutdown` and `ConsoleFormatter` (since `displayFatal` is always called even if no error has occurred) are in need of some refactoring, at the moment the `Shutdown` process is only used to handle fatals, but will there ever be use-cases for registering custom shutdown methods?

When I write _sort of_ resolves #999, this fixes the issue with fatal errors being shown (in a formatted way) in PHP 7+, but the issue in #999 would have looped infinitely unless Xdebug is enabled and has a `max_nesting_limit` set, and when Xdebug throws an `E_ERROR` the custom registered shutdown function doesn't get reached.

### DIFF
--- a/features/exception_handling/developer_is_notified_which_example_caused_a_fatal_error.feature
+++ b/features/exception_handling/developer_is_notified_which_example_caused_a_fatal_error.feature
@@ -1,4 +1,3 @@
-@php:~5.6
 Feature: Developer is notified of which scenario caused a fatal error
   As a Developer
   I want to know in which scenario or example my script was running
--- a/spec/PhpSpec/Process/Shutdown/ShutdownSpec.php
+++ b/spec/PhpSpec/Process/Shutdown/ShutdownSpec.php
@@ -3,19 +3,21 @@
 namespace spec\PhpSpec\Process\Shutdown;
 
 use PhpSpec\ObjectBehavior;
+use PhpSpec\Process\Shutdown\Shutdown;
 use PhpSpec\Process\Shutdown\ShutdownAction;
 use Prophecy\Argument;
 
 class ShutdownSpec extends ObjectBehavior
 {
-    function it_has_type_shutdown()
+    function it_is_initializable()
     {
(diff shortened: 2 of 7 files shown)


ciaranmcnulty:
@unfunco Any idea why this is failing on travis?

Is this a bug fix (i.e. should target 2.x?)

unfunco:
Sorry, I should have marked as [WIP] – I'll take another look at it tomorrow.

tomasliubinas:
Any progress with this one?

ciaranmcnulty:
I think this is out of date?

PR #1034, opened 2016-12-01
Avoid using the Application::getTerminalWidth() method

veewee:
Since the release of Symfony/console 3.2 yesterday, the Application::getTerminalWidth() method is throwing a deprecated warning.

More info: https://github.com/symfony/console/blob/master/Application.php#L703
This is causing Travis to fail on one of the projects I maintain.

### DIFF
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -22,6 +22,7 @@ use Symfony\Component\Console\Application as BaseApplication;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Terminal;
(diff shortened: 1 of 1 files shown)


veewee:
Thanks for the quick merge! :)

unfunco:
@veewee Are you using version 2 or 3 of PHPSpec?

veewee:
@unfunco 
The problem was in both version 2 and 3.
I was using version 3 for this PR.

unfunco:
Thanks @veewee – I was just checking to prepare new releases, the commit will likely be reverted for the 2.5 branch since 2.5 supports PHP 5.3+ and later versions of Symfony console support later versions of PHP so they're not compatible, I will need to check through the dependencies but will hopefully get newer releases out tomorrow.

stof:
There is no need to revert this commit. There is only the need to make the phpspec code itself compatible with PHP 5.3 (i.e. not using ``::class`` or the inline instantiation)

PR #1036, opened 2016-12-02
Fix a syntactic error in the 2.5.4 branch

unfunco:
Fixing syntactic issues relating to PHP 5.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,7 @@ matrix:
   fast_finish: true
 
 before_install:
-  - if [[ ${TRAVIS_PHP_VERSION:0:4} != "hhvm" ]]; then phpenv config-rm xdebug.ini; fi
+  - phpenv config-rm xdebug.ini || true
 
 install:
   - export COMPOSER_ROOT_VERSION=dev-master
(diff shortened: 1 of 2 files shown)


unfunco:
`::class` – and `(new Terminal)->getWidth()`, and that 2.5 is tested on 5.3+

ciaranmcnulty:
Of course. I forgot how old those changes are. Maybe the way to proceed is to make it work in those versions?

I'm AFK for a few days (on mobile) so if you want to do a release (either of this reversion or a fix) I would be very grateful

unfunco:
And it's still failing on 5.3 anyway, I think because of the symfony dependency in `composer.json`.

unfunco:
The issue is fixed but Travis is reporting the same issues with 5.3 and 5.4, but I'm almost entirely convinced that the issue is in Travis, I confirmed this by restarting a previous build that was previously passing and that's now failing too (https://travis-ci.org/phpspec/phpspec/builds/168078578), I've also downloaded the Travis PHP container from quay.io/travisci/travis-php and everything is passing inside the container.

screen shot 2016-12-04 at 11 32 56

Personally I'd be happy to merge this and tag a new release, as I'm certain the issue is on Travis, and I cannot replicate it, even by using their containers from Quay.io.

ciaranmcnulty:
I'm not happy tagging another patch release on the basis 'probably will work'. I'll try clearing the Travis cache this evening perhaps

unfunco:
I've tried clearing the caches to no avail. And it would be nicer to have it passing, but if you go here: https://travis-ci.org/phpspec/phpspec/builds/168077685 and restart that passing build, that will then fail with the same issues. I'll do a bit more investigation.

PR #1038, opened 2016-12-03
Bump sebastian/exporter version to 2

mattsches:
This prevents installation problems with current versions of PHPUnit.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php":                      "^5.6 || ^7.0",
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "^2.0",
+        "sebastian/exporter":       "^1.0 || ^2.0",
         "symfony/console":          "^2.7 || ^3.0",
         "symfony/event-dispatcher": "^2.7 || ^3.0",
         "symfony/process":          "^2.7 || ^3.0",
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php":                      "^5.6 || ^7.0",
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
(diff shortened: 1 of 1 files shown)


unfunco:
I've checked the [changes between 1.0 and 2.0 of exporter](https://github.com/sebastianbergmann/exporter/compare/1.0.0...2.0.0) and there's nothing that I can see that prevents this from being merged, this is currently breaking my upgrade to PHPUnit 5.7 also.

```
Loading composer repositories with package information
Updating dependencies (including require-dev)
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit 5.7.0 requires sebastian/exporter ~2.0 -> satisfiable by sebastian/exporter[2.0.0].
    - phpunit/phpunit 5.7.1 requires sebastian/exporter ~2.0 -> satisfiable by sebastian/exporter[2.0.0].
    - phpunit/phpunit 5.7.2 requires sebastian/exporter ~2.0 -> satisfiable by sebastian/exporter[2.0.0].
    - Conclusion: don't install sebastian/exporter 2.0.0
    - Installation request for phpunit/phpunit ^5.7 -> satisfiable by phpunit/phpunit[5.7.0, 5.7.1, 5.7.2].
```

The only place I can see this being used is in `ObjectEngine` and the current usage should work fine with 2.0.

ciaranmcnulty:
We can retain support for 1.0 if there aren't any breaking changes on our part - @mattsches can you modify your change to allow either version?

mattsches:
You mean like this: `^1.0||^2.0`?

unfunco:
Yep, the others in `composer.json` that use the same format though have spaces either side of the pipes, so probably best to keep that consistent. ;)

stof:
We should indeed support both (btw, Prophecy was able to support both without any change on its side)

ciaranmcnulty:
Thanks @mattsches!

PR #1041, opened 2016-12-05
Reduce verbosity of the deploy logs

stof:
Listing all files from the vendor folder when removing them makes the log too big for the travis UI for something which is not relevant


ciaranmcnulty:
👍 Can you target 2.5 so we can merge forwards please?

stof:
@ciaranmcnulty done

PR #1042, opened 2016-12-05
Use more compact formatters on CI

stof:
The pretty formatter actually make it really hard to see why the CI build failed, because the failure is somewhere in the middle of a very long scroll.
The progress formatter of Behat and PHPunit (and its dot equivalent in PhpSpec) is actually better for CI, as it gives room to failures rather than to successful cases, and these are the one we want to see on CI (going to see details for a green build is not the common usage for the CI server)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,9 +32,9 @@ install:
   - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
 
 script:
-   - bin/phpspec run --format=pretty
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
I agree. Like the other PR can you target 2.5 branch? Then I will merge forward onto master

unfunco:
Is it also worth doing the same in `appveyor.yml`?

stof:
@unfunco totally. I forgot about it.

@ciaranmcnulty sure. I'm on it

PR #1044, opened 2016-12-09
Deprecate the Resource interface

unfunco:
This deprecates the `Resource` interface and replaces it with `CompositeResource`, the reason for this is that `resource` is soft reserved in PHP 7. I have created a new `CompositeResource` interface and the original `Resource` interface extends this, allowing for an easy removal in the next major.

I chose `CompositeResource` since the interface represents a composite between the specification and the subject, but I'm open to a different name if we can think of something better?

### DIFF
--- a/docs/manual/upgrading-to-phpspec-4.rst
+++ b/docs/manual/upgrading-to-phpspec-4.rst
@@ -1,0 +1,10 @@
+Upgrading to phpspec 4
+======================
+
+Here is a guide to upgrading a test suite or an extension, based on BC-breaking
+changes made in phpspec 4.
+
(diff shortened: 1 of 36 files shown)


ciaranmcnulty:
PhpSpec 4 is at least 6 months away, and PhpSpec 3 will live another 18 months ;-)

ciaranmcnulty:
Only reason I've not merged this is I'm not convinced CompositeResource is a good name - it still feels vague.

ciaranmcnulty:
I'm going to close this, @unfunco  - this naming needs improving but I'm not sure what a good name is yet

PR #1047, opened 2016-12-18
Add missing variable available to the documentation

unfunco:

PR #1048, opened 2016-12-18
Add missing variable available to the documentation

unfunco:
Added missing variable to documentation.

### DIFF
--- a/docs/cookbook/templates.rst
+++ b/docs/cookbook/templates.rst
@@ -80,6 +80,7 @@ template you are overriding:
    - ``%name%``  the specification name
    - ``%namespace%`` the specification namespace
    - ``%subject%`` the name of the class being specified
+   - ``%subject_class%`` the name of the subject class being specified
 
 **class**
    - ``%filepath%`` the file path of the class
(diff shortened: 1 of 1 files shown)


unfunco:
I did fix the other branch much easier to rebase by using the `--onto` flag, however, because that branch has now been `+`pushed...

screen shot 2016-12-23 at 09 05 42

So this branch will need to be the one that's merged (assuming you want it merged) and #1047 can remain closed.

docs/cookbook/templates.rst
```diff
@@ -80,6 +80,7 @@ template you are overriding:
    - ``%name%``  the specification name
    - ``%namespace%`` the specification namespace
    - ``%subject%`` the name of the class being specified
+   - ``%subject_class%`` the name of the subject class being specified
```
jakzal: The description is not really clear. Subject's description could be also clarified. How about:

```
- ``%subject%``the fully qualified name of the class being specified
- ``%subject_class%`` the name of the class being specified
```

ciaranmcnulty: Yep I agree

ciaranmcnulty:
Thanks!

PR #1049, opened 2016-12-29
Update the version number and update CHANGES.md

unfunco:


ciaranmcnulty:
That was quick!

unfunco:
It popped up in Slack, it was a joyous moment. Sad it's aiming for 7.0 and not 7.1 though, glad to lose the HHVM stuff, sad to keep Windows (but apparently they're growing again), have loads of ideas but was waiting to see that 4.0 appear...

ciaranmcnulty:
HHVM is only a temporary removal until I figure out how to add it to travis again (it currently fails to `composer update`)

PR #1050, opened 2016-12-30
[WIP] Remove PHP annotations

unfunco:
Expected to fail and will investigate later.


unfunco:
Updated to just remove the `@php:~5.6` annotation which is causing the failure, the removal of the `@php:~7.0` annotations are now in a different branch and passing, and pointing at 4.0, and I've changed the target branch of this one to master, since this is a bug.

PR #1051, opened 2016-12-30
Updated dependencies and the phpunit xsd

unfunco:
Updated dependencies and use a valid phpunit.xml configuration.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,20 +25,20 @@
         "php":                      "^7.0",
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "^1.0 || ^2.0",
-        "symfony/console":          "^2.7 || ^3.0",
-        "symfony/event-dispatcher": "^2.7 || ^3.0",
-        "symfony/process":          "^2.7 || ^3.0",
-        "symfony/finder":           "^2.7 || ^3.0",
-        "symfony/yaml":             "^2.7 || ^3.0",
+        "sebastian/exporter":       "^2.0",
+        "symfony/console":          "^3.2",
+        "symfony/event-dispatcher": "^3.2",
+        "symfony/process":          "^3.2",
+        "symfony/finder":           "^3.2",
+        "symfony/yaml":             "^3.2",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -25,20 +25,20 @@
...
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "^1.0 || ^2.0",
-        "symfony/console":          "^2.7 || ^3.0",
-        "symfony/event-dispatcher": "^2.7 || ^3.0",
-        "symfony/process":          "^2.7 || ^3.0",
-        "symfony/finder":           "^2.7 || ^3.0",
-        "symfony/yaml":             "^2.7 || ^3.0",
+        "sebastian/exporter":       "^2.0",
+        "symfony/console":          "^3.2",
```
avant1: Do these changes bring any improvements (like removing code to support both 2.7 and 3.x, or we require 3.x because of useful features that we want to use but they exist only in 3.x)?

If I correctly understand how composer version constraints work this will forbid using future versions of `phpspec` with symfony 2.7 (which is LTS and will be maintained until mid 2018).

I'm not sure if updates just for the sake of updates worth it.

unfunco: Some of them will allow code to be removed that was required to support 2.7 and 3.x – another commit has been added to this repository:

https://github.com/phpspec/phpspec/pull/1051/files#diff-bc97c89cde2c88ff4eac8d8009db7b42L86

unfunco: Another comment by Ciaran yesterday is that PhpSpec 3.x will be around for 18 months or so yet. Whatever can be targeted at 3.x will be, it depends on whether PhpSpec 4 should support Symfony 2.7, given that the support lifetime of PhpSpec 3.x and Symfony 2.7 are similar, and how much time is left to support Symfony 2.7 once PhpSpec 4 is released.

avant1: Ah, now I see that PR target branch is 4.0. Probably by the time phpspec 4 will be released symfony 2.7 will be close to its end of life.
Now this makes sense for me, thanks for the explanation.

Issue #1058, opened 2017-01-29
Constructor is made private multiple times


ciaranmcnulty:
Spotted by @carlosV2 during a workshop.

When writing two specs using named constructors, and not running PhpSpec inbetween:

```
class FooSpec extends ObjectBehavior
{
    function it_is_initializable_from_string()
    {
        $this->beConstructedFromString('abc');
        $this->shouldHaveType(Foo::class);
    }
    function it_is_initializable_from_int()
    {
        $this->beConstructedFromInt(123);
        $this->shouldHaveType(Foo::class);
    }
}
```

User is prompted with the following sequence of questions

 * Do you want me to create `Foo::fromString()` for you?                                                                                                     
 * Do you want me to make the constructor of Foo private for you?                                        
 * Do you want me to create `Foo::fromInt()` for you? 
 * Do you want me to make the constructor of Foo private for you?   

Answering yes to all will result in the class having two private methods called `__construct`, consequently causing fatal errors

ciaranmcnulty:
This if fixed by @carlosV2

PR #1059, opened 2017-01-30
Bugfix: Constructor is made private multiple times

carlosV2:
This PR fixes the issue raised in https://github.com/phpspec/phpspec/issues/1058.

===========================================================

The solution is to add a new Generator called `OneTimeGenerator` which decorates any other Generator so that it only calls the parent the first time it is called per classname.

### DIFF
--- a/features/bootstrap/ApplicationContext.php
+++ b/features/bootstrap/ApplicationContext.php
@@ -119,15 +119,7 @@ class ApplicationContext implements Context
      */
     public function iRunPhpspecAndAnswerWhenAskedIfIWantToGenerateTheCode($answer, $option=null)
     {
-        $arguments = array (
-            'command' => 'run'
-        );
-
-        $this->addOptionToArguments($option, $arguments);
-
-        $this->prompter->setAnswer($answer=='y');
-
-        $this->lastExitCode = $this->tester->run($arguments, array('interactive' => true));
+        $this->runPhpSpecAndAnswerQuestions($answer, 1, $option);
     }
 
     /**
@@ -135,12 +127,34 @@ class ApplicationContext implements Context
      */
     public function iRunPhpspecAndAnswerToBothQuestions($answer)
     {
+        $this->runPhpSpecAndAnswerQuestions($answer, 2);
+    }
+
+    /**
+     * @When I run phpspec and answer :answer to the three questions
+     */
+    public function iRunPhpspecAndAnswerToTheThreeQuestions($answer)
+    {
+        $this->runPhpSpecAndAnswerQuestions($answer, 5);
+    }
+
+    /**
+     * @param string  $answer
+     * @param integer $times
+     * @param string  $option
+     */
+    private function runPhpSpecAndAnswerQuestions($answer, $times, $option = null)
+    {
         $arguments = array (
             'command' => 'run'
         );
 
-        $this->prompter->setAnswer($answer=='y');
-        $this->prompter->setAnswer($answer=='y');
+        $this->addOptionToArguments($option, $arguments);
+
+        $i = 0;
+        while ($i++ < $times) {
+            $this->prompter->setAnswer($answer=='y');
+        }
 
         $this->lastExitCode = $this->tester->run($arguments, array('interactive' => true));
     }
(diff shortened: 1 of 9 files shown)


features/bootstrap/ApplicationContext.php
```diff
@@ -120,28 +120,42 @@ public function iRunPhpspec($formatter = null, $option = null, $interactive = nu
...
     {
+        $this->runPhpSpecAndAnswerQuestions($answer, 2);
+    }
+
+    /**
+     * @When I run phpspec and answer :answer to the three questions
+     */
+    public function iRunPhpspecAndAnswerToTheThreeQuestions($answer)
+    {
+        $this->runPhpSpecAndAnswerQuestions($answer, 5);
```
stof: why 5 ?

features/bootstrap/ApplicationContext.php
```diff
@@ -120,28 +120,42 @@ public function iRunPhpspec($formatter = null, $option = null, $interactive = nu
...
-        $this->lastExitCode = $this->tester->run($arguments, array('interactive' => true));
+        $this->runPhpSpecAndAnswerQuestions($answer, 1, $option);
     }
 
     /**
      * @When I run phpspec and answer :answer to both questions
      */
     public function iRunPhpspecAndAnswerToBothQuestions($answer)
     {
+        $this->runPhpSpecAndAnswerQuestions($answer, 2);
```
ciaranmcnulty: I suggest combining this with your new one and transforming `both` / `three` into ints

carlosV2:
### Changes made:
1/ Combined both behat steps into a single one
2/ Add a behat step to make sure that no extra questions are made
3/ Create another decorator for confirming before generating code
4/ Remove functionality out of the NamedConstructorNotFoundListener

### Why the `ConfirmingGenerator`?
Ideally I wanted it to be called the `PromptingGenerator` but it already exists. Instead of reusing the PromptingGenerator I though creating a decorator would be better as it helps the testability, the composition of the generators and the single responsibility of each.

### What about the message?
At the moment, I am not convinced about injecting the message from the container but, potentially, it might help on collecting all the messages and, eventually, allow to easy translate them (if there is any will to). In order to not inject the message, it should be returned from the final generator which would look a bit odd in my opinion. Also I am not convinced about the placeholders like `{CLASSNAME}` in the message. Any ideas for both?

### Final behaviour
After those changes, the final logic is:
1/ Ask the user for creating the named constructor
2/ Ask the user for creating the private constructor
3.../ Ask the user for creating any additional named constructor
Thanks to the `OneTimeGenerator`, it does not ask for creating the private constructor again regardless of the answer on the second question (even if the answer was to not create it). IMHO, I think this is a desired behaviour as if the user doesn't want the private constructor for the first named constructor, then it most likely means that he does not want the private constructor at all.

ciaranmcnulty:
I'm not really happy getting the message from the container - it feels like a premature abstraction to assume the confirminggenerator is reusable but only classname needs to be replaced in the message.

ciaranmcnulty:
OK I think this looks good. @carlosV2 can you rebase against 2.5 as it's a bugfix?

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -320,11 +329,17 @@ private function setupGenerators(IndexedServiceContainer $container)
         }, ['code_generator.generators']);
 
         $container->define('code_generator.generators.private_constructor', function (IndexedServiceContainer $c) {
-            return new CodeGenerator\Generator\PrivateConstructorGenerator(
-                $c->get('console.io'),
-                $c->get('code_generator.templates'),
-                $c->get('util.filesystem'),
-                $c->get('code_generator.writers.tokenized')
+            return new CodeGenerator\Generator\OneTimeGenerator(
```
ciaranmcnulty: Please define these as services as well, so that extensions can add their own decorators

ciaranmcnulty: Actually looking at it, it's not necessary right now

ciaranmcnulty:
Thanks!

PR #1061, opened 2017-02-07
Upgrading test dependencies

Sam-Burns:
Uses PHPUnit 6 for development with PHP 7.  Bumps Behat version a bit.

### DIFF
--- a/integration/PhpSpec/Console/Prompter/QuestionTest.php
+++ b/integration/PhpSpec/Console/Prompter/QuestionTest.php
@@ -3,12 +3,13 @@
 namespace integration\PhpSpec\Console\Prompter;
 
 use PhpSpec\Console\Prompter\Question;
+use PHPUnit\Framework\TestCase;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 /**
  * @requires function \Symfony\Component\Console\Helper\QuestionHelper::ask
  */
-class QuestionTest extends \PHPUnit_Framework_TestCase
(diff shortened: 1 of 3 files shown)


integration/PhpSpec/Console/Prompter/QuestionTest.php
```diff
@@ -3,12 +3,13 @@
 namespace integration\PhpSpec\Console\Prompter;
 
 use PhpSpec\Console\Prompter\Question;
+use PHPUnit\Framework\TestCase;
```
unfunco: When the forward compatibility changes were first introduced in PHPUnit, the FQCN was `phpunit\framework\TestCase`, which will cause autoload issues, it was updated in either 5.5 or 5.6 to use the case you're using here, so this part shouldn't be changed until phpspec 4.0 unless the minimum version of PHPUnit is increased.

Sam-Burns: Thanks for that.  I've upgraded the minimum PHPUnit version to `5.5`, so the build passes with `composer update --prefer lowest`.

ciaranmcnulty:
@Sam-Burns can you target 4.0 branch please?

Sam-Burns:
@ciaranmcnulty I wonder if you would consider removing the `bc-break` label from the PR, and consider it for merging into the `master` (PhpSpec v3) branch?

ciaranmcnulty:
D'oh of course it still works with 5.x - my bad

PR #1063, opened 2017-02-12
Add autocomplete helper

fullpipe:
Add autocomplete helper from composer classmap. Very useful for complex projects.

### DIFF
--- a/src/PhpSpec/Console/Command/DescribeCommand.php
+++ b/src/PhpSpec/Console/Command/DescribeCommand.php
@@ -19,6 +19,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\Question;
 use Composer\Autoload\ClassLoader;
+use Symfony\Component\Finder\Finder;
 
 /**
  * Command line command responsible to signal to generators we will need to
@@ -81,33 +82,36 @@ EOF
         $container->get('code_generator')->generate($resource, 'specification');
     }
 
-
     /**
-     * Get composer class loader namespaces.
+     * Get suites namespaces.
      *
      * @return array
      */
     private function getNamespaces()
     {
-        $autoloadFunctions = spl_autoload_functions();
-        $classLoader = null;
+        $container = $this->getApplication()->getContainer();
+        $srcPaths = array();
         $namespaces = array();
 
-        foreach ($autoloadFunctions as $function) {
-            if ($function[0] instanceof ClassLoader) {
-                $classLoader = $function[0];
+        foreach ($container->getByTag('locator.locators') as $locator) {
+            if ($locator instanceof \PhpSpec\Locator\PSR0\PSR0Locator) {
+                $srcPaths[] = $locator->getSrcPath();
             }
         }
 
-        if (!$classLoader) {
-            return $namespaces;
+        $finder = new Finder();
+        foreach ($finder->files()->name('*.php')->in($srcPaths) as $phpFile) {
+            preg_match('/namespace\s+(.*);/', $phpFile->getContents(), $matches);
+
+            if ($matches) {
+                $namespaces[] = $matches[1];
+            }
         }
 
-        foreach ($classLoader->getClassMap() as $class => $file) {
-            $classParts = explode('\\', $class);
-            unset($classParts[count($classParts) - 1]);
-            $namespaces[] = implode('/', $classParts).'/';
-        }
+        $namespaces = array_map(function($ns) {
+            $nsParts = explode('\\', $ns);
+            return implode('/', $nsParts).'/';
+        }, $namespaces);
 
         return array_unique($namespaces);
     }
--- a/src/PhpSpec/Console/Command/DescribeCommand.php
+++ b/src/PhpSpec/Console/Command/DescribeCommand.php
@@ -18,6 +18,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\Question;
+use Composer\Autoload\ClassLoader;
 
 /**
  * Command line command responsible to signal to generators we will need to
@@ -80,28 +81,34 @@ EOF
         $container->get('code_generator')->generate($resource, 'specification');
     }
 
+
+    /**
+     * Get composer class loader namespaces.
+     *
+     * @return array
+     */
     private function getNamespaces()
     {
-        $vendorPath = realpath(__DIR__.'/../../../../../../');
-        $classmapPath = $vendorPath.'/composer/autoload_classmap.php';
-
-        if (!file_exists($classmapPath)) {
-            return array();
-        }
-
-        $classmap = require $classmapPath;
+        $autoloadFunctions = spl_autoload_functions();
+        $classLoader = null;
         $namespaces = array();
 
-        foreach ($classmap as $class => $file) {
-            if (strpos($file, $vendorPath) !== false) {
-                continue;
+        foreach ($autoloadFunctions as $function) {
+            if ($function[0] instanceof ClassLoader) {
+                $classLoader = $function[0];
             }
+        }
 
+        if (!$classLoader) {
+            return $namespaces;
+        }
+
+        foreach ($classLoader->getClassMap() as $class => $file) {
             $classParts = explode('\\', $class);
             unset($classParts[count($classParts) - 1]);
             $namespaces[] = implode('/', $classParts).'/';
         }
 
-        return $namespaces;
+        return array_unique($namespaces);
     }
 }
--- a/src/PhpSpec/Console/Command/DescribeCommand.php
+++ b/src/PhpSpec/Console/Command/DescribeCommand.php
@@ -17,6 +17,7 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
 
 /**
  * Command line command responsible to signal to generators we will need to
@@ -31,7 +32,7 @@ final class DescribeCommand extends Command
         $this
             ->setName('describe')
             ->setDefinition(array(
-                    new InputArgument('class', InputArgument::REQUIRED, 'Class to describe'),
+                    new InputArgument('class', InputArgument::OPTIONAL, 'Class to describe'),
                 ))
             ->setDescription('Creates a specification for a class')
             ->setHelp(<<<EOF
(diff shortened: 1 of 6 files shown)


ciaranmcnulty:
This looks really useful, thanks!

ciaranmcnulty:
This doesn't autocomplete for me - it looks like it's quite sensitive to the location of the composer file.

To do it more cleanly we could perhaps get the composer Classloader object (which is obtainable in `bin/phpspec` where we include the `autoload.php` and use its API to query the namespaces?

fullpipe:
I`m getting `ClassLoader` from `spl_autoload_functions`, looks better

avant1:
I agree, this could be useful.
But current implementation doesn't work for me too.

First of all, seems like now `desc` command will offer to describe classes that belong to dependencies of project. I don't think that developers may want to describe third-party code (the code they do not own):
image

Another possible problem I see is that `\Composer\Autoload\ClassLoader::getClassMap()` returns list of classes only for dependencies with "classmap" autoloader:
```
"autoload": {
    "classmap": ["src/"]
},
```
But it is not the only (and not the most common) way you can define autoload config for composer: https://getcomposer.org/doc/04-schema.md#autoload.
At least this is how your code works for one of my projects (it uses psr-4 autoloading, and my project namespaces are not autocompleted).

Another possible problem is with multiple registered autoloaders (`composer` allows this). For example when you install `phpspec` globally or use it from `phar` binary. For now you use the last `\Composer\Autoload\ClassLoader` autoloader, and it is probably the best we can do.

And is it possible to write scenario to check autocomplete values?

ciaranmcnulty:
@fullpipe New version autocompletes for me - nice to see it in action!

I think so far we don't have any autocomplete prompts, it'd be good if it could fit in more with the visual style of the other prompts. 

Also @avant1 raises a bunch of good points - I think you should look at `\Composer\Autoload\ClassLoader::getPrefixes` / `getPsr4Prefixes` too

fullpipe:
Problems mentioned by @avant1 are real. And best possible solution I see is direct scan of "src" dirs from `suites`.

src/PhpSpec/Console/Command/DescribeCommand.php
```diff
@@ -64,9 +67,52 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
         $container->get('code_generator')->generate($resource, 'specification');
     }
+
+    /**
+     * Get suites namespaces.
+     *
+     * @return array
+     */
+    private function getNamespaces()
```
avant1: As this method becomes more complex maybe it makes sense to introduce a separate class (or even interface) like `NamespacesAutocompleteProvider` and cover it with tests. What do you think?

src/PhpSpec/Console/Command/DescribeCommand.php
```diff
@@ -64,9 +67,52 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+        foreach ($container->getByTag('locator.locators') as $locator) {
+            if ($locator instanceof \PhpSpec\Locator\PSR0\PSR0Locator) {
+                $srcPaths[] = $locator->getSrcPath();
+            }
+        }
+
+        $finder = new Finder();
+        foreach ($finder->files()->name('*.php')->in($srcPaths) as $phpFile) {
+            preg_match('/namespace\s+(.*);/', $phpFile->getContents(), $matches);
```
avant1: This would work for most cases but will be broken for this code:
```php
<?php
namespace Foo {
    class A {}
}
```

Also pattern can match text occurrences.
More reliable way of fetching namespace is using php tokenizer: http://php.net/manual/en/function.token-get-all.php

And BTW this would not work with projets that use PSR0 with underscores autoloading. I'm not sure if we should support it too for autoloading.

src/PhpSpec/Console/Command/DescribeCommand.php
```diff
@@ -64,9 +67,52 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+            if ($matches) {
+                $namespaces[] = $matches[1];
+            }
+        }
+
+        $namespaces = array_map(function($ns) {
+            $nsParts = explode('\\', $ns);
+            return implode('/', $nsParts).'/';
```
avant1: `desc` command takes class name as parameter, and class names use `\` as separator, but you use `/` here. Is it intentional?

src/PhpSpec/Console/Command/DescribeCommand.php
avant1: Or maybe sources directory structure can be analyzed to provide autocompletion. 
That could be more simple and edge-cases free method.

src/PhpSpec/Console/Command/DescribeCommand.php
fullpipe: Sure.

src/PhpSpec/Console/Command/DescribeCommand.php
fullpipe: Yes, token_get_all is safer. I'll use it.

`directory structure` sounds interesting. But what to do with `Test` and `Resources`, just ignore them?

src/PhpSpec/Console/Command/DescribeCommand.php
fullpipe: Your right. I just accustomed to use `/` with "bin / phpspec desc".

src/PhpSpec/Console/Command/DescribeCommand.php
avant1: `Tests` and `Resources` dir names are commonly used in `symfony` bundles, right?
`Resources` can be a valid namespace part outside of bundles, so I'm not sure here.

Even if `phpspec` will show them in autocomplete I think it would be fine.

fullpipe:
@avant1 should I make some more fixes?

src/PhpSpec/Console/Provider/NamespacesAutocompleteProvider.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Console\Provider;
+
+use Symfony\Component\Finder\Finder;
+
+final class NamespacesAutocompleteProvider
```
avant1: `NamespacesAutocompleteProvider` was just a quick guess of possible name for that class, so feel free to change it if more suitable name can be found :smile:

src/PhpSpec/Console/Provider/NamespacesAutocompleteProvider.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Get namespaces from paths.
+     *
+     * @param  array $paths
+     *
+     * @return array of namespases
+     */
+    public function getNamespaces($paths)
```
avant1: You can add `array` type hint here.

src/PhpSpec/Console/Provider/NamespacesAutocompleteProvider.php
```diff
@@ -0,0 +1,70 @@
...
+    {
+        $namespaces = array();
```
avant1: Since `phpspec` requires PHP 5.6 or higher it makes sense to use short array syntax in new code.

src/PhpSpec/Console/Command/DescribeCommand.php
```diff
@@ -17,6 +17,9 @@
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+use Composer\Autoload\ClassLoader;
```
avant1: This seems to be unused.

src/PhpSpec/Console/Command/DescribeCommand.php
```diff
@@ -64,9 +67,37 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $container = $this->getApplication()->getContainer();
         $container->configure();
 
-        $classname = $input->getArgument('class');
-        $resource  = $container->get('locator.resource_manager')->createResource($classname);
+        if ($input->getArgument('class')) {
+            $classname = $input->getArgument('class');
+        } else {
+            $questionHelper = $this->getApplication()->getHelperSet()->get('question');
+            $question = new Question('<info>Enter class to describe: </info>');
```
avant1: One thing that I'm worried about is the way of interaction with `Symfony\Component\Console\Question\Question`.

There is `\PhpSpec\Console\ConsoleIO` class that implements work with console input/output.
And there is also `\PhpSpec\Console\Prompter` interface.
Maybe this asking stuff should be moved there too.

This will allow us to reuse autocompletion functionality in other commands, and it will make possible to write behat scenario to check autocomplete values.

greg0ire:
[See here how to tell vendors and non-vendors apart](https://github.com/symfony/symfony/pull/21539/files#diff-551c3855d7f43179583069216759396aR64), I hope this can be of some help here.

ciaranmcnulty:
@greg0ire If you're thinking of using this also in your PR, adding a component for it with a relevant interface would be good rather than having it in 2 places

greg0ire:
I started something in #1078

src/PhpSpec/Console/Command/DescribeCommand.php
fullpipe: Something like this
```
\PhpSpec\Console\ConsoleIO
    askQuestion(\PhpSpec\Console\QuestionInterface)

\PhpSpec\Console\QuestionInterface
    askQuestion

\PhpSpec\Console\Question\RequestDescribedClass implements QuestionInterface
```
And in `DescribeCommand`
`$classname = $container->get('question.request_decribed_class')->askQuestion()`
?

src/PhpSpec/Console/Provider/NamespacesAutocompleteProvider.php
fullpipe: `\PhpSpec\Console\Provider\NamespacesProvider`? 
And I want to inject `locator.locators` in it directly.

avant1: There is attempt to create namespaces provider interface in #1078
If it makes sense to merge autoloading namespaces provider and autocomplete namespaces provider then you can implement it and name it in a way telling that it collects list of namespaces for autocompletion.

src/PhpSpec/Console/Command/DescribeCommand.php
avant1: To be honest I was thinking about something more simple like adding the following method to `\PhpSpec\Console\Prompter` interface:
```php
public function askQuestion(string $question, array $autocompleteValues = []): string;
```

and implementing it in `\PhpSpec\Console\Prompter\Question` (it already has access to symfony `QuestionHelper`).
But creating hierarchy of questions may be useful too.

I don't want to force you to specific design choice, just wanted to make sure that new console-specific code will be  reusable and testable.

ciaranmcnulty:
@fullpipe now we have a component for listing the available namespaces (see #1078) could you perhaps use that for this feature?

fullpipe:
@ciaranmcnulty yes, sure.

src/PhpSpec/Console/Command/DescribeCommand.php
ciaranmcnulty: This should definitely go through the Prompter 👍

src/PhpSpec/Console/Provider/NamespacesAutocompleteProvider.php
ciaranmcnulty: This can now be replaced by the namespace provider introduced in #1122 - it isn't currently registered as a service but it probably should be!

src/PhpSpec/Console/Command/DescribeCommand.php
ciaranmcnulty: I don't think we should reflect the existence of the symfony Question in our own interface - we can keep the prompter interface simple, I like @avant1's suggestion

src/PhpSpec/Console/Provider/NamespacesAutocompleteProvider.php
ciaranmcnulty: actually you know what - this is different to the namespace provider concept used for the composer autoconfiguration - what you have is actually fine.

ciaranmcnulty:
Thanks for the hard work!

fullpipe:
Great! Thanks! If you'll get some troubles with this, let me know )

Issue #1065, opened 2017-02-18
Warning: mkdir(): Invalid path when src_path is empty


bartosz-zasada:
I've set `src_path` to an empty string, like this:
```
suites:
    main_suite:
        namespace: Bamiz\ScreamingWebshop
        psr4_prefix: Bamiz\ScreamingWebshop\
        src_path: ""
```
Generating classes and methods works fine. However, every time when I execute `run` or `desc` commands, I get the following warning:
`PHP Warning:  mkdir(): Invalid path in (...)/vendor/phpspec/phpspec/src/PhpSpec/Console/ContainerAssembler.php on line 393`

vitorf7:
I have managed to find a solution to hide that warning and submitted a Pull Request for the maintainers to have a look. It will be up to people like @ciaranmcnulty to now have a look and if accepted it would be merged in.

vitorf7:
@bartosz-zasada The PR that I have made for this issue has been accepted and has been merged in by @ciaranmcnulty which means it should be in the next version that is launched.

ciaranmcnulty:
This is fixed in today's releases, thanks to both of you

PR #1066, opened 2017-02-23
Fix PHP Warning when src_path is empty

vitorf7:
Added the check to see if the user has set src_path to an empty string like found in this issue:

[Issue 1065](https://github.com/phpspec/phpspec/issues/1065)

This was causing the a PHP warning when creating the Spec class since it would pass the check:

```php
if (!is_dir($config['src_path'])) {
    mkdir($config['src_path'], 0777, true);
}
```

But then the ```mkdir``` command would not be able to execute since it is passed an empty string. So I added the check to make sure it is also not empty the ```src_path``` and after testing it locally it works fine. I added only to src_path because I can image people might want to have their classes in the root of the folder but did not do the same for spec_path because that would not make sense to be in the root as those should definitely be in a separate folder.

### DIFF
--- a/src/PhpSpec/Console/ContainerAssembler.php
+++ b/src/PhpSpec/Console/ContainerAssembler.php
@@ -357,7 +357,7 @@ class ContainerAssembler
 
                 $config = array_merge($defaults, $suite);
 
-                if (!is_dir($config['src_path']) && !empty($config['src_path'])) {
+                if (!empty($config['src_path']) && !is_dir($config['src_path'])) {
                     mkdir($config['src_path'], 0777, true);
                 }
                 if (!is_dir($config['spec_path'])) {
--- a/src/PhpSpec/Console/ContainerAssembler.php
+++ b/src/PhpSpec/Console/ContainerAssembler.php
@@ -357,7 +357,7 @@ class ContainerAssembler
 
                 $config = array_merge($defaults, $suite);
 
-                if (!is_dir($config['src_path'])) {
+                if (!is_dir($config['src_path']) && !empty($config['src_path'])) {
                     mkdir($config['src_path'], 0777, true);
                 }
                 if (!is_dir($config['spec_path'])) {


src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -389,7 +389,7 @@ private function setupLocator(IndexedServiceContainer $container)
 
                 $config = array_merge($defaults, $suite);
 
-                if (!is_dir($config['src_path'])) {
+                if (!is_dir($config['src_path']) && !empty($config['src_path'])) {
```
jdreesen: Wouldn't it be better to check for `!empty()` first?

vitorf7: Yes you are correct I made that small change as it makes more sense checking if it is not empty first.

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -389,7 +389,7 @@ private function setupLocator(IndexedServiceContainer $container)
 
                 $config = array_merge($defaults, $suite);
 
-                if (!is_dir($config['src_path'])) {
+                if (!empty($config['src_path']) && !is_dir($config['src_path'])) {
```
ciaranmcnulty: I don't like the use of `empty` here (we know it's set from the defaults) - negation alone should be enough I think

ciaranmcnulty:
You can target this at 2.5 (to be included in the last release)

vitorf7:
@ciaranmcnulty thanks for letting me know through Carlos. I have changed the target of the PR to the branch 2.5.

ciaranmcnulty:
@vitorf7 You need to rebase so that the entire history of master branch isn't also part of the PR

vitorf7:
Oh yeah :P oops sorry noob error

vitorf7:
@ciaranmcnulty Everything seems to be fine now. Thanks to @CarlosV2 help :)

PR #1068, opened 2017-02-25
Fix dependencies

Sam-Burns:
Tests previously failed in PHP 7 after `composer update --prefer-lowest`.  Problem related to instantiation of final classes with `doctrine/instantiator` `<1.0.5`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ matrix:
     - php: 5.6
     - php: hhvm
     - php: 7.0
+    - php: 7.1
+      env: DEPENDENCIES='low'
   allow_failures:
     - env: DEPENDENCIES='dev'
   fast_finish: true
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
Can you retarget this to 2.5 please?

Sam-Burns:
@ciaranmcnulty Retargeted to `2.5` branch. Added extra commit changing a Behat test and the `symfony/console` dependency version, to work in PHP 7.1 with `composer install --prefer-lowest`.

ciaranmcnulty:
Why does symfony/console need to be so high? This may cause issues with us testing older symfony sites :/

Sam-Burns:
@ciaranmcnulty When I added a PhpSpec 2.5 build with `composer update --prefer-lowest` in PHP 7.1, the [build failed](https://travis-ci.org/phpspec/phpspec/jobs/208923831#L282).  It was Behat dying half way through with a 255 exit code.

I've investigated again, and it turns out it only does this in 7.1 with `symfony/console` <=2.6, and only when Behat is run with the `--format=progress` flag.  Other formats of output are fine, making this basically a build problem and not an application problem.

That being the case, I'd suggest closing the PR, and I'll do another one that just changes the Doctrine dependency thing on its own.  Up to you though.

PR #1071, opened 2017-03-03
Latest version of PHPUnit

Sam-Burns:


composer.json
```diff
@@ -25,7 +25,7 @@
         "php":                      "^7.0",
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "^2.0",
+        "sebastian/exporter":       "^3.0",
```
stof: please allow both versions in parallel instead, to be nicer for other tools installed in projects (should be dead easy as there is no BC break except the min PHP version)

Sam-Burns:
Looks like this has been taken care of in another PR already.

PR #1072, opened 2017-03-03
Prophecy version with better PHP 7.1 support

Sam-Burns:


stof:
I'm not sure bumping the *minimum* version is actually required here.

ciaranmcnulty:
I don't see a compelling reason to drop support yet (I'm more wary when projects don't have published roadmaps - with Symfony etc we stated a while ago when we'd be dropping support)

PR #1073, opened 2017-03-03
Backport the support of sebastian/exporter 2 to 2.5

stof:
There was no change needed to support it (the main reason to bump the major version was dropping support for older PHP versions).
This is the same than https://github.com/phpspec/phpspec/pull/1038 in PhpSpec 3.

Some projects are stuck to phpspec 2.5 due to PHP version support (I'm talking about Prophecy here). And enforcing usage of ``sebastian/exporter`` forces to install PHPUnit 4.8 when using both phpspec and phpunit through composer (which we do). This is bad news, as PHPUnit 4.8 has only partial support for PHP 7.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "php":                      ">=5.3.3",
         "phpspec/prophecy":         "~1.4",
         "phpspec/php-diff":         "~1.0.0",
-        "sebastian/exporter":       "~1.0",
+        "sebastian/exporter":       "~1.0|~2.0",
         "symfony/console":          "~2.3|~3.0",
         "symfony/event-dispatcher": "~2.1|~3.0",
         "symfony/process":          "^2.6|~3.0",


stof:
btw, this also forbids Prophecy to test its own compatibility with ``sebastian/exporter`` 2, as phpspec forbids us from installing it (while we tell composer that Prophecy itself is compatible and so our users may end up using it)

ciaranmcnulty:
I guess if we wave our hands a little this is a 'bugfix' but bear in mind PhpSpec 2.x is going to be EOL in ~3-4 months.

If Prophecy is going to continue to maintain support for these older versions of PHP then the testing strategy will get complex (e.g. you may need to rely on abandoned PhpSpec versions)

stof:
@ciaranmcnulty Once we complete the switch away from phpdoc doubles, we can probably allow both 2.x and 3.x until we drop support for old PHP versions (using 2.x only for old PHP versions).

PR #1074, opened 2017-03-03
Allow sebastian/exporter v3

remicollet:
As there is no change in code

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php":                      "^5.6 || ^7.0",
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "^1.0 || ^2.0",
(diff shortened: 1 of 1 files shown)


stof:
@ciaranmcnulty as the only BC break is removing support for older PHP versions, I suggest we also backport it to older versions (see #1073). What do you think ?

stof:
@ciaranmcnulty any answer to my comment ?

ciaranmcnulty:
I merged your backport to allow 2.0, will do the same with 3

stof:
OK, then opening the PR

stof:
@ciaranmcnulty PR opened

stof:
@ciaranmcnulty can you then tag a 2.5.x release with the backports in it ?

ciaranmcnulty:
I will when I've gone through the rest of the open PRs

stof:
thanks

PR #1077, opened 2017-03-10
Allow PHP5 keywords methods generation on PHP7

avant1:
Based and depends on the phpspec/prophecy#309
While that PR is not merged tests may fail, so marked as WIP.


avant1:
The related `prophecy` PR is not merged yet, but tests are green.
It is a bit strange, but that is even better, no need to check if installed `prophecy` version supports this functionality.

ciaranmcnulty:
This is great, thanks!

PR #1084, opened 2017-03-31
Allow .dotfile config to aid in .gitignore rules/exporting projects

shrikeh:


ciaranmcnulty:
I'm always unclear which config files should be hidden files and which should not - is there a best practice?

If not then allowing both makes sense. If we're doing it wrong I'd rather allow both and change it in 4.0

shrikeh:
The general Unix convention of hidden dotfiles has been implemented by Travis CI and Scrutinizer, however within the PHP community this hasn't been as widely adopted (`phpcs.xm`l, `phpunit.xml`, `behat.yml`, etc). However, having `.gitattributes` rules that allow a global `.* export-ignore` simplify excluding this and other files.

It's a matter of preference, but the rule of thumb would be to have non-prod files as dotfiles where possible (which of course your tests hopefully aren't being deployed to live).

ciaranmcnulty:
I guess I was hoping for 'what files should be hidden in UNIX' written in the 70s

I don't see any harm allowing both

shrikeh:
Ah well Linus Torvalds did say in his 1976 blog post _Why phpspec, Why Do You Hate Us So_ that "it would only have taken the designers of this YAML file a few minutes to Google and solve this problem. One day a messiah will come and make this config file go from my console".

ciaranmcnulty:
Thanks!

PR #1085, opened 2017-04-01
Removed black background of line numbers for 'pretty' format.

a-ast:
**Problem**
Phpspec displays line numbers if `pretty` format has been chosen for `run` command. 
The background of line numbers is always black. It makes an output difficult to read on light console backgrounds.

**Before:**
phpspec-pretty-format-before

**After:**
phpspec-pretty-format-after


MarcelloDuarte:
:+1:

ciaranmcnulty:
Looks good, needs checking it's still legible on a dark-background console.

This is enough of a bugfix to target the 2.5 branch, too

avant1:
It looks the same with and without this changes on console with black background. So everything seems fine.

ciaranmcnulty:
@a-ast can you please rebase this branch against 2.5 please?

PR #1086, opened 2017-04-06
Remove black background of line numbers for 'pretty' format.

a-ast:
Problem
Phpspec displays line numbers if pretty format has been chosen for run command.
The background of line numbers is always black. It makes an output difficult to read on light console backgrounds.

**Before:**
phpspec-pretty-format-before

**After:**
phpspec-pretty-format-after

**After (on black background):**
phpspec-pretty-format-after-black

**Also looks good on any other background:**
phpspec-pretty-format-after-blue


ciaranmcnulty:
Great, thanks!

PR #1087, opened 2017-04-10
PHP7 typehints

Sam-Burns:
- Adding PHP7 typehints to part of the codebase.
- Removing redundant docblocks.
- Bumping min version of Prophecy to one which is less panicky about scalar typehints.
- Stubbing behaviour of doubles where necessary, to keep tests passing in the presence of return typehinting.


ciaranmcnulty:
Awesome - this may take a while to review :-)

Sam-Burns:
Docs changes pending

PR #1089, opened 2017-05-08
Bump PHPUnit version

Sam-Burns:
Increases PHPUnit version to 6.1, now PHP5 no longer required

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
     "require-dev": {
         "behat/behat":           "^3.3",
         "symfony/filesystem":    "^3.2",
-        "phpunit/phpunit":       "^5.7|^6.0"
+        "phpunit/phpunit":       "^6.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -25,7 +25,7 @@
         "php":                      "^7.0",
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "^1.0 || ^2.0 || ^3.0",
+        "sebastian/exporter":       "^3.1",
```
stof: I would revert this change, as it makes it harder for projects to avoid a compatibility hell (this is a normal requirement, so it affects users). Version 2 of the exporter is also compatible with PHP 7.
As the API of the exporter is the same, there is absolutely no maintenance cost here.

ciaranmcnulty:
If we're not using any of the features I don't see the point TBH

PR #1091, opened 2017-05-08
Start of upgrade docs for PhpSpec 4

Sam-Burns:


docs/manual/upgrading-to-phpspec-4.rst
```diff
@@ -0,0 +1,20 @@
...
+-------------------
+
+If you are using 3rd party **phpspec** extensions, you may have to increase the version numbers for those as well.
+
+As PHP 5 is no longer supported language versions, you will need to upgrade to PHP 7 to use **phpspec** 4.
+
+Upgrading for Extension Authors
+-------------------------------
+
+Several internal interfaces have had scalar typehints and return typehinting added.  This includes:
```
ciaranmcnulty: Rather than list them we can just say that all/many interfaces now have type hints.

For the user documentation above it'll be important to list the interfaces and spell out what to change

Sam-Burns: Sure, can do.  Just sort of copied last year's really.

Do spec authors need to worry about the return typehints in internal classes?  I thought that might be of more interest to extension authors.

ciaranmcnulty: Only place I can think of is in defining new Matchers, but there may be other places we should be aware of

PR #1099, opened 2017-05-18
Avoid Behat version with PHP5.4+ code in it

Sam-Burns:
The build may break on `2.5` branch the next time it runs, because a [Behat file we're using](https://github.com/Behat/Behat/blob/v3.3.1/src/Behat/Testwork/Argument/MixedArgumentOrganiser.php) has had more short array syntax added recently, and does not in fact work in PHP5.3.

This PR restricts the Behat version used to one where the tests will pass.

(Build break observed [on fork](https://travis-ci.org/Sam-Burns/phpspec/jobs/233561625), where the PhpSpec 2.5/PHP5.3 tests failed.)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     },
 
     "require-dev": {
-        "behat/behat":           "^3.0.11",
+        "behat/behat":           "^3.0.11,!=3.3.1",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
Cool, thanks

PR #1103, opened 2017-06-01
Fix HHVM build

Sam-Burns:
Switching Ubuntu version on Travis so HHVM works


Sam-Burns:
This is in relation to build breaks like https://travis-ci.org/phpspec/phpspec/jobs/238255558 .

PR #1102, opened 2017-05-31
Reactivate canary build

Sam-Burns:
Ref https://github.com/phpspec/phpspec/issues/1101

PR #1104, opened 2017-06-01
Fix PHP 7.2 canary build

Sam-Burns:
Proposed alternative to https://github.com/phpspec/phpspec/pull/1012 .
Depends on https://github.com/phpspec/phpspec/pull/1102 .

Fixes PHP 7.2 build.

### DIFF
--- a/src/PhpSpec/Formatter/Html/HtmlIO.php
+++ b/src/PhpSpec/Formatter/Html/HtmlIO.php
@@ -18,9 +18,11 @@ use PhpSpec\IO\IO;
 final class HtmlIO implements IO
 {
     /**
-     * @param $message
+     * @param string       $message
+     * @param integer|null $indent
+     * @param bool         $newline
      */
-    public function write($message)
+    public function write($message, $indent = null, $newline = false)
     {
         echo $message;
     }
--- a/spec/PhpSpec/Listener/MethodNotFoundListenerSpec.php
+++ b/spec/PhpSpec/Listener/MethodNotFoundListenerSpec.php
@@ -68,6 +68,9 @@ class MethodNotFoundListenerSpec extends ObjectBehavior
     function it_does_not_prompt_for_method_generation_if_input_is_not_interactive($exampleEvent, $suiteEvent, $io, MethodNotFoundException $exception)
     {
         $exampleEvent->getException()->willReturn($exception);
+        $exception->getSubject()->willReturn(new \stdClass());
+        $exception->getMethodName()->willReturn('');
+        $exception->getArguments()->willReturn([]);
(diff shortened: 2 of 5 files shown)


ciaranmcnulty:
Are these actual issues or is PHP 7.2 being an idiot?

Sam-Burns:
If you're asking about the 'get_class' commit, the issue is that you are no longer allowed to call `get_class()` on null.  Therefore, several of our tests need `$someException->getSubject()` to return an object to pass in PHP 7.2.

If you mean the method signature thing, it is https://github.com/phpspec/phpspec/blob/2e969c7d8c6437490b7aa0ab51a3302d15bb7211/src/PhpSpec/IO/IO.php#L21 vs https://github.com/phpspec/phpspec/blob/2e969c7d8c6437490b7aa0ab51a3302d15bb7211/src/PhpSpec/Console/ConsoleIO.php#L184 which is causing issues in the 7.2 build lately.

src/PhpSpec/Formatter/Html/HtmlIO.php
```diff
@@ -18,9 +18,11 @@
 final class HtmlIO implements IO
 {
     /**
-     * @param $message
+     * @param string       $message
+     * @param integer|null $indent
+     * @param bool         $newline
      */
-    public function write($message)
+    public function write($message, $indent = null, $newline = false)
```
roukmoute: c'est quoi donc se fait-ce ça ?

ciaranmcnulty:
The get_class change seems to only fix the tests to not trigger the error case, is that right? I've never been clear what the null case is here anyway.

If PHP has broken the ability to add extra optional parameters to inherited methods let's stop that happening because it's a massive BC break

Sam-Burns:
In relation to the `get_class()` stuff, it is lines like https://github.com/phpspec/phpspec/blob/f5a8388e6d6ec0b279a8e09375c5dcf8975c4432/src/PhpSpec/Listener/MethodNotFoundListener.php#L73 which are causing issues.  This is avoided by changing the relevant spec, so that the exception stub actually returns something when you call `getSubject()` on it.  There are several others very similar to this.

ciaranmcnulty:
I see so there's no reason to ever call null here outside of the test suite?

In that case the get_class change is a no-brainer and mergeable. For the arguments I'd rather wait and see if 7.2 really makes a huge BC break like that once stable.

Can you open 2 PRs and we'll merge one, leave the other hanging?

Sam-Burns:
Closing in favour of https://github.com/phpspec/phpspec/issues/1106 and https://github.com/phpspec/phpspec/issues/1107.

Issue #1057, opened 2017-01-27
shouldThrow matcher does not work with Generators


carlosV2:
Today I came with an interesting problem. It seems like PHP does not execute a Generator code if you don't consume it.

This is not correctly handled in phpSpec. As an example, take a look at this spec:
```
<?php

namespace spec;

use GeneratorTest;
use PhpSpec\ObjectBehavior;
use Prophecy\Argument;

class GeneratorTestSpec extends ObjectBehavior
{
    function it_returns_a_generator()
    {
        $this->getGenerator(false)->shouldIterateAs([1, 2, 3]);
    }

    function it_throws_an_exception()
    {
        $this->shouldThrow()->duringGetGenerator(true);
    }
}
```
And this code:
```
<?php

class GeneratorTest
{
    public function getGenerator($throwException)
    {
        if ($throwException) {
            throw new RuntimeException();
        }

        yield 1;
        yield 2;
        yield 3;
    }
}
```

The expected behaviour would be for both tests to pass. However, because the function is returning a Generator, the code execution is deferred until you start consuming the Generator. As this is not happening in the test, no exception is throw and therefore, phpspec complains about no exception being thrown.

carlosV2:
I've been thinking on potential fixes to this problem but I'm starting to think that not all Generators are not testable.

As an example, take a look at the following code:
```
<?php

class GeneratorTest
{
    public function getGenerator()
    {
        for ($i = 0; $i < 100000000; $i++) {
            yield $i;
        }

        throw new RuntimeException('Reached the end of the generator');
    }
}
```

In order to trigger the exception, phpspec will need to traverse 100.000.000 items which is impracticable in real time.

ciaranmcnulty:
As we discussed offline, something like the following is testable via `$this->shouldThrow()->duringFoo()`:

```
public function foo()
{
    if ($condition) {
        throw new Exception();
    }

    return $this->bar();
}

private function bar() {
    yield 1;
}
```

However exceptions only thrown during the generator itself should correctly _not_ be detected by shouldThrow :-/

Perhaps we need `->shouldThrow()->duringIteration()`?

bpolaszek:
Hello there,

I am completely new to phpspec and in order to contribute to an open-source project already using it, my first spec was to assert that iterating over a generator returned by a method will not throw an exception.

Let's assume the following class (to simplify the example, I directly implement `IteratorAggregate`):

```php
class MyClass implements IteratorAggregate
{
[...]

PR #1105, opened 2017-06-01
[Do not merge] Traversable Throw matcher

bpolaszek:
To engage discussion about #1057, here is a proof of concept. 

Todo:

- [x] Match that an exception should be / shouldn't be thrown when looping on a generator / an instanciated object implementing `Traversable`
- [ ] Match that an exception should be / shouldn't be thrown when looping on an `iterable` returned by a method of the object being tested
- [ ] Match that an error should be / shouldn't be triggered when looping on a generator / an instanciated object implementing `Traversable`
- [ ] Match that an error should be / shouldn't be triggered when looping on an `iterable` returned by a method of the object being tested

What are your thoughts?

### DIFF
--- a/features/matchers/developer_uses_traversable_throw_matcher.feature
+++ b/features/matchers/developer_uses_traversable_throw_matcher.feature
@@ -1,0 +1,41 @@
+Feature: Developer uses traversable-throw matcher
+  As a Developer
+  I want a traversable-throw matcher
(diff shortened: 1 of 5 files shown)


ciaranmcnulty:
I think `->shouldThrow()->duringCurrent()` is enough for me (and quite explicit about where during iteration the exception is thrown)

PR #1106, opened 2017-06-01
Making get_class() work in PHP 7.2

Sam-Burns:
Follows discussion on https://github.com/phpspec/phpspec/pull/1104.  Alternative to https://github.com/phpspec/phpspec/pull/1012.  Changes specs to avoid `get_class(null)` issue in PHP 7.2.


ciaranmcnulty:
Thanks, sam!

PR #1107, opened 2017-06-01
Compatibility with PHP7.2 strict method signatures

Sam-Burns:
Follows discussion on https://github.com/phpspec/phpspec/pull/1104.  Makes `IO::write()`, `HtmlIO::write()` and `ConsoleIO::write()` all have the same method signature.  This is to fix build breaks like https://travis-ci.org/phpspec/phpspec/jobs/238440400#L211-L213.

### DIFF
--- a/src/PhpSpec/Formatter/Html/HtmlIO.php
+++ b/src/PhpSpec/Formatter/Html/HtmlIO.php
@@ -18,9 +18,11 @@ use PhpSpec\IO\IO;
 final class HtmlIO implements IO
 {
     /**
-     * @param $message
+     * @param string       $message
+     * @param integer|null $indent
+     * @param bool         $newline
      */
-    public function write($message)
+    public function write($message, $indent = null, $newline = false)
     {
         echo $message;
     }
--- a/src/PhpSpec/IO/IO.php
+++ b/src/PhpSpec/IO/IO.php
@@ -16,9 +16,11 @@ namespace PhpSpec\IO;
 interface IO
 {
     /**
-     * @param string $message
+     * @param string       $message
+     * @param integer|null $indent
+     * @param bool         $newline
      */
-    public function write($message);
+    public function write($message, $indent = null, $newline = false);
 
     /**
      * @return bool


stof:
Are you sure that PHP 7.2 actually forbids addign more arguments ? I think this rather looks like a bug in Prophecy, where it takes the signature from the interface rather than the ConsoleIO class (and so tries to remove arguments again)

stof:
Reading the NEWS file in the php source code, I think I know what happened: when overwriting methods several times, PHP was checking the compatibility with the original signature, not the parent one, and they fixed this bug, which revealed the Prophecy bug.

```php

interface IO
{
    function write($message);
}

class ConsoleIO implements IO
{
    function write($message, $indent = null) {}
}

class ConsoleIODouble extends ConsoleIO
{
    function write($message) {}
}
```

PHP 7.1 and older were validating `ConsoleIODouble::write` against `IO:write` rather than `ConsoleIO::write`

ciaranmcnulty:
@stof ah so inheritance is ok in our code but prophecy is adding an incorrect third level when generating the double?

stof:
yep, according to the error message in the Travis, I think this is what happens

stof:
And the PHP bug is the reason we never figured it out

ciaranmcnulty:
```
interface IO
{
    function write($message);
}

class ConsoleIO implements IO
{
    function write($message, $indent = null) {}
}

$x = (new \Prophecy\Prophet())->prophesize(ConsoleIO::class);
```
This seems to work for me

stof:
@ciaranmcnulty you need to reveal the double, otherwise the double class might not be created.

ciaranmcnulty:
Revealed and even called the method.

I looked at the generated class src and the interface is from the direct parent, so not sure what combination of factors is triggering the error yet

stof:
have you tried putting classes in different files too, so that inheritance is resolved at runtime rather than compile-time ?

ciaranmcnulty:
Still seems ok, but yeah maybe it's the order of loading?

stof:
are you doing your tests on PHP 7.2 (seems like a silly mistake if no, but always worth checking) ?

ciaranmcnulty:
Yes `PHP 7.2.0-dev (cli) (built: Jun  1 2017 01:32:36)` - which displays the error when I add the `ConsoleIODouble` directly as in your example

stof:
@ciaranmcnulty which version of Prophecy are you using ?

ciaranmcnulty:
1.7.0

stof:
yeah, this is the same version than in the failed Travis build. This eliminates a potential explanation

ciaranmcnulty:
Actually, prophesizing the PhpSpec object causing the error in my testing script didn't trigger the error either.

I checked again and the problem is fixed by #1108 

I've not got time right now to see why this edge case was triggering a prophecy error, but from what I can see prophecy is handling these cases fine

PR #1108, opened 2017-06-02
Fix broken typehint triggering PHP 7.2 issues

ciaranmcnulty:
Type hint in the `let` and the spec were mismatched, triggering an issue with Prophecy


stof:
Can we provide a better debugging experience for such kind of mistake for PhpSpec users ?

ciaranmcnulty:
I guess so - since the typehints on the specs aren't actually used this sort of thing is possible

ciaranmcnulty:
Fixed the nightly though 😎

mbeccati:
@ciaranmcnulty @stof nice, my build is green again too! 

https://revive.beccati.com/bamboo/browse/PHP-PHPSPEC-1100

immagine

PR #1110, opened 2017-06-08
Added section about skipping examples

gquemener:
Hello :wave: 

It has been brought to my attention that the pull request containing the documentation about skipping examples had never been merged (https://github.com/phpspec/phpspec2-site/pull/39) because the repository had moved.

So after 3 long years of questioning about how possible it is to skip this damn example, here we go with the solution!


ciaranmcnulty:
Thanks @gquemener - not sure it's best practice but it's certainly a FAQ!

PR #1113, opened 2017-06-15
Fix bug when parameters can be ignored from config file

borNfreee:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no

Problem
================

There is a bug when you add parameters (e.g. `bootstrap`) **after** `extensions` key in the configuration file which leads to ignoring them completely by `$container`.

#### How to reproduce?

Consider the following config:

```yml
suites:
    some_suite:
        namespace: Some

extensions:
    Memio\SpecGen\MemioSpecGenExtension: ~
    
bootstrap: NotExistingBootstrapFile.php
```

Please note the extension `Memio\SpecGen\MemioSpecGenExtension`. 

When I run `bin/phpspec run`, it does not complain about not existing file. And here is a reason:

```php

// PhpSpec\Console\Application class

protected function loadConfigurationFile(InputInterface $input, IndexedServiceContainer $container)
{
    $config = $this->parseConfigurationFile($input);

    foreach ($config as $key => $val) {
        if ('extensions' === $key && is_array($val)) {
            foreach ($val as $class => $extensionConfig) {
                $this->loadExtension($container, $class, $extensionConfig ?: []);
            }
        }
        [...simplified...]
        else {
            $container->setParam($key, $val);
        }
    }
}
```

This code iterates through keys from `phpspec.yml` file and if it is `extensions` key - it loads all extensions, otherwise, it populates container parameter.

So when `bootstrap` parameter is placed **after** `extensions` key, it is handled after all extensions are loaded.

The good example to reproduce this issue is a `Memio\SpecGen\MemioSpecGenExtension` extensions because in `load` method it calls `$container->get('console.io')` which in its turn calls the following PhpSpec's container code:

```php
$container->define('console.io', function (IndexedServiceContainer $c) {
    return new ConsoleIO(
        $c->get('console.input'),
        $c->get('console.output'),
        new OptionsConfig(
            $c->getParam('stop_on_failure', false),
            $c->getParam('code_generation', true),
            $c->getParam('rerun', true),
            $c->getParam('fake', false),
            $c->getParam('bootstrap', false) // <----------- THIS PARAMETER IS NOT YET POPULATED
        ),
        $c->get('console.prompter')
    );
});
```

So, at the time when extension loads `console.io` service and creates `OptionsConfig` object, our parameter `bootstrap` **is not yet loaded**.

There is a simple workaround to move `bootstrap` parameter *before* `extensions` key in the config, in this case, container populates it and then loads extensions.

Proposed solution
============

I just moved code that populates the container with parameters **before** any extension is loaded. 

In this case, even if any extensions calls `$container->get('console.io')` we will have our container populated with all parameters by that time.

That's it!

### DIFF
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -154,20 +154,15 @@ final class Application extends BaseApplication
                 }
             }
             elseif ('matchers' === $key && is_array($val)) {
-                foreach ($val as $class) {
-                    $container->define(sprintf('matchers.%s', $class), function () use ($class) {
-                        return new $class();
-                    }, ['matchers']);
-                }
+                $this->registerCustomMatchers($container, $val);
             }
-
         }
     }
 
     private function populateContainerParameters(IndexedServiceContainer $container, array $config)
     {
         foreach ($config as $key => $val) {
-            if ('extensions' !== $key && 'matchers'!== $key) {
+            if ('extensions' !== $key && 'matchers' !== $key) {
                 $container->setParam($key, $val);
             }
         }
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -145,6 +145,8 @@ final class Application extends BaseApplication
     {
         $config = $this->parseConfigurationFile($input);
 
+        $this->populateContainerParameters($container, $config);
+
         foreach ($config as $key => $val) {
             if ('extensions' === $key && is_array($val)) {
                 foreach ($val as $class => $extensionConfig) {
@@ -152,9 +154,20 @@ final class Application extends BaseApplication
                 }
             }
             elseif ('matchers' === $key && is_array($val)) {
-                $this->registerCustomMatchers($container, $val);
+                foreach ($val as $class) {
+                    $container->define(sprintf('matchers.%s', $class), function () use ($class) {
+                        return new $class();
+                    }, ['matchers']);
+                }
             }
-            else {
+
+        }
+    }
+
+    private function populateContainerParameters(IndexedServiceContainer $container, array $config)
+    {
+        foreach ($config as $key => $val) {
+            if ('extensions' !== $key && 'matchers'!== $key) {
                 $container->setParam($key, $val);
             }
         }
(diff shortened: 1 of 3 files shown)


src/PhpSpec/Console/Application.php
```diff
@@ -145,16 +145,29 @@ protected function loadConfigurationFile(InputInterface $input, IndexedServiceCo
...
+
         foreach ($config as $key => $val) {
             if ('extensions' === $key && is_array($val)) {
                 foreach ($val as $class => $extensionConfig) {
                     $this->loadExtension($container, $class, $extensionConfig ?: []);
                 }
             }
             elseif ('matchers' === $key && is_array($val)) {
-                $this->registerCustomMatchers($container, $val);
+                foreach ($val as $class) {
```
avant1: Why have you replaced `registerCustomMatchers()` call with partial content of that method? You missed the code that ensured that passed matcher class is really a Matcher. Thats why the tests are red.

For me it makes sense to leave this block unchanged, as it seems to not be related to the PR.

src/PhpSpec/Console/Application.php
```diff
@@ -145,16 +145,29 @@ protected function loadConfigurationFile(InputInterface $input, IndexedServiceCo
...
             }
-            else {
+
+        }
+    }
+
+    private function populateContainerParameters(IndexedServiceContainer $container, array $config)
+    {
+        foreach ($config as $key => $val) {
+            if ('extensions' !== $key && 'matchers'!== $key) {
```
avant1: Missing space in `'matchers'!== $key`

src/PhpSpec/Console/Application.php
borNfreee: It's a stupid mistake, sorry about that. You are completely right

src/PhpSpec/Console/Application.php
borNfreee: fixed

borNfreee:
Finally, tests are green. Please review it again @avant1

ciaranmcnulty:
This looks good to me - as it's a bug fix can you please target this PR to the 2.x branch? We can include it in the final 2.5 release

ciaranmcnulty:
Actually this is different enough on 2.5 to need a manual backport.  Thanks!

PR #1122, opened 2017-07-09
Autodetect namespaces from composer

greg0ire:
Might need more tests, but I have something that works for me ™️


src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -18,29 +18,30 @@
...
+use PhpSpec\Util\ClassFileAnalyser;
+use PhpSpec\Util\Filesystem;
+use PhpSpec\Util\MethodAnalyser;
+use PhpSpec\Util\ReservedWordsMethodNameChecker;
 use PhpSpec\Wrapper;
-use PhpSpec\Config\OptionsConfig;
+use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\Process\PhpExecutableFinder;
-use PhpSpec\Message\CurrentExampleTracker;
-use PhpSpec\Process\Shutdown\Shutdown;
```
greg0ire: My editor reorder the statements, I put that change in a separate commit in case you guys don't want it.

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -390,8 +391,26 @@ private function setupLocator(IndexedServiceContainer $container)
         });
 
         $container->addConfigurator(function (IndexedServiceContainer $c) {
-            $suites = $c->getParam('suites', array('main' => ''));
+            $suites = [];
+            $namespaceProviderFactory = new Factory();
```
greg0ire: ATM there is only one namespace provider... tell me if you fell this is overkill, I will just remove the factory and simplify the configuration to something less forward compatible, like `composer_namespace_provider: ~`

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -390,8 +391,26 @@ private function setupLocator(IndexedServiceContainer $container)
...
+            foreach ($c->getParam('namespace_providers', []) as $namespaceProviderId => $namespaceProviderConfig) {
+                $namespaceProvider = $namespaceProviderFactory->getProvider(
+                    $namespaceProviderId,
+                    $namespaceProviderConfig
+                );
+                foreach ($namespaceProvider->getNamespaces() as $namespace => $location) {
+                    $suites[md5($namespace)] =  [
```
greg0ire: That `md5` is ugly, any recommendation for a slugification method I should use to generate a decent name for the suites?

roukmoute: `uniqid($namespace)` ?

greg0ire: Still not great, but better I guess :)

features/config/developer_can_use_composer_to_create_suites.feature
```diff
@@ -0,0 +1,31 @@
...
+  So I may enable the Composer namespace provider and get one suite per autoload rule
+
+  Scenario: Using composer as namespace_provider
+    Given the config file located in "." contains:
+      """
+      namespace_providers:
+          - composer: ~
+      """
+    When I start describing the "MilkyWay/OrionCygnusArm/LocalBubble" class with the "phpspec.yml" custom config
+    Then a new spec should be generated in the "spec/LocalBubbleSpec.php":
```
greg0ire: I just fixed this expectation, that was plain wrong, and now see that I must compute the psr4 prefix by analyzing the location and the namespace.

src/PhpSpec/NamespaceProvider/NamespaceLocation.php
```diff
@@ -0,0 +1,34 @@
+<?php
+
+namespace PhpSpec\NamespaceProvider;
+
+use Composer\Autoload\ClassLoader;
+
+final class NamespaceLocation
```
greg0ire: Tell me if you prefer a simple array instead of this POPO

roukmoute: Why prefer an array instead of immutable object?

greg0ire: I don't know... less code?

ciaranmcnulty:
This needs a rebase against master

greg0ire:
Rebased @ciaranmcnulty

docs/cookbook/configuration.rst
```diff
@@ -49,6 +52,15 @@ With this config running:
 will now put the spec in ``spec/MarkdownSpec.php`` and the class will be created
 in  ``src/Markdown.php``.
 
+Alternatively, you can choose to use Composer to provide the necessary
+configuration:
+
+.. code-block:: yaml
+
+    namespace_providers:
```
ciaranmcnulty: This could be more user-friendly - I think the concept of a namespace provider is pretty internal.

Maybe a simple boolean here?

features/bootstrap/autoloader/autoload.php
```diff
@@ -3,3 +3,27 @@
     $classname = __DIR__ . '/../src/' . str_replace("\\", "/", trim($classname, "\\")) . ".php";
     if (file_exists($classname)) { include $classname; }
 });
+if (!class_exists('FakeLoader')) {
```
ciaranmcnulty: Having this class created in the bootstrap makes the behaviour in the gherkin scenarios harder to understand.

Can it please be generated based on a Given in those steps? 

Currently to understand the behaviour in a particular Scenario I have to refer to this file

src/PhpSpec/Console/ContainerAssembler.php
ciaranmcnulty: please simplify it - other extensions can register new providers but they can provide their own configuration to do so

src/PhpSpec/NamespaceProvider/NamespaceLocation.php
ciaranmcnulty: Value objects are good :-)

docs/cookbook/configuration.rst
ciaranmcnulty: I'd like this behaviour to be the default in PhpSpec 5 so it's important we have a syntax that supports negation too.

Therefore maybe a boolean like `detect_composer_namespaces` is enough?

ciaranmcnulty:
I've removed the WIP title because this is already 'good enough' and can only get more polished before merge :-0

features/bootstrap/autoloader/autoload.php
greg0ire: I think I could, but you would still have to refer to this file because of the `return FakeLoader();` statement that still would have to be here.

docs/cookbook/configuration.rst
greg0ire: > Therefore maybe a boolean like detect_composer_namespaces is enough?

I'm afraid it's not, because you can't pass configuration to the namespace provider with a boolean, which means you cannot customize the `root_directory` or the `spec_prefix` arguments.

greg0ire: I forgot to document this BTW, it seems, so let's do it

features/bootstrap/autoloader/autoload.php
ciaranmcnulty: I think that would be ok - the key is to get the 'why' into the scenario

features/config/developer_can_use_composer_to_create_suites.feature
```diff
@@ -9,6 +9,7 @@ Feature: Composer can be leveraged to create suites
       """
       composer_namespace_provider: true
       """
+      And the fake autoloader is setup
```
ciaranmcnulty: I want a reader of just the feature to understand the relationship between When and Then

In the Gherkin it should be something like:
`And there is a PSR-4 namespace "MilkyWay\OrionCygnusArm\" configured for the "src" folder`

Currently the When/Then relationship is only understandable if you read the autoload file

ciaranmcnulty:
Great work, thank you!

PR #1129, opened 2017-08-02
Fix invalid typehint on Boolean instead of bool

nightlinus:
Cause of exception [err:TypeError("Return value of PhpSpec\Runner\Maintainer\ErrorMaintainer::errorHandler() must be an instance of PhpSpec\Runner\Maintainer\Boolean, boolean returned")] has been thrown.

### DIFF
--- a/spec/Phpspec/Runner/Maintainer/ErrorMaintainerSpec.php
+++ b/spec/Phpspec/Runner/Maintainer/ErrorMaintainerSpec.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace spec\Phpspec\Runner\Maintainer;
+
+use PhpSpec\Exception\Example as ExampleException;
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Thanks for the fix, what caused it to be triggered?

nightlinus:
multiple tests of code that use trigger_error to generate custom warning:

    function it_can_be_created_without_expiration_date()
    {
        $this->beConstructedWith($this->name, true, $this->comment);
        $this->shouldNotThrow()->duringInstantiation();
    }

    function it_shows_that_feature_is_expired()
    {
        $this->beConstructedWith($this->name, true, $this->comment, new \DateTimeImmutable('now'));
        $this->shouldTrigger(E_USER_WARNING)->duringInstantiation();
        @$this->shouldBeExpired();
    }

PR #1131, opened 2017-08-02
Clarify inline matcher signature in docs

nightlinus:
without array typehint it generate
PHP Fatal error:  Declaration of spec\*Spec::getMatchers() must be compatible with PhpSpec\Matcher\MatchersProvider::getMatchers(): array


ciaranmcnulty:
Thanks!

PR #1136, opened 2017-08-05
Fix white on white bug in console formatter

MarcelloDuarte:
With a white background, we don't see the exception being printed:

screen shot 2017-08-05 at 17 50 51

with this fix we can:

screen shot 2017-08-05 at 17 51 02

Incidentally, this also fixes the formater tag appearing here:

screen shot 2017-08-05 at 17 50 26

and, with the fix:

screen shot 2017-08-05 at 17 50 12


ciaranmcnulty:
Thanks @MarcelloDuarte!

PR #1141, opened 2017-08-24
bugfix: return null values on typed methods on Event classes

### LINKED ISSUE #1137: TypeError thrown when calling `ExampleEvent::getTime()` on event constructed w/ nullable `$time`
Getting into the world of test-driven development, and this tool makes it _really_ awesome!
So thanks for that.

However, I am getting an exception when running phpspecs via PHPStorm's built-in phpSpec runner (PHP 7.1).

Not sure if this is actually an issue with the PHPStorm wrapper, or an incorrect return type in the library. I'm expecting the latter, so that's why I'm reporting an issue over here, first.

```
# This is the command that was executed when I ran my phpspecs from PHPStorm:
$ IDE_PHPSPEC_EXE=/path/to/project/bin/phpspec /usr/local/opt/php71/bin/php /private/var/folders/zl/rlgvjmrj0g75ndwmc2dcv6dc0000gn/T/ide-phpspec.php run --format=phpstorm --no-ansi --no-interaction -c /path/to/project/phpspec.yml.dist spec/AppBundle/FooSpec.php

##teamcity[enteredTheMatrix]

##teamcity[testSuiteStarted name='AppBundle\Foo' locationHint='phpspec://\spec\AppBundle\FooSpec']
 Fatal error happened while executing the following

    Uncaught TypeError: Return value of PhpSpec\Event\ExampleEvent::getTime() must be of the type float, null returned in /path/to/project/vendor/phpspec/phpspec/src/PhpSpec/Event/ExampleEvent.php:142
Stack trace:
#0 /private/var/folders/zl/rlgvjmrj0g75ndwmc2dcv6dc0000gn/T/ide-phpspec.php(65): PhpSpec\Event\ExampleEvent->getTime()
#1 [internal function]: PhpStormFormatter->beforeExample(Object(PhpSpec\Event\ExampleEvent), 'beforeExample', Object(Symfony\Component\EventDispatcher\EventDispatcher))
#2 /path/to/project/vendor/symfony/event-dispatcher/EventDispatcher.php(212): call_user_func(Array, Object(PhpSpec\Event\ExampleEvent), 'beforeExample', Object(Symfony\Component\EventDispatcher\EventDispatcher))
#3 /path/to/project/vendor/symfony/event-dispatcher/EventDispatcher.php(44): Symfony\Component\EventDispatcher\EventDispatcher->doDispatch(Array, 'beforeExample', Object(PhpSpec\Event\E in /path/to/project/vendor/phpspec/phpspec/src/PhpSpec/Event/ExampleEvent.php on line 142
```

## Root cause

I looked into the code of both parties (phpspec + PHPStorm's wrapper).

The problem seems to be originating from this code snippet:

```php
<?php // /vendor/phpspec/phpspec/src/PhpSpec/Event/ExampleEvent.php

class ExampleEvent extends Event implements PhpSpecEvent
{
    /**
     * @var float
     */
    private $time;                     // $time can NOT be null

    public function __construct(
        ExampleNode $example,
        float $time = null,            // $time CAN be null
        int $result = null,
        \Exception $exception = null
    ) {
        // ...
        $this->time = $time;
    }

    // ...

    public function getTime(): float  // $time can NOT be null
    {
        return $this->time;
    }
}
```

Also see the usage example (PHPStorm wrapper script), as mentioned below.

It is currently _always_ possible to call `anExampleEvent->getTime()`. So this also covers the case where you used `null` as value for `$time` in the constructor. This will result in the exception I mentioned above.
In my understanding, there should never be a type exception in the getter, when `null` is actually valid as value for the `$time` argument (either as default value, or manually specified) during construction.

Is calling this method in this case, seen as "unexpected behavior"? E.g., is it "user error", the event being used in the wrong way / context, and should it be fixed in the wrapper script instead? Or would this actually classify as a bug within this framework (what would be my understanding).

### Possible fixes

If this is indeed a 'bug' of some kind, I would expect that either:

  - (1) `ExampleEvent::getTime()` gets a `?float` return type; still float, but nullable

  - (2) `ExampleEvent::$time` defaults to '0.0' in the constructor arguments (or during assignment)
    ```
      __constructor(
    -     float $time = null,
    +     float $time = 0.0,
    ```

    or:

    ```
    - $this->time = $time;
    + $this->time = $time ?? 0.0;
    ```

  - (3) `ExampleEvent::getTime()` forces its return value to be a float:
    ```
    - return $this->time;
    + return $this->time ?? 0.0;
    ```

  - ...

## Varia

> It might be worth noting that this exception is _only_ happening while running specs with the `--format=phpstorm` flag.
> Obviously due the snippet pasted below, which is doing the `$event->getTime()` call during `beforeExample`.

### PHPStorm wrapper script for phpspec

A sample from the cause in the (generated ???) PHPStorm wrapper script (for reference purposes):

```php
<?php /private/var/folders/zl/rlgvjmrj0g75ndwmc2dcv6dc0000gn/T/ide-phpspec.php

class PhpStormFormatter extends \PhpSpec\Formatter\ConsoleFormatter
{
    // ...

    public function beforeExample(\PhpSpec\Event\ExampleEvent $event)
    {
        $testName = $event->getTitle();
        $duration = (int)(round($event->getTime(), 2) * 1000);  // < the culprit of this issue >:)
        $duration = 0;
        $params = array(
            'name' => $testName,
            "duration" => $duration
        );

        $className = $event->getSpecification()->getClassReflection()->getName();
        $methodName = $event->getExample()->getFunctionReflection()->getName();
        $params['locationHint'] = "phpspec://\\$className::$methodName";
        $this->printEvent('testStarted', $params);
    }

    // ...
}
```

### Environment configuration

Some additional information about my local setup:

```
$ php71 -v
PHP 7.1.7 (cli) (built: Jul 19 2017 15:58:29) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies

$ composer show phpspec/*
phpspec/php-diff v1.1.0 A comprehensive library for generating differences between two hashable objects (strings or arrays).
phpspec/phpspec  4.0.2  Specification-oriented BDD framework for PHP 5.6+
phpspec/prophecy v1.7.0 Highly opinionated mocking framework for PHP 5.3+

# PHPStorm
PHPStorm 2017.2.1
Build #PS-172.3544.41, built on August 2, 2017

Plugins:
  PHPSpec BDD Framework: 172.3544.41
```

### Similar issues

Probably the same kinds of issues would apply in case where `$result = null`, and you're calling `ExampleEvent::getResult()`.

So e.g. a default value or nullable return type could fix these cases as well.

## Next steps

If this issue does not belong in this repo, it can be closed.

Otherwise, I'd be glad to help out on this issue (e.g. PRs, support, ...).
I already made some of the above modifications in my local phpspec vendor files, which seem to fix the my issue.

oxkhar:
Default values to null for fields that are returned with typed value launch TypeError 

Fix #1137

### DIFF
--- a/spec/PhpSpec/Event/ExampleEventSpec.php
+++ b/spec/PhpSpec/Event/ExampleEventSpec.php
@@ -55,4 +55,18 @@ class ExampleEventSpec extends ObjectBehavior
     {
         $this->getException()->shouldReturn($exception);
     }
+
+    function it_initializes_a_default_result(ExampleNode $example)
+    {
+        $this->beConstructedWith($example);
+
+        $this->getResult()->shouldReturn($this->PASSED);
+    }
+
+    function it_initializes_a_default_time(ExampleNode $example)
+    {
+        $this->beConstructedWith($example);
+
+        $this->getTime()->shouldReturn((double)0.0);
+    }
 }
--- a/spec/PhpSpec/Event/ExpectationEventSpec.php
+++ b/spec/PhpSpec/Event/ExpectationEventSpec.php
@@ -74,4 +74,14 @@ class ExpectationEventSpec extends ObjectBehavior
     {
         $this->getException()->shouldReturn($exception);
     }
+
+    function it_initializes_a_default_result(ExampleNode $example, Matcher $matcher, $subject)
+    {
+        $method = 'calledMethod';
+        $arguments = array('methodArguments');
+
+        $this->beConstructedWith($example, $matcher, $subject, $method, $arguments);
+
+        $this->getResult()->shouldReturn($this->PASSED);
+    }
 }
--- a/spec/PhpSpec/Event/SpecificationEventSpec.php
+++ b/spec/PhpSpec/Event/SpecificationEventSpec.php
@@ -42,4 +42,18 @@ class SpecificationEventSpec extends ObjectBehavior
     {
         $this->getResult()->shouldReturn(Example::FAILED);
     }
+
+    function it_initializes_a_default_result(SpecificationNode $specification)
+    {
+        $this->beConstructedWith($specification);
+
+        $this->getResult()->shouldReturn(Example::PASSED);
+    }
+
+    function it_initializes_a_default_time(SpecificationNode $specification)
+    {
+        $this->beConstructedWith($specification);
+
+        $this->getTime()->shouldReturn((double) 0.0);
+    }
 }
--- a/spec/PhpSpec/Event/SuiteEventSpec.php
+++ b/spec/PhpSpec/Event/SuiteEventSpec.php
@@ -53,4 +53,18 @@ class SuiteEventSpec extends ObjectBehavior
 
         $this->isWorthRerunning()->shouldReturn(false);
     }
+
+    function it_initializes_a_default_result(Suite $suite)
+    {
+        $this->beConstructedWith($suite);
+
+        $this->getResult()->shouldReturn(Example::PASSED);
+    }
+
+    function it_initializes_a_default_time(Suite $suite)
+    {
+        $this->beConstructedWith($suite);
+
+        $this->getTime()->shouldReturn((double) 0.0);
+    }
 }
--- a/src/PhpSpec/Event/ExampleEvent.php
+++ b/src/PhpSpec/Event/ExampleEvent.php
@@ -76,8 +76,8 @@ class ExampleEvent extends Event implements PhpSpecEvent
      */
     public function __construct(
         ExampleNode $example,
-        float $time = null,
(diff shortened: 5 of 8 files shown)


MarcelloDuarte:
@oxkhar thanks for this. This will indeed fix the problem. Can you please add a spec documenting the cases in which default values are used, that we will not get nulls back. This is to prevent regression. Thanks again!

oxkhar:
Ok, specs defined for default values...
Done!

MarcelloDuarte:
thanks! 🍺

PR #1144, opened 2017-09-02
[bug] Construction with childen interfaces and name names

sroze:
This is a rather dramatic failure:
```
Fatal error: Class Double\Foo\Bar\P3 cannot implement previously implemented interface Runner\ConstructorExample11\Foo in [...]/phpspec/phpspec/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code on line 2
```

I don't have time to fix this just yet, but at least here's a failing test :)


ciaranmcnulty:
From what I can see this appears to be the same as this Prophecy bug (i.e. PhpSpec isn't making things worse) https://github.com/phpspec/prophecy/issues/192

PR #1145, opened 2017-09-04
Fix invalid ErrorMaintainer spec path and namespace

nightlinus:
My previous PR intoduce spec with wrong path and namespace :(. This fix it.


ciaranmcnulty:
Good spot!

PR #1148, opened 2017-09-07
Speed up code analysis

bendavies:
I noticed when running the https://github.com/Sylius/Sylius test suite that the time to wait before the specs starting running was surprisingly long, so I thought I'd investigate.

The issue seems to be that `is_array` is pretty slow, and noticeably so on a large test suite like sylius where it's called 18,689,042 times

Profile before commit:
https://blackfire.io/profiles/1f53b17c-25ce-4534-9d5a-8f80bc5f80fc/graph

After:
https://blackfire.io/profiles/94c981eb-f7cb-48d0-8a69-c3150ad067c7/graph

Comparison:
https://blackfire.io/profiles/compare/c5bb8576-d83c-4ce0-80f4-89d0448aac64/graph

A pretty decent speed up of "-48%"

### DIFF
--- a/src/PhpSpec/CodeAnalysis/TokenizedNamespaceResolver.php
+++ b/src/PhpSpec/CodeAnalysis/TokenizedNamespaceResolver.php
@@ -47,7 +47,7 @@ final class TokenizedNamespaceResolver implements NamespaceResolver
                         $this->currentNamespace = trim($this->currentNamespace);
                         $this->state = self::STATE_DEFAULT;
                     }
-                    elseif (is_array($token)) {
+                    elseif (((array) $token === $token)) {
                         $this->currentNamespace .= $token[1];
                     }
                     break;
@@ -59,7 +59,7 @@ final class TokenizedNamespaceResolver implements NamespaceResolver
                     elseif (',' == $token) {
                         $this->storeCurrentUse();
                     }
-                    elseif (is_array($token)) {
+                    elseif (((array) $token === $token)) {
                         $this->currentUse = $this->currentUseGroup . trim($token[1]);
                     }
                     break;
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Great, thanks!

ciaranmcnulty:
BTW this is crazy that that's slower but LOL PHP

bendavies:
Thanks for the merge! that was fast

fabpot:
Does it give a significant impact on real-world scenarios? I mean, changing a function call by a cast will always have a big impact for Blackfire as the function call is instrumented (so slower), the cast is not.

bendavies:
@fabpot you are absolutely right. i really should have real world tested if first.
I'm going to revert this PR.
The real result is....no change, or actually casting is slightly slower.

ciaranmcnulty:
Oh! I didn't realise blackfire introduced this sort of effect - thanks @fabpot

nicolas-grekas:
The correct fix for this (if there is anything to improve one Blackfire's overhead has been removed), is to prefix the call by a `\\`: this will allow opcache to use not a function call but a much faster instruction of the VM.

bendavies:
@nicolas-grekas nice, i forgot there was a perf gain for doing that. I'll give it a go.

bendavies:
@nicolas-grekas the problem with that, obviously, is opcache being disabled on the CLI by default.

nicolas-grekas:
I'm not even sure this is an opcache-only optim, @jpauli wdyt?

avant1:
Generated opcodes are different.
Adding `\` prefix gives [2 opcodes less](https://3v4l.org/SSalN/vld#output) compared to [version without it](https://3v4l.org/0D2Pv/vld#output).

Since there are less instructions, it should be executed faster, right?

nicolas-grekas:
@avant1 thanks! This means it doesn't rely on opcache, so all good (Yes, a type check is much faster than a function call)

PR #1150, opened 2017-09-07
Revert "is_array is slow as a dog. cast instead"

bendavies:
reverts https://github.com/phpspec/phpspec/pull/1148.
see discussion there.


ciaranmcnulty:
Well, it was good while it lasted :-)

PR #1152, opened 2017-09-12
fully quantify native function invocation

bendavies:
This is the 2nd go at https://github.com/phpspec/phpspec/pull/1148

I have prefixed all native function calls with `\` as suggested by @nicolas-grekas

Again, on the https://github.com/Sylius/Sylius testsuite, (just the code analysis stage)
Blackfire before:
https://blackfire.io/profiles/394774de-ade8-4c03-a62b-be9ab8060b75/graph
After:
https://blackfire.io/profiles/e2ee0855-f82a-4f3b-aaa6-10c78278b421/graph
Comparison:
https://blackfire.io/profiles/compare/7886df20-3b18-4070-a634-552013797fae/graph

Without blackfire:
before:
```
time bin/phpspec run
bin/phpspec run  6.04s user 0.25s system 99% cpu 6.311 total
```
after:
```
time bin/phpspec run
bin/phpspec run  5.40s user 0.22s system 99% cpu 5.636 total
```

Around 10% faster in real terms.

### DIFF
--- a/src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
+++ b/src/PhpSpec/CodeAnalysis/StaticRejectingNamespaceResolver.php
@@ -50,7 +50,7 @@ final class StaticRejectingNamespaceResolver implements NamespaceResolver
             'iterable',
         ];
 
-        if (in_array($typeAlias, $nonObjectTypes, true)) {
+        if (\in_array($typeAlias, $nonObjectTypes, true)) {
             throw new DisallowedNonObjectTypehintException("Non-object type $typeAlias cannot be resolved within a namespace");
         }
     }
--- a/src/PhpSpec/CodeAnalysis/TokenizedNamespaceResolver.php
+++ b/src/PhpSpec/CodeAnalysis/TokenizedNamespaceResolver.php
@@ -47,7 +47,7 @@ final class TokenizedNamespaceResolver implements NamespaceResolver
                         $this->currentNamespace = trim($this->currentNamespace);
                         $this->state = self::STATE_DEFAULT;
                     }
-                    elseif (is_array($token)) {
+                    elseif (\is_array($token)) {
                         $this->currentNamespace .= $token[1];
                     }
                     break;
@@ -59,7 +59,7 @@ final class TokenizedNamespaceResolver implements NamespaceResolver
                     elseif (',' == $token) {
                         $this->storeCurrentUse();
                     }
-                    elseif (is_array($token)) {
+                    elseif (\is_array($token)) {
                         $this->currentUse = $this->currentUseGroup . trim($token[1]);
                     }
                     break;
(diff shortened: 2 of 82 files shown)


src/PhpSpec/CodeAnalysis/MagicAwareAccessInspector.php
```diff
@@ -33,22 +33,22 @@ public function __construct(AccessInspector $accessInspector)
      */
     public function isPropertyReadable($object, string $property) : bool
     {
-        return method_exists($object, '__get') || $this->accessInspector->isPropertyReadable($object, $property);
+        return \method_exists($object, '__get') || $this->accessInspector->isPropertyReadable($object, $property);
```
nicolas-grekas: this one won't change anything, that's always a function call.
on type-related function are replaced by better opcodes.

nicolas-grekas:
There are only a few function where this is useful, you should not change all of them.
Here is the list:
https://github.com/php/php-src/blob/f2db305fa4e9bd7d04d567822687ec714aedcdb5/Zend/zend_compile.c#L3872

src/PhpSpec/CodeAnalysis/MagicAwareAccessInspector.php
bendavies: more good knowledge.

bendavies:
@nicolas-grekas thanks. more education :)

src/PhpSpec/CodeAnalysis/MagicAwareAccessInspector.php
stof: well, it still removes the need to check for a function in the current namespace (but this resolution is cached by PHP and so happens only once per function and namespace).

bendavies:
@nicolas-grekas however, i assume there is no penalty for prefixing other methods?  The plus is that it is consistent in styling.

src/PhpSpec/CodeAnalysis/MagicAwareAccessInspector.php
nicolas-grekas: @stof correct, which the benefit is not measurable in practice :)

avant1:
> i assume there is no penalty for prefixing other methods? The plus is that it is consistent in styling, so i am tempted to leave it.

From my point of view prefixing all function calls does not make much sense here. As I know it is not a common approach for modern PHP projects.
So it will be surprising to all new (and existing) contributors, and it will require us to add a code style check for that (not sure if Scrutinizer supports this) to make sure it will stay consistent in future.

Changing several most frequently called places with a proper comment will bring performance improvement without forcing changes to code style.

bendavies:
@nicolas-grekas @avant1 now only prefixing the list of functions provided by nicolas.

ciaranmcnulty:
Great stuff 👍

How can we add regular performance profiling? Is Blackfire free for open source?

nicolas-grekas:
The Hack edition is free for everyone. For more advanced features, I don't think there is a plan for OSS.

bendavies:
@ciaranmcnulty you could to what doctrine do and run https://github.com/phpbench/phpbench benchmarks as part of the build /cc @ocramius

ciaranmcnulty:
Thanks, @bendavies

bendavies:
Thanks @ciaranmcnulty, and thanks @nicolas-grekas for the tips

PR #1155, opened 2017-09-20
[RFC] Add strict types to class template

avant1:
Hey there.
Since PHPSpec 4 requires PHP 7 it makes sense to add strict types declaration to generated classes and interfaces by default.

But there are several questions that might be discussed:
1. Is it okay at all?
I think using strict types by default on PHP 7 for all new classes will allow to see errors early, and they will be marked explicitly. So developers will be happier.

2. Is it a BC break?
I don't think so. I can't imagine scenario where this change will break anything since it changes only newly generated files. But maybe you see such scenarios.

3. Where should `declare` block be placed?
I've seen two main formats by now. The first one places `declare` block on the next line after opening `<?php` tag:
```php
<?php
declare(strict_types=1);

namespace Acme;

class Foo {}
```

The second one puts `declare` block on the same line with opening `<?php` tag:
```php
<?php declare(strict_types=1);

namespace Acme;

class Foo {}
```

The first one seems more standard for me, PHP documentation [uses it in examples](http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.strict). The second one is a bit more compact.

There is also a variation of the first format:
```php
<?php

declare(strict_types=1);

namespace Acme;

class Foo {}
```

There is additional empty line between opening tag and `declare` block.

This PR uses first format.

What do you think about this?

### DIFF
--- a/features/code_generation/developer_generates_class.feature
+++ b/features/code_generation/developer_generates_class.feature
@@ -10,6 +10,7 @@ Feature: Developer generates a class
     Then a new class should be generated in the "src/CodeGeneration/ClassExample1/Markdown.php":
       """
       <?php
+      declare(strict_types=1);
 
       namespace CodeGeneration\ClassExample1;
 
@@ -60,6 +61,7 @@ Feature: Developer generates a class
     Then a new class should be generated in the "src/MyNamespace/Markdown.php":
     """
     <?php
+    declare(strict_types=1);
 
     namespace Behat\Tests\MyNamespace;
 
@@ -76,6 +78,7 @@ Feature: Developer generates a class
     Then a new class should be generated in the "src/CodeGeneration/ClassExample1/Text/Markdown.php":
       """
       <?php
+      declare(strict_types=1);
 
       namespace CodeGeneration\ClassExample1;
 
@@ -92,6 +95,7 @@ Feature: Developer generates a class
     Then a new class should be generated in the "src/CodeGeneration/Class_Example2/Text/Markdown.php":
       """
       <?php
+      declare(strict_types=1);
 
       namespace CodeGeneration\Class_Example2;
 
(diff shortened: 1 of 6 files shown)


ciaranmcnulty:
I really don't see the point of strict_types by default, but might be in a minority :-/

ciaranmcnulty:
I've been thinking about this; because this is an opt-in feature in PHP I don't see a need to push it on users when I don't particularly think it's a best practice.

Possibly the docs around custom templates could use this as an example though?

Happy to hear counter-arguments!

avant1:
I was thinking about following examples where strict types might be useful: https://3v4l.org/s3YGm

Without strict types all of those calls "work", and they may lead to unexpected results.

None of those examples will work with enabled strict types mode. You will get explicit error message, describing that you are doing something wrong (adding `false` to `4` does not make any sense).

From other point of view enabling strict types by default may confuse developers who are not aware of primitive types in PHP (cause passing wrong types to builtin functions will not work anymore). But this will nudge them to read about that new features in PHP.

This also touches other arguable question: is throwing error as soon as possible better than trying to keep working as long as possible (though that may lead to unexpected results)?

For now I think that explicit and early errors are better than wrong calculation results. And even exceptions on production are useful if you monitor them carefully.
But I agree this might not be best choice for all teams and projects.

So feel free to close this PR if no other opinions appear.

yvoyer:
Maybe it could be added using a configuration. It could be off by default unless the user explicitly enables it.

ciaranmcnulty:
The configuration would be 'use a custom template'

sgolemon:
Adding strict types to generated files doesn't really push anything on end users since THEIR calling scopes can still be weak.  Turning it on in the generated files just enforces that THOSE files are correct.

If you're not confident in your generated code being correct then that's a separate problem.

ciaranmcnulty:
@sgolemon yes I understand how the mechanism works. I'm not keen on 'pushing' including the optional non-default declaration by default.

ciaranmcnulty:
I don't want to make developers enable a non-default feature by default - will accept any docs changes to make the process for adding a custom template simpler, of course

Issue #1032, opened 2016-11-28
Add `shouldIterateLike` as a complement to `shouldIterateAs`


adamcameron:
As per discussion with @ciaranmcnulty on #bddLondon Slack channel:

 I ran into an issue with `shouldIterateAs`. Not sure if it's by design or not.
https://github.com/phpspec/phpspec/blob/master/src/PhpSpec/Matcher/Iterate/IterablesMatcher.php#L49

Here it's using `!==` to compare the value. Which means for objects it needs to be *exactly* the same object for it to match.

Which seems slightly unfeasible to me

ciaranmcnulty:
> Hm
> We have separate matchers for loose/strict comparison
> raise an issue?
> We could add `shouldIterateLike` perhaps

Raising issue as indicated.

PR #1158, opened 2017-10-13
Add an iterateLike matcher

sroze:
So we can have non-strict comparison for iterators.

### DIFF
--- a/src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
+++ b/src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
@@ -29,7 +29,7 @@ final class IterablesMatcher
      * @throws SubjectHasFewerElementsException
      * @throws SubjectHasMoreElementsException
      */
-    public function match($subject, $expected, bool $strict = false)
+    public function match($subject, $expected, bool $strict = true)
     {
         if (!$this->isIterable($subject)) {
             throw new \InvalidArgumentException('Subject value should be an array or implement \Traversable.');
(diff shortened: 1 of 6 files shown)


ciaranmcnulty:
Looks useful - would benefit from some tests + doc changes

ciaranmcnulty:
Also, see #1032

sroze:
@ciaranmcnulty anything special you'd like me to have a look to in this issue? 😃

ciaranmcnulty:
Just linking so I remember to close it :-)

Could you also add an equivalent of the shouldIterateStarting?

sroze:
`shouldIterateStartingLike` ?

src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
```diff
@@ -22,13 +22,14 @@ public function __construct(Presenter $presenter)
...
      * @param array|\Traversable $expected
+     * @param bool               $strict
      *
      * @throws \InvalidArgumentException
      * @throws SubjectElementDoesNotMatchException
      * @throws SubjectHasFewerElementsException
      * @throws SubjectHasMoreElementsException
      */
-    public function match($subject, $expected)
+    public function match($subject, $expected, bool $strict = false)
```
ciaranmcnulty: The old behaviour was to be strict, so this should default to `true`

ciaranmcnulty:
Thanks @sroze !

sroze:
Thank you for taking it over @ciaranmcnulty!

PR #1161, opened 2017-10-14
Avoid deprecated symfony/console method call in tests

### LINKED ISSUE #740: Behat test fails randomly
the scenario 'Generating a class outside of autoloadable paths gives a warning' in features/code_generation/developer_generates_class.feature fails randomly. 
I found the issue came from the terminal width. With some widths the word 'autoload' is split (see screenshot) causing the test to fail.
phpspec
I don't really know if the test should be rewritten of if it is a beat issue.
I'm running the tests on Mac OS X (last version) with iterm 2.

avant1:
That method will be removed when `symfony/console` 4 will be released.

Closes #740 (the issue is already fixed, but now fix is compatible with `symfony/console` v4).

> The actual output should wrap better IMO

@ciaranmcnulty, I'm not sure if I understand that phrase. If you meant something else I can check if it can be improved.

### DIFF
--- a/features/bootstrap/ApplicationContext.php
+++ b/features/bootstrap/ApplicationContext.php
@@ -49,12 +49,18 @@ class ApplicationContext implements Context
 
         $this->application = new Application('2.1-dev');
         $this->application->setAutoExit(false);
-        $this->application->setTerminalDimensions(130, 30);
+        $this->setFixedTerminalDimensions();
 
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
I meant rather than set the dimensions to something that doesn't trigger the bad wrapping, the wrapping should handle more dimensions correctly

PR #1163, opened 2017-10-20
[RFC] Refactor default matchers into an extension

ciaranmcnulty:
I want to start decomposing the *Assembler objects into internal extensions and this seemed like a good place to start.

The idea will be to take things like the prophecy integration, the symfony console, etc and end up with a stack of core extensions, similar to behat. 

An end goal may be to decouple into separate repos

### DIFF
--- a/src/PhpSpec/Extensions/DefaultMatchers/Extension.php
+++ b/src/PhpSpec/Extensions/DefaultMatchers/Extension.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace PhpSpec\Extensions\DefaultMatchers;
+
+use PhpSpec\Factory\ReflectionFactory;
+use PhpSpec\Matcher;
+use PhpSpec\ServiceContainer;
+use PhpSpec\ServiceContainer\IndexedServiceContainer;
+
+class Extension implements \PhpSpec\Extension
+{
+    /**
+     * @param ServiceContainer $container
+     * @param array $params
+     */
+    public function load(ServiceContainer $container, array $params)
+    {
+        $container->define('matchers.identity', function (IndexedServiceContainer $c) {
+            return new Matcher\IdentityMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.comparison', function (IndexedServiceContainer $c) {
+            return new Matcher\ComparisonMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.throwm', function (IndexedServiceContainer $c) {
+            return new Matcher\ThrowMatcher($c->get('unwrapper'), $c->get('formatter.presenter'), new ReflectionFactory());
+        }, ['matchers']);
+        $container->define('matchers.trigger', function (IndexedServiceContainer $c) {
+            return new Matcher\TriggerMatcher($c->get('unwrapper'));
+        }, ['matchers']);
+        $container->define('matchers.type', function (IndexedServiceContainer $c) {
+            return new Matcher\TypeMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.object_state', function (IndexedServiceContainer $c) {
+            return new Matcher\ObjectStateMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.scalar', function (IndexedServiceContainer $c) {
+            return new Matcher\ScalarMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.array_count', function (IndexedServiceContainer $c) {
+            return new Matcher\ArrayCountMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.array_key', function (IndexedServiceContainer $c) {
+            return new Matcher\ArrayKeyMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.array_key_with_value', function (IndexedServiceContainer $c) {
+            return new Matcher\ArrayKeyValueMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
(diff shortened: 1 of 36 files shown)


src/PhpSpec/Extensions/DefaultMatchers/Extension.php
```diff
@@ -0,0 +1,88 @@
...
+    public function load(ServiceContainer $container, array $params)
+    {
+        $container->define('matchers.identity', function (IndexedServiceContainer $c) {
+            return new Matcher\IdentityMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.comparison', function (IndexedServiceContainer $c) {
+            return new Matcher\ComparisonMatcher($c->get('formatter.presenter'));
+        }, ['matchers']);
+        $container->define('matchers.throwm', function (IndexedServiceContainer $c) {
+            return new Matcher\ThrowMatcher($c->get('unwrapper'), $c->get('formatter.presenter'), new ReflectionFactory());
```
avant1: Since `ReflectionFactory` parameter was removed from constructor it can be removed here too.

sroze:
Very nice way of moving classes while keeping the BC 👏 

> An end goal may be to decouple into separate repos

Read-only splits are literally other source repos? If it's the last one, I'm not sure it's a great target as it adds a huge management overhead.

src/PhpSpec/Factory/ReflectionFactory.php
```diff
@@ -16,6 +16,8 @@
 /**
  * Class ReflectionFactory is a simple factory wrapper to create reflection
  * classes
+ *
+ * @deprecated
```
sroze: Why is this deprecated? What else should I use instead? When will it be removed?

src/PhpSpec/Extensions/DefaultMatchers/Extension.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace PhpSpec\Extensions\DefaultMatchers;
+
+use PhpSpec\Matcher;
+use PhpSpec\ServiceContainer;
+use PhpSpec\ServiceContainer\IndexedServiceContainer;
+
+class Extension implements \PhpSpec\Extension
```
sroze: Maybe you could name the extension "DefaultMatchersExtension"? Just feels weird to have an implementation named the same as the contract.

src/PhpSpec/Extensions/DefaultMatchers/Extension.php
```diff
@@ -0,0 +1,87 @@
...
+{
+    /**
+     * @param ServiceContainer $container
```
sroze: It's style but hey... we don't need this PhpDoc anymore :))

src/PhpSpec/Extensions/DefaultMatchers/Extension.php
ciaranmcnulty: because then it'd be called `PhpSpec\Extension\DefaultMatchers\DefaultMatchersExtension` and I'd go crazy

src/PhpSpec/Factory/ReflectionFactory.php
ciaranmcnulty: It was not used for anything, not even the tests

sroze: I mean, you should just answer these questions in the `@deprecated` comment I think. That's how deprecation comments are done in Symfony and that's quite helpful.

src/PhpSpec/Extensions/DefaultMatchers/Extension.php
sroze: 🤷‍♂️

src/PhpSpec/Factory/ReflectionFactory.php
ciaranmcnulty: I guess I should start triggering warnings?

sroze: As well, yep!

src/PhpSpec/Extensions/DefaultMatchers/Extension.php
everzet: How about renaming it to `The`?

ciaranmcnulty: "Oh cool, @everzet commented on a PR. I bet it's something insightful"

MarcelloDuarte:
Put the `$deprecated_classes` inside the anonymous function, so it's not in the global scope. And while you are in it, also rename it to `$deprecatedClasses`.

src/deprecated.php
```diff
@@ -0,0 +1,41 @@
...
+        'PhpSpec\Matcher\TraversableCountMatcher' => 'PhpSpec\Extensions\DefaultMatchers\Matcher\TraversableCountMatcher',
+        'PhpSpec\Matcher\TraversableKeyMatcher' => 'PhpSpec\Extensions\DefaultMatchers\Matcher\TraversableKeyMatcher',
+        'PhpSpec\Matcher\TraversableKeyValueMatcher' => 'PhpSpec\Extensions\DefaultMatchers\Matcher\TraversableKeyValueMatcher',
+        'PhpSpec\Matcher\TriggerMatcher' => 'PhpSpec\Extensions\DefaultMatchers\Matcher\TriggerMatcher',
+        'PhpSpec\Matcher\TypeMatcher' => 'PhpSpec\Extensions\DefaultMatchers\Matcher\TypeMatcher',
+    ];
+    
+    if (array_key_exists($class, $deprecatedClasses)) {
+        $new = $deprecatedClasses[$class];
+        class_alias($new, $class);
```
jakzal: This is neat! :beer:

ciaranmcnulty: In my local branch I'm also triggering a E_USER_DEPRECATED in there... but then everything breaks because I need to find all the places we're using it internally ;)

PR #1166, opened 2017-10-23
Added Symfony 4 support

gnugat:
Good news everyone!

According to its tests, phpspec is compatible with Symfony 4!
Now the bad news is that phpspec depends on Behat which prevents `symfony/console` and `symfony/yaml` to be upgraded, but at least there's no effort to do in phpspec.

I'd be happy to wait for Symfony 4 stable to be released before merging this Pull request, what do you think?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,18 +26,18 @@
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
         "sebastian/exporter":       "^1.0 || ^2.0 || ^3.0",
-        "symfony/console":          "^3.2",
-        "symfony/event-dispatcher": "^3.2",
-        "symfony/process":          "^3.2",
-        "symfony/finder":           "^3.2",
-        "symfony/yaml":             "^3.2",
+        "symfony/console":          "^3.2 || ^4.0@beta",
+        "symfony/event-dispatcher": "^3.2 || ^4.0@beta",
+        "symfony/process":          "^3.2 || ^4.0@beta",
+        "symfony/finder":           "^3.2 || ^4.0@beta",
+        "symfony/yaml":             "^3.2 || ^4.0@beta",
         "doctrine/instantiator":    "^1.0.5",
         "ext-tokenizer":            "*"
     },
 
     "require-dev": {
         "behat/behat":           "^3.3",
-        "symfony/filesystem":    "^3.2",
+        "symfony/filesystem":    "^3.2 || ^4.0@beta",
         "phpunit/phpunit":       "^5.7|^6.0"
     },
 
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
We only depend on it in autoload-dev - the trick is to test phpspec in CI effectively..

ciaranmcnulty:
Let's wait until at least an RC, too

composer.json
```diff
@@ -26,18 +26,18 @@
         "phpspec/prophecy":         "^1.5",
         "phpspec/php-diff":         "^1.0.0",
         "sebastian/exporter":       "^1.0 || ^2.0 || ^3.0",
-        "symfony/console":          "^3.2",
-        "symfony/event-dispatcher": "^3.2",
-        "symfony/process":          "^3.2",
-        "symfony/finder":           "^3.2",
-        "symfony/yaml":             "^3.2",
+        "symfony/console":          "^3.2 || ^4.0@beta",
```
alcaeus: The `@beta` suffix should be dropped: the CI build will test it in the `DEPENDENCIES=dev` matrix build.

Nyholm: Correct. 

Could this PR me updated and merged? There are a ton of other projects that depends on this library that are waiting for this.

Nyholm:
Awesome, Behat just merged a PR to make them compatible with SF4.

ciaranmcnulty:
The phpspec suite is failing, which makes me think it is not actually symfony 4 compatible yet?

ciaranmcnulty:
Hmph no, it's one of the other dependencies that's failing I think

ciaranmcnulty:
If anyone can explain/resolve the build failure, I'll merge

alcaeus:
I'd have to guess, but could the [parent constructor in `SplFileInfo`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Finder/SplFileInfo.php#L31) have something to do with it?

gnugat:
I'm not sure why we get an error, since the tests are running with Symfony3...

alcaeus:
@gnugat not quite - where not blocked by other dependencies, the `dev` build installs `dev-master` which is 4.0: https://travis-ci.org/phpspec/phpspec/jobs/304433561#L501

gnugat:
@alcaeus oh you're right! Looks like the issue comes from Prophecy, I'll see what I can do there

alcaeus:
@gnugat Thanks!

gnugat:
Should be fixed once https://github.com/phpspec/prophecy/pull/373 has been merged.

ciaranmcnulty:
Still failures after building against prophecy dev-master after that patch is merged

sroze:
Proposed #1181 to replace this PR, adding a few fixes to PhpSpec for Symfony 4 support.

PR #1168, opened 2017-10-24
Update symfony components to allow 4.x version.

dbrumann:
In order to run the test suites (phpunit, phpspec & behat) locally I did the following changes to composer.json that were not commited:

```
    "repositories": [
        {
            "type": "package",
            "package": {
                "name": "behat/behat",
                "version": "3.3.99",
                "type": "path",
                "dist": {
                    "type": "path",
                    "url": "/path/to/local/behat"
                }
            }
        }
    ],
    "minimum-stability": "dev",
```

I changed the local behat to contain the 4.x versions as well, otherwise using `--no-dev` should work as well. The more important bit is `"minimum-stability": "dev"` to make sure the 4.x versions are used by composer. I assume this has to be reflected in .travis.yml as well, but I'm not sure what the best way would be as I assume adding the line to the composer.json is not desired.


ciaranmcnulty:
I'm sorry to report that @gnugat got there already with #1166

ciaranmcnulty:
Really appreciate the effort though!

PR #1188, opened 2018-01-04
Fix matching float values

mheki:
Currently `IdentityMatcher` for simple spec using float values like 
```
$this->subtract(1.2, 1)->shouldReturn(0.2);
```
 returns false instead of true for correct implementation. 

Float values comparison in PHP is a tricky thing. Matching by identity doesn't always return the result we expect. Example:
```php
$a = 0.3 - 0.1;
$b = 0.1 + 0.1;

var_dump($a === $b); // false
var_dump($a == $b); // false
```

To fix it I'd like to introduce `FloatValueMatcher` (I'm open to suggestions with better naming).
Current implementation fixes issue for IdentityMatcher, I would need to add also similar fix for ComparisonMatcher. Creating WIP PR to get some feedback.

Todo:
- [ ] update docs
- [ ] implement fix for ComparisonMatcher

### DIFF
--- a/src/PhpSpec/Matcher/FloatValueMatcher.php
+++ b/src/PhpSpec/Matcher/FloatValueMatcher.php
@@ -67,7 +67,7 @@ final class FloatValueMatcher extends BasicMatcher
     protected function matches($subject, array $arguments): bool
     {
         if (extension_loaded('bcmath')) {
-            return \bccomp($subject, $arguments[0], 20) === 0;
+            return \bccomp($subject, $arguments[0], 17) == 0;
         }
 
         return (string) $subject === (string) $arguments[0];
--- a/src/PhpSpec/Matcher/FloatValueMatcher.php
+++ b/src/PhpSpec/Matcher/FloatValueMatcher.php
@@ -1,0 +1,98 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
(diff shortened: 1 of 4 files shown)


src/PhpSpec/Matcher/FloatValueMatcher.php
```diff
@@ -0,0 +1,98 @@
...
+            && 1 === \count($arguments)
+            && \is_float($arguments[0]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority(): int
+    {
+        return 150;
```
mheki: this has to be higher than IdentityMatcher to execute this one first

src/PhpSpec/Matcher/FloatValueMatcher.php
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function matches($subject, array $arguments): bool
+    {
+        if (extension_loaded('bcmath')) {
```
mheki: bcmath extension is not loaded by default on linux distributions.

src/PhpSpec/Matcher/FloatValueMatcher.php
```diff
@@ -0,0 +1,98 @@
...
+            return \bccomp($subject, $arguments[0], 20) === 0;
```
mheki: I wasn't sure what value to use for the `scale` parameter. Whatever we use it will have to be documented.

ciaranmcnulty:
Did you try shouldBeApproximately()?

mheki:
Nope, @ciaranmcnulty you're right. I can get what I wanted with `shouldBeApproximately()`.
Cheers.

PR #1193, opened 2018-02-09
Allows the property $title of an ExampleNode to be altered.

drupol:
Hello,

I recently wrote my first [PHPSpec extension.](https://github.com/drupol/phpspec-annotation) It allows you to add an annotation to tests methods.

However, I also wanted to alter the title of existing methods but the $title property of an ExampleNode is something that you cannot alter.

This PR allows you to alter of an ExampleNode with a method.

To give you an idea of what I'm trying to achieve, I just updated the code of my extension (https://github.com/drupol/phpspec-annotation/blob/master/src/Listeners/PhpspecAnnotationListener.php#L38), if this PR is accepted, I'll remove the commented line and it will work directly.

Tests are included as well.

### DIFF
--- a/spec/PhpSpec/Loader/Node/ExampleNodeSpec.php
+++ b/spec/PhpSpec/Loader/Node/ExampleNodeSpec.php
@@ -22,7 +22,15 @@ class ExampleNodeSpec extends ObjectBehavior
         $this->getTitle()->shouldReturn('example node');
     }
 
-    function it_provides_a_link_to_function($function)
+    function it_can_set_a_title()
+    {
+      $this->setTitle('node example');
+      $this->getTitle()->shouldReturn('node example');
+      $this->setTitle('example node');
+    }
+
+
+  function it_provides_a_link_to_function($function)
     {
         $this->getFunctionReflection()->shouldReturn($function);
     }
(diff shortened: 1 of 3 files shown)


spec/PhpSpec/Loader/Node/ExampleNodeSpec.php
```diff
@@ -22,7 +22,15 @@ function it_provides_a_link_to_title()
         $this->getTitle()->shouldReturn('example node');
     }
 
-    function it_provides_a_link_to_function($function)
+    function it_can_set_a_title()
+    {
+      $this->setTitle('node example');
+      $this->getTitle()->shouldReturn('node example');
+      $this->setTitle('example node');
```
jakzal: Why are you setting the title again here, after an assertion?

jakzal:
Why do you need to alter the title?

ciaranmcnulty:
It's for an extension that lets people use annotations for naming examples, instead of it being inferred from the method name

jakzal:
I assume it's not possible for the extension to create the `ExampleNode` and pass the title via the constructor?

spec/PhpSpec/Loader/Node/ExampleNodeSpec.php
drupol: This is indeed not necessary, do you want me to remove it ?

drupol:
I guess it would be possible, but I haven't found a proper way to remove a ExampleNode from a SpecificationNode... unless to create a new one.

drupol:
Hi @jakzal , @ciaranmcnulty ,

I added an entry in the extensions files so it would show up on the site.
Should I do something else to push this thing forward ?

Thanks!

drupol:
Ping @jakzal & @ciaranmcnulty ?

drupol:
@ciaranmcnulty Should I do something to get this moving forward ?

Thanks!

drupol:
Oooh I just noticed you merged this !!!

Thanks !!!!

PR #1197, opened 2018-03-29
Add more @method phpdocs to ObjectBehavior

l3l0:
Added more @method phpdoc tags to helps IDEs autocompletion.
This is basically can helps when SuS is collection/iterator like object so we can match it using syntax like:

```php
$this->shouldIterateAs(['aa'], ['cc']);
```

which IDEs does not catch :)

See #946 and #951


ciaranmcnulty:
Great! thank you

PR #1199, opened 2018-04-04
Added check for mismatched spec path and defined namespace

aduss:
Currently in case there is mismatch between spec class and defined namespace tests are silently ignored. This PR solves it by reporting the case.

Solves #354

### DIFF
--- a/spec/PhpSpec/Loader/ResourceLoaderSpec.php
+++ b/spec/PhpSpec/Loader/ResourceLoaderSpec.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace spec\PhpSpec\Loader;
+
+use PhpSpec\Loader\Node\SpecificationNode;
+use PhpSpec\Locator\Resource;
+use PhpSpec\Locator\ResourceManager;
+use PhpSpec\ObjectBehavior;
+use PhpSpec\Util\MethodAnalyser;
+
+class ResourceLoaderSpec extends ObjectBehavior
+{
+
+    public function let(
+        ResourceManager $resourceManager,
+        MethodAnalyser $methodAnalyser
+    ) {
+        $this->beConstructedWith(
+            $resourceManager,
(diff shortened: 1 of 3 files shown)


src/PhpSpec/Loader/ResourceLoader.php
```diff
@@ -61,6 +61,8 @@ public function load(string $locator = '', int $line = null): Suite
                 }
             }
             if (!class_exists($resource->getSpecClassname(), false)) {
+                $error = new \Error('Spec path does not match defined namespace.');
```
ciaranmcnulty: I think this message is too specific; all we know here is that the spec was not found in the file - maybe we can report the expected FQCN and file path and be quite explicit?

Also not sure about `\Error` here - this can be a more specific exception

spec/PhpSpec/Loader/ResourceLoaderSpec.php
```diff
@@ -0,0 +1,60 @@
...
+        $expectedMessage = 'Spec path does not match defined namespace.';
+
+        $resource->getSrcClassname()->willReturn('src/classname');
+        $resource->getSrcFilename()->willReturn('src/filename.php');
+        $resource->getSpecClassname()->willReturn('spec/mismatched/classname');
+        $resource->getSpecFilename()->willReturn('spec/filename.php');
+
+        $resourceManager->locateResources('locator')->willReturn([$resource]);
+
+        $this->load('locator')->shouldMatchExampleClosureExceptionMessage($expectedMessage);
```
ciaranmcnulty: Can you name this something like shouldLocateSpecThatThrowsExceptionWithMessage?

ciaranmcnulty:
@aduss can you look at the travis failure please?

ciaranmcnulty:
@aduss I'm closing this because of inactivity, but please re-open if you get time to address the above

PR #1204, opened 2018-05-13
Rework appveyor

Taluu:
As done in behat. :}

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,3 +1,4 @@
+version: '{build}'
 build: false
 shallow_clone: true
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
It seems to be failing

Taluu:
Behat tests seems to fail, but then we would need to check if it run on the current appveyor config... Could you trigger a build from master or something ?

jakzal:
@Taluu for build fix see #1205

ciaranmcnulty:
Great! Thanks

PR #1205, opened 2018-05-14
Fix the appveyor build

jakzal:
* Include the test output in the exception message so a failing test could be understod.
* Update the message we look for to "could not be loaded". Output on windows is misaligned (see screenshot below).

image


ciaranmcnulty:
@jakzal travis doesn't like it

jakzal:
Now it's broken on travis... we should probably make sure terminal width is the same on both ;)

ciaranmcnulty:
Or have it so the wrapping doesn't break on different platforms

jakzal:
Both are happy now.

PR #1206, opened 2018-05-14
Add `void` typehints

kix:
This required some changes for the specs to pass, e.g. `->willReturn();` → `->should(function() {})`.
Also, the codestyle is not consistent: sometimes it has a space before the colon, sometimes it doesn't. I've added spaces where the typehints were added.

### DIFF
--- a/features/bootstrap/Fake/ReRunner.php
+++ b/features/bootstrap/Fake/ReRunner.php
@@ -16,7 +16,7 @@ class ReRunner implements BaseReRunner
         return true;
     }
 
-    public function reRunSuite()
+    public function reRunSuite() : void
     {
         $this->hasBeenReRun = true;
     }
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
 
     "require": {
         "php":                      "^7.1, <7.3",
-        "phpspec/prophecy":         "^1.5",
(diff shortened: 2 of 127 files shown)


ciaranmcnulty:
It looks like there are some CS fails, and the minimum Prophecy version will need to be bumped to a newer tag that supports these void returns.

ciaranmcnulty:
Maybe just bump it in dev? I'm not sure how we'd effectively test it against older versions though

kix:
@jakzal, agree. Updated the PR removing all whitespace before `:`.

ciaranmcnulty:
Great, thanks! The Appveyor config is fixed on master so I’ll merge across later

features/bootstrap/Fake/ReRunner.php
```diff
@@ -16,7 +16,7 @@ public function isSupported()
         return true;
     }
 
-    public function reRunSuite()
+    public function reRunSuite() : void
```
kix: Ouch, missed this one :(

ciaranmcnulty: I don’t really care about white space

jakzal: 😱

PR #1208, opened 2018-06-20
fixed tests for phpspec/prophecy#408

meridius:
fixed tests for https://github.com/phpspec/prophecy/pull/408


meridius:
The tests are failing because they rely on phpspec/prophecy#408 being merged to phpspec/prophecy:"^1.5" (in `composer.json`), but it is not yet.

ciaranmcnulty:
@meridius we will need the tests to pass with older versions of prophecy still

stof:
Closing in favor of #1224

PR #1209, opened 2018-07-16
Fix return type for ObjectStateMatcher::negativeMatch()

nightlinus:
Is there any case for returning ?DelayedCall, maybe just void would be right?

### DIFF
--- a/spec/PhpSpec/Matcher/ObjectStateMatcherSpec.php
+++ b/spec/PhpSpec/Matcher/ObjectStateMatcherSpec.php
@@ -1,11 +1,11 @@
 <?php
+declare(strict_types = 1);
 
 namespace spec\PhpSpec\Matcher;
 
+use PhpSpec\Formatter\Presenter\Presenter;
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Can you explain why this is an improvement?

The return type is dictated by the parent `Matcher` interface; some matchers return a `DelayedCall`

nightlinus:
Thank you for quick reply.
Ok, just asked about need of `DelayedCall`. In any case this pr fixes missing `return null;` or one got error
`[err:TypeError("Return value of PhpSpec\Matcher\ObjectStateMatcher::negativeMatch() must be an instance of PhpSpec\Wrapper\DelayedCall or null, none returned")]`

ciaranmcnulty:
Oh I see, I wonder why the tests didn't catch it at all?

nightlinus:
let me add one positive test

nightlinus:
Ah, all is valid behaviour when using non-phar global version.

ciaranmcnulty:
Thanks!

PR #1215, opened 2018-08-10
Reject all non-objects for method calls, rather than just null

### LINKED ISSUE #1194: Tell what method is not found in ExceptionFactory::methodNotFound()
In spec I had a typo:

```
$this->foo('bar')->shoulReturn('baz');
```

Typo is:

```diff
- shoulReturn
+ shouldReturn
```

And got error:

```
exception [err:TypeError("Argument 1 passed to PhpSpec\Exception\ExceptionFactory::methodNotFound() must be of the type string, null given, called in phpspec/phpspec/src/PhpSpec/Wrapper/Subject/Caller.php on line 363")] has been thrown.
```

It would be cool if some friendlier error message is thrown, at least to tell what method is not found.

ciaranmcnulty:
Should resolve #1194

### DIFF
--- a/src/PhpSpec/Wrapper/Subject/Caller.php
+++ b/src/PhpSpec/Wrapper/Subject/Caller.php
@@ -87,7 +87,7 @@ class Caller
      */
     public function call(string $method, array $arguments = array()): Subject
     {
-        if (null === $this->getWrappedObject()) {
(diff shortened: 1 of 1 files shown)


Sam-Burns:
I suggest rebasing the branch now the build has been fixed

PR #1216, opened 2018-08-10
fix composer dependencies for 2.5

lifeofguenter:
`^1.0.1` translates to `>=1.0.1, <2.0.0` - which is correct. But unfortunately the upstream-package did not version according to semver and introduced a breaking change in a minor version bump, meaning their version `1.1.0` now requires `php7.1+`, see:

* https://github.com/doctrine/instantiator/issues/46
* https://github.com/doctrine/instantiator/releases/tag/1.1.0

The changes below will make sure that only `>=1.05, <1.1.0` version will be installed.

Workaround until this is merged + released is to manually require the correct version in your project:

```
$ composer require "doctrine/instantiator" "~1.0.5"
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "symfony/process":          "^2.6|~3.0",
         "symfony/finder":           "~2.1|~3.0",
         "symfony/yaml":             "~2.1|~3.0",
-        "doctrine/instantiator":    "^1.0.1",
+        "doctrine/instantiator":    "~1.0.5",
         "ext-tokenizer":            "*"
     },
 


ciaranmcnulty:
Why does the PHP version bump count as a BC break? What issue is it causing for you?

lifeofguenter:
@ciaranmcnulty version `1.1.0` requires php7.1+ - which I do not have installed. phpspec 2.5 only requires php5, meaning there is a version mismatch

ciaranmcnulty:
Yes but if you run composer update you will get the appropriate version, won’t you?

lifeofguenter:
Yes, but why would you want to specify a wrong version range, especially if it is known that versions >=1.1.0 are incompatible?

But all good, if you don't want to change :)

ciaranmcnulty:
I think we can agree to disagree :-)

GrahamCampbell:
Bumping minimum package or php requirements does not constitute a breaking change, since composer will just resolve the version that you need for your environment.

PR #1220, opened 2018-08-20
Fix developer_is_shown_diffs.feature to make it pass with prophecy 1.8.0

nightlinus:


nightlinus:
I can imagine we can support both version with «almost see» where we remove all indentations and new lines.

nightlinus:
Can you look at solution?

nightlinus:
btw, can we remove tags in appveyor ~@php-version,@php5.4&&~@hhvm since package is ^7.1

ciaranmcnulty:
Thanks for this 👍

PR #1221, opened 2018-08-21
Update the documentation version

jakzal:
The version shown in [docs](http://www.phpspec.net/en/stable/cookbook/matchers.html) and google is wrong:

image


Sam-Burns:
@jakzal we fixed the tests, if you rebase your branch it will probably pass the build

PR #1224, opened 2018-09-19
Fix Prophecy error message bug

Sam-Burns:
Prophecy 1.8 rephrases an error message. This makes two tests fail in all branches of Behat. This PR makes the tests less closely coupled to the wording of the expected error message.

### DIFF
--- a/features/bootstrap/ApplicationContext.php
+++ b/features/bootstrap/ApplicationContext.php
@@ -193,6 +193,24 @@ class ApplicationContext implements Context
     }
 
     /**
+     * @Then I should see the error that :methodCall was not expected on :class
+     */
+    public function iShouldSeeTheErrorThatWasNotExpectedOn($methodCall, $class)
+    {
+        $this->checkApplicationOutput((string)$methodCall);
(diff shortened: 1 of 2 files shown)


DonCallisto:
Did you get rid of

> expected calls were: [...]

part or did you forget it?

Sam-Burns:
@DonCallisto yeah, that's a good point. Do you think it would be better if I added that?

DonCallisto:
Yes, you have a valid POV and I agree.
I was just asking if this "forgetfulness" was intentional or not

Sam-Burns:
Happy to change the test in any way people like, if it will get the PR out the door. Quite keep to make the tests pass on `master` for the benefit of other contributors though.

ciaranmcnulty:
The changes look ok; can you target the `4.3` branch please (as it allows recent Prophecy)

Sam-Burns:
@ciaranmcnulty PR retargeted

ciaranmcnulty:
Thanks @Sam-Burns !

Issue #1202, opened 2018-04-19
setting verbosity in  configuration file


sp-daniel-pinyol:
Hi,
when I run phpspec from phpstorm, I didn't find a way to configure verbosity in order see eg. mismatched strings in an assertEquals.
Is there a way to do it in the configuration file? how?

thanks

jakzal:
Not sure what you mean. PhpSpec doesn't have assertEquals.

It is possible to pass the `--verbose` flag in the command line (or `-v`, `-vv`, `-vvv`) to get more output. You can configure this in phpstorm's test runner options for phpspec.

Available PhpSpec config options are documented here: http://www.phpspec.net/en/stable/cookbook/configuration.html

ciaranmcnulty:
There isn't currently a way to do it via the config file

DonCallisto:
If that's still an interesting feature, I can work on it as I've checked and it doesn't seem to be too complex to implement a solution based on config and console override (if no verbosity setted, take the one from config, otherwise take the level specified in console)

DonCallisto:
This can be closed as we provided it an implementation for this use case.

gquemener:
This issue could indeed be closed :)

ping @ciaranmcnulty

PR #1226, opened 2018-10-07
Added verbose option in configuration file

DonCallisto:
Issue/Feature #1202 

Let's discuss it if something is not clear.

### DIFF
--- a/docs/cookbook/configuration.rst
+++ b/docs/cookbook/configuration.rst
@@ -83,8 +83,8 @@ describing. In phpspec, you can group specification files by a certain namespace
   specifications. The complete namespace for specifications is
   ``%spec_prefix%\%namespace%``;
 * ``src_path`` [**default**: ``src``] - The path to store the generated
-  classes. By default paths are relative to the location where **phpspec** was 
-  invoked. **phpspec** creates the directories if they do not exist. This does 
+  classes. By default paths are relative to the location where **phpspec** was
+  invoked. **phpspec** creates the directories if they do not exist. This does
   not include the namespace directories;
 * ``spec_path`` [**default**: ``.``] - The path of the specifications. This
   does not include the spec prefix or namespace.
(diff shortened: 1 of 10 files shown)


docs/cookbook/configuration.rst
```diff
@@ -201,6 +201,17 @@ You can also set your tests to stop on failure by setting ``stop_on_failure``:
 
     stop_on_failure: true
 
+Moreover you can turn on verbosity by setting ``verbose``:
+
+.. code-block:: yaml
+
+    verbose: true
```
ciaranmcnulty: I'd like to reserve the option to use the higher verbosity levels for something in future rather than just a boolean

features/options/developer_chooses_verbose_output.feature
```diff
@@ -0,0 +1,139 @@
...
+      }
+
+      """
+    And the class file "src/Verbose/SpecExample1/ConfigVerbosityConsoleNotSetted.php" contains:
+      """
+      <?php
+
+      namespace Verbose\SpecExample1;
+
+      class ConfigVerbosityConsoleNotSetted
```
ciaranmcnulty: `-NotSet` please

features/options/developer_chooses_verbose_output.feature
```diff
@@ -0,0 +1,139 @@
...
+      {
+          public function getValue()
+          {
+              return [0];
+          }
+      }
+
+      """
+    When I run phpspec
+    Then Output should contains:
```
ciaranmcnulty: `The the output should contain:` please

docs/cookbook/configuration.rst
DonCallisto: So you mean that accepted value should be `vvv` (or `3` that's debug value)? Is not counterintuitive? I mean, if I use `vvv` or whatever I would expect that also other levels are handled so I would try with `v` and `vv`.
WDYT?

DonCallisto:
@ciaranmcnulty I've fix all you asked for in a separate commit, in order to let the review be easier. 
Just one thing: CI is broken. I've tried onto my laptop and if I run single behat test (so if I specify feature file) test will pass, otherwise it won't. I really don't know if, maybe?, behat w/o single feature file runs with any command options (like verbosity?).

DonCallisto:
@ciaranmcnulty your intuition was good. Take a look to last commit, it should fix it the right way.
BTW we still have an unresolved conversation above, please check it out, thanks!

docs/cookbook/configuration.rst
ciaranmcnulty: Yeah boolean's probably fine for now then

features/bootstrap/ApplicationContext.php
```diff
@@ -76,6 +78,11 @@ private function setupReRunner()
         $this->application->getContainer()->set('process.rerunner.platformspecific', $this->reRunner);
     }
 
+    private function resetShellVerbosity()
+    {
+        putenv(sprintf('SHELL_VERBOSITY=%d', OutputInterface::VERBOSITY_NORMAL));
```
ciaranmcnulty: Is there not a way to reset the ApplicationTester instead?

DonCallisto:
@ciaranmcnulty travis is green :tada:

PR #1229, opened 2018-10-29
Script to run to check release is ok

ciaranmcnulty:
Script to run before pushing pre-release updates

### DIFF
--- a/check-release.php
+++ b/check-release.php
@@ -32,14 +32,14 @@ echo "👍 composer.json contains branch alias {$matches[0]}\n";
 
 $changelog = file_get_contents('CHANGES.md');
 
-if(!preg_match("/## \[$major.$minor.$patch\]/", $changelog, $matches)) {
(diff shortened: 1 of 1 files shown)


check-release.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+/**
+ * Sanity-checks a release for consistency
+ */
+
+
+
+// get the version reported by phpspec --version
+if (!preg_match_all('/(?<major>[5-9])\.(?<minor>[0-9]+)\.(?<patch>[0-9]+)/', file_get_contents('bin/phpspec'), $matches) || count($matches[0])!=1) {
```
stof: why `preg_match_all` rather than `preg_match` when you care only about a single match ?

ciaranmcnulty: So that it's deliberately brittle if a similar-looking string appears in that file at some point in the future which isn't the version number

PR #1231, opened 2018-10-29
Assume people are using PSR-4

### LINKED ISSUE #1230: Update documentation of installation
When I go on this page: http://www.phpspec.net/en/stable/manual/installation.html and I read psr-0, I want to cry.

greg0ire:
PSR-0 is deprecated (see https://www.php-fig.org/psr/psr-0), and as a
result PSR-4 is more likely to be seen in a beginner's composer.json
file, so let us not confuse them with PSR-0.
Fixes #1230

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,7 +12,7 @@ Create a ``composer.json`` file:
         "config": {
             "bin-dir": "bin"
         },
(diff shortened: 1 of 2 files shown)


greg0ire:
Argh that's correct, I forgot it wasn't just about the spec autoloading.

PR #1235, opened 2018-11-20
Re-enable PHAR builds

ciaranmcnulty:
This silently turned off when we dropped 7.0 support and I only recently noticed


jakzal:
👍

Issue #1210, opened 2018-07-21
Use statements are not sorted alphabetically


unfunco:
Having been working on a project beginning with the letter T for a year now, and manually editing `use` statements to be alphabetically ordered each time a specification is generated (so that they're consistent with how the IDE inserts them) I wrote [this extension](https://github.com/unfunco/phpspec-sort-use-statements). Would this be a feature suitable for version 5.x if refactored, or is it better off remaining an extension?

jdreesen:
Your link misses an "s" at the end. The correct link is: https://github.com/unfunco/phpspec-sort-use-statements

unfunco:
Whoops! Updated, thanks.

ciaranmcnulty:
This is for whether the subject namespace is before or after the PhpSpec/Prophecy ones?

unfunco:
It is, it could be done with a custom template per project too, but that would assume one root namespace.

ciaranmcnulty:
It makes sense to me in core then - we’d just need to document the existing template somewhere in the docs for people who want more control.

Please feel free to send a PR :-)

DonCallisto:
Can't we simply use [php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and integrate it with CI pipeline?

unfunco:
Closing this issue as it's now merged.

PR #1238, opened 2018-12-17
Sort generated use statements alphabetically

unfunco:
This implements #1210 
If this is okay I will update the documentation accordingly.


ciaranmcnulty:
This looks good, thanks!

Docs would be welcome

ciaranmcnulty:
Thanks @unfunco

PR #1241, opened 2019-02-25
Removed `Prophecy\Argument;` from templates and everywhere it wasn't used

DonCallisto:
`Prophecy\Argument;` is no longer used in this template so we can safely remove it from `use` import

### DIFF
--- a/docs/manual/getting-started.rst
+++ b/docs/manual/getting-started.rst
@@ -34,7 +34,6 @@ navigate to the spec folder and see the spec there:
 
     use Markdown;
     use PhpSpec\ObjectBehavior;
-    use Prophecy\Argument;
 
     class MarkdownSpec extends ObjectBehavior
     {
--- a/features/code_generation/developer_generates_class.feature
+++ b/features/code_generation/developer_generates_class.feature
@@ -111,7 +111,6 @@ Feature: Developer generates a class
     use CodeGeneration\MethodExample2\UserRepository;
     use CodeGeneration\MethodExample2\User;
     use PhpSpec\ObjectBehavior;
-    use Prophecy\Argument;
 
     class ForgotPasswordSpec extends ObjectBehavior
     {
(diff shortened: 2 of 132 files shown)


DonCallisto:
@ciaranmcnulty if it LGTY, you can directly squash the commit via GitHub.
Thanks.

DonCallisto:
ps.: take a look at https://github.com/phpspec/phpspec/blob/53e364e64a789581b127641f055b27ec90d2f1a3/features/matchers/developer_uses_identity_matcher.feature

This works but is not coherent with this modification. I suppose I should fix this everywhere in order to make it coherent.
Let me know what you think.

ciaranmcnulty:
The `use` is there so that users don't need to add it themselves when they add an Argument matcher to a spec

I'm on the fence about removing it - I guess many more people are using IDEs that'll autocomplete than there used to be? What do you think?

DonCallisto:
Umh, I think that's not a "good enough reason".
Try to think about this as any other "dependency": you can't prevent user to make mistaskes (eg.: forget about a `use`, import a different dependency but with the same name, ecc) speaking about *their* code. 
There are times where I never use the `Argument` so see that `use` there is very confusing to me.
Who develop in PHP should know about `use` keyword, regardless IDEs.

ciaranmcnulty:
Was mentioning it so you knew what the reason was :-)

Yeah I think we can remove this now tooling has moved on, if you want to fix the conflicts and make sure build is passing

DonCallisto:
Yeah, I understand the reason why you explained it: didn't want to be "rude". If you perceived my comment that way, I'm really sorry. It was not my intention, I was just replying, maybe "too fast" to have the right perceptions of my words.
BTW i'll resolve the conflict ;)

DonCallisto:
@ciaranmcnulty fixed and all tests green.
This comment https://github.com/phpspec/phpspec/pull/1241#issuecomment-467216458 should be addressed btw.
WDYT about?

ciaranmcnulty:
I think we should remove it in all features to be consistent (and make them shorter)

DonCallisto:
@ciaranmcnulty it should be done now

ciaranmcnulty:
I'm going to merge this into the next major (which won't be long)

ciaranmcnulty:
(bc-break label is just to remind me to do it in the next major, it's not a real bc break)

PR #1253, opened 2019-06-21
always fire `afterSpecification`

chris-kruining:
this will allow formatters to add 'closing' output


ciaranmcnulty:
It makes sense, my only concern really is whether it's a breaking change - I wonder whether extensions are assuming that afterSpecification means all is well?

ciaranmcnulty:
(No problem with this in the next major though)

PR #1254, opened 2019-07-23
Quick Start snippet updated to latest release

pfazzi:
I was installing phpspec on a new project following the Quick Start procedure and i end up installing an old version so I updated the docs/index.rst updating the lib versione from 4.0 to 5.1


MarcelloDuarte:
Thanks, all! 🍻

PR #1255, opened 2019-08-01
Avoid segmentation fault when calling getEventsCount

lombartec:
This avoids segmentation fault caused by `DotFormatter` when reaching a certain amount of collected events, instead of doing an `array_merge` over all event arrays and doing a count, this change sums the count of all those events separately.

After this change looks like `getAllEvents` would only be used by the `StatisticsCollectorSpec` so I don't know if it would be safe just to delete it or we should deprecate it if some external libs or extensions are still using it.

Before:

image

After:

image

### DIFF
--- a/src/PhpSpec/Listener/StatisticsCollector.php
+++ b/src/PhpSpec/Listener/StatisticsCollector.php
@@ -132,7 +132,7 @@ class StatisticsCollector implements EventSubscriberInterface
 
     public function getEventsCount() : int
     {
-        return \count($this->getAllEvents());
+        return array_sum($this->getCountsHash());
     }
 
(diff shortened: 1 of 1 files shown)


roukmoute:
There is no tests for this Collector ?

sp-carlos-lombarte:
Hi @roukmoute, I noticed there were a couple of specs already covering the modified method and thought they were good enough, I could add more if really needed

lombartec:
Hi @roukmoute, I noticed there were a couple of specs already covering the modified method and thought they were good enough, I could add more if really needed

ciaranmcnulty:
Thanks @lombartec, looks good

PR #1256, opened 2019-08-04
Update install instructions

driesvints:
The current instructions detail how to install Composer which is already explained right before it. It seems this is just an oversight as this line should explain how PhpSpec itself is installed through Composer. I also updated the Composer binary to the global one as that's the default when installing Composer.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,7 +21,7 @@ Install **phpspec** with composer:
 
 .. code-block:: bash
 
-    php composer.phar install
+    composer require --dev phpspec/phpspec
 
 Start writing specs:
 
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,7 +21,7 @@ Install **phpspec** with composer:
 
 .. code-block:: bash
 
-    composer require --dev phpspec/phpspec
+    composer install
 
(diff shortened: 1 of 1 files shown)


drupol:
I think this could be improved, it's a bit misleading according to me.

driesvints:
@drupol what do you mean? It's a simple composer require.

ciaranmcnulty:
In the context of the page we're saying 'create composer.json, then run install'

If you want to add a 'require' step the other content needs to be revised too (and include a 'set up the autoloader' step)

Happy to have this revised but if this line changes the way you suggest the flow of the steps doesn't make sense any more

driesvints:
Oh my god, I totally missed the `composer.json` above 🤦‍♂ Sorry about that. I'll update this pr to a simple `composer install`.

driesvints:
FWIW: I think that this could just be simpler with a `composer require`. I wouldn't document setting up the autoloading as that's pretty much standard with most frameworks these days. I've not seen too many libraries document that as well. Belongs more to the composer docs imo. A `composer require` will also create a `composer.json` for you.

ciaranmcnulty:
Thanks @driesvints

PR #1258, opened 2019-08-09
Implement stream_set_option in StreamWrapper to avoid warning with PHP 7.4

ciaranmcnulty:
This is now called by require_once since php 7.4

AFAIK is the only phpspec blocker on PHP 7.4 but there is a Prophecy issue that needs to be addressed.

### DIFF
--- a/src/PhpSpec/Loader/StreamWrapper.php
+++ b/src/PhpSpec/Loader/StreamWrapper.php
@@ -91,6 +91,6 @@ class StreamWrapper
 
     public function stream_set_option ( int $option , int $arg1 , int $arg2 ) : bool
     {
-        return true;
+        return false;
     }
 }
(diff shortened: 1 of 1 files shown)


drupol:
No phpdoc ?

ciaranmcnulty:
Not sure it adds anything TBH

src/PhpSpec/Loader/StreamWrapper.php
```diff
@@ -88,4 +88,9 @@ public function stream_eof(): bool
     {
         return feof($this->fileResource);
     }
+
+    public function stream_set_option ( int $option , int $arg1 , int $arg2 ) : bool
```
jdreesen:
```suggestion
    public function stream_set_option(int $option, int $arg1, int $arg2): bool
```

drupol: Oh I missed that one. I definitely agree.

stof:
shouldn't it set the option on the fileResource instead ?

ciaranmcnulty:
@stof not sure how - I've changed to `return false` because the upgrading notes seem to reflect that that's ok behaviour https://github.com/php/php-src/commit/8a4171ac457c048e1b05b2d9fc9b74f4816272a0

PR #1259, opened 2019-09-03
add official JSON formatter

chris-kruining:
I have used this formatter(https://github.com/chris-kruining/phpspec-json) as a personal extension for a while now and I think it could be beneficial to others.

antoher argument would be that the help lists json as a format, yet there is none

this implementation does require https://github.com/phpspec/phpspec/pull/1253 to work!

### DIFF
--- a/src/PhpSpec/Formatter/JsonFormatter.php
+++ b/src/PhpSpec/Formatter/JsonFormatter.php
@@ -73,7 +73,7 @@ final class JsonFormatter extends BasicFormatter
 
     public function afterSuite(SuiteEvent $event)
     {
-        $this->data['satus'] = self::STATUS_NAME[$event->getResult()];
+        $this->data['status'] = self::STATUS_NAME[$event->getResult()];
         $this->data['time'] = $event->getTime();
 
         $this->getIO()->write(json_encode($this->data));
--- a/src/PhpSpec/Formatter/JsonFormatter.php
+++ b/src/PhpSpec/Formatter/JsonFormatter.php
@@ -18,7 +18,11 @@ use PhpSpec\Event\SuiteEvent;
 
 final class JsonFormatter extends BasicFormatter
 {
-    private $data = [];
+    private $data = [
+        'status' => '',
+        'time' => 0,
+        'specifications' => [],
+    ];
 
     private const STATUS_NAME = [
         ExampleEvent::PASSED  => 'passed',
@@ -30,7 +34,11 @@ final class JsonFormatter extends BasicFormatter
 
     public function beforeSpecification(SpecificationEvent $event)
     {
-        $this->data[$event->getSpecification()->getTitle()] = [];
+        $this->data['specifications'][$event->getSpecification()->getTitle()] = [
+            'status' => '',
+            'time' => 0,
+            'examples' => [],
+        ];
     }
 
     public function afterExample(ExampleEvent $event)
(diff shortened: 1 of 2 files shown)


chris-kruining:
I have no clue how to fix the build, `\PhpSpec\Formatter\Presenter\Differ\ArrayEngine::convertArrayToString` does not contain any `str_repeat` calls

src/PhpSpec/Formatter/JsonFormatter.php
```diff
@@ -0,0 +1,77 @@
...
+ * design by specification.
+ *
+ * (c) Chris Kruining <chrise@keruining.eu>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+ 
+namespace PhpSpec\Formatter
+{
```
stof: don't use the curly braces for the namespace. This forces to indent the whole file for nothing.

src/PhpSpec/Formatter/JsonFormatter.php
```diff
@@ -0,0 +1,77 @@
...
+            $this->data[$event->getSpecification()->getTitle()] = [];
+        }
+
+        public function afterExample(ExampleEvent $event)
+        {
+            $specification = $event->getSpecification()->getTitle();
+            $example = $event->getTitle();
+
+            $this->data[$specification][$example] = [
+                'status' => static::STATUS_NAME[$event->getResult()],
```
stof: use `self::` to access private constants. Late static binding to access private things is not working (well, even though you cannot trigger the bug as inheritance is forbidden, but then *any* late static binding is useless anyway and only moves work from the compile time to the runtime)

src/PhpSpec/Formatter/JsonFormatter.php
```diff
@@ -0,0 +1,77 @@
...
+                'time' => $event->getTime(),
+            ];
+
+            $exception = $event->getException();
+
+            if($exception === null)
+            {
```
stof: please respect the PSR-2 coding standards used by the project

src/PhpSpec/Formatter/JsonFormatter.php
```diff
@@ -0,0 +1,75 @@
...
+ * file that was distributed with this source code.
+ */
+ 
+namespace PhpSpec\Formatter;
+
+use PhpSpec\Event\ExampleEvent;
+use PhpSpec\Event\SpecificationEvent;
+use PhpSpec\Event\SuiteEvent;
+
+final class JsonFormatter extends ConsoleFormatter
```
stof: this does not seem to use any of the helper of ConsoleFormatter (except the fact that it has a reference to the IO). So should it actually extend it ?

chris-kruining:
@stof was it actually my code or was master already failing to build?

chris-kruining:
@stof Any idea what I did wrong to make the build fail?

stof:
I have no idea about the failure. /cc @ciaranmcnulty

ciaranmcnulty:
I triggered a rebuild on `master` and we got the same issue

chris-kruining:
@ciaranmcnulty good to merge than?

ciaranmcnulty:
see https://github.com/sebastianbergmann/exporter/pull/29

Will see when that's merged

chris-kruining:
🤣 that was quick

ciaranmcnulty:
This looks good to me.

It means we have to maintain BC around events, I guess...

ciaranmcnulty:
@stof are you happy with the changes?

ciaranmcnulty:
Looks like the comments were addressed, good job @chris-kruining

PR #1263, opened 2019-10-01
Throw better exception when constructor returns null

ddziaduch:
Use the same logic and exceptions when creating objects from callables

### DIFF
--- a/spec/PhpSpec/Factory/ObjectFactorySpec.php
+++ b/spec/PhpSpec/Factory/ObjectFactorySpec.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of PhpSpec, A php toolset to drive emergent
+ * design by specification.
+ *
+ * (c) Marcello Duarte <marcello.duarte@gmail.com>
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace spec\PhpSpec\Factory;
+
+use PhpSpec\Exception\Fracture\FactoryDoesNotReturnObjectException;
+use PhpSpec\ObjectBehavior;
+
+class ObjectFactorySpec extends ObjectBehavior
+{
+    function it_throws_exception_when_factory_method_returns_non_object()
+    {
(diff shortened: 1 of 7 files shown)


src/PhpSpec/Factory/ObjectFactory.php
```diff
@@ -0,0 +1,49 @@
...
+ * file that was distributed with this source code.
+ */
+
+declare(strict_types=1);
+
+namespace PhpSpec\Factory;
+
+use PhpSpec\Exception\Fracture\FactoryDoesNotReturnObjectException;
+
+class ObjectFactory
```
ciaranmcnulty: Can we make this an instance rather than static calls? Because the callable isn't guaranteed pure it may have side effects

spec/PhpSpec/Factory/ObjectFactorySpec.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+declare(strict_types=1);
+
+namespace spec\PhpSpec\Factory;
+
+use PhpSpec\Exception\Fracture\FactoryDoesNotReturnObjectException;
+use PhpSpec\ObjectBehavior;
+
+class ObjectFactorySpec extends ObjectBehavior
```
ciaranmcnulty: Can you write some examples of successful instantiation here please

src/PhpSpec/Factory/ObjectFactory.php
```diff
@@ -0,0 +1,49 @@
...
+                'The %s did not return an object, returned %s instead',
+                self::callableToString($callable),
+                \gettype($instance)
+            ));
+        }
+
+        return $instance;
+    }
+
+    private static function callableToString(callable $callable): string
```
ciaranmcnulty: As this takes callable it needs to handle

a) arrays with `[0]` as an instance
b) Closure objects

ciaranmcnulty:
This looks great, thanks

I'm just working out when to release PhpSpec 6, and whether this should be saved for then or done in 5.x

ciaranmcnulty:
Actually this is a bugfix so should go to 5.x

ciaranmcnulty:
Thanks @ddziaduch, cherry-picked into 5.1

ddziaduch:
No problem @ciaranmcnulty 🙂

dantleech:
Thanks :)

Issue #1266, opened 2019-10-03
Handle deprecation warnings differently


ciaranmcnulty:
Major projects like Symfony, and PHP itself, are emitting deprecation warnings more regularly

Currently these are fails but we could potentially handle them differently by either:

- Passing the example but listing deprecations differently
- Having a new test result that is reported in a new colour, but counts as a pass
- Suppressing them entirely and just counting as a pass

ddziaduch:
The same thing was mentioned in my PR: https://github.com/phpspec/phpspec/pull/1269

stof:
What the Symfony PHPUnit bridge is doing for PHPUnit is that:

- in tests marked as `@group legacy`, deprecations are OK (and can be asserted btw). Tests are reported as passing and we only report a total number of legacy deprecations in the testsuite for info.
- in other tests, a config setting allows to decide whether deprecations should fail the test or no (to allow enforcing migrating away from deprecations) and get reported

PR #1269, opened 2019-10-15
#1268 PHP 7.4 support

### LINKED ISSUE #1268: PHP 7.4 support
When do you plan to support PHP 7.4?
After GA or earlier?
I'd like to know when to expect it, so I can start using PHP 7.4 as soon as possible.

Thanks for answer

ddziaduch:
This fixes #1268

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
 
     "require": {
         "php":                      "^7.2, <7.4",
-        "phpspec/prophecy":         "^1.7",
+        "phpspec/prophecy":         "^1.9",
         "phpspec/php-diff":         "^1.0.0",
         "sebastian/exporter":       "^1.0 || ^2.0 || ^3.0",
         "symfony/console":          "^3.4 || ^4.0",
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     ],
 
     "require": {
-        "php":                      "^7.2",
+        "php":                      "^7.2, <7.5",
         "phpspec/prophecy":         "^1.9",
         "phpspec/php-diff":         "^1.0.0",
         "sebastian/exporter":       "^1.0 || ^2.0 || ^3.0",
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     ],
 
     "require": {
-        "php":                      "^7.2, <7.4",
+        "php":                      "^7.2",
         "phpspec/prophecy":         "^1.9",
         "phpspec/php-diff":         "^1.0.0",
         "sebastian/exporter":       "^1.0 || ^2.0 || ^3.0",
--- a/src/PhpSpec/Matcher/ScalarMatcher.php
+++ b/src/PhpSpec/Matcher/ScalarMatcher.php
@@ -127,6 +127,10 @@ final class ScalarMatcher implements Matcher
             return false;
         }
 
+        if (PHP_VERSION_ID >= 70400 && $name === 'real') {
+            return false;
+        }
+
         $expected = lcfirst(substr($name, 2));
         if ($expected == 'boolean') {
             return 'is_bool';
--- a/src/PhpSpec/Matcher/ScalarMatcher.php
+++ b/src/PhpSpec/Matcher/ScalarMatcher.php
@@ -127,8 +127,8 @@ final class ScalarMatcher implements Matcher
             return false;
         }
 
-        if (PHP_VERSION_ID >= 70400 && $name === 'real') {
-            return false;
+        if ($name == 'beReal') {
+            trigger_error('`beReal` matcher is deprecated, please use `beFloat`', E_USER_DEPRECATED);
         }
 
         $expected = lcfirst(substr($name, 2));
--- a/src/PhpSpec/Matcher/ScalarMatcher.php
+++ b/src/PhpSpec/Matcher/ScalarMatcher.php
@@ -127,13 +127,13 @@ final class ScalarMatcher implements Matcher
             return false;
         }
 
-        if ($name == 'beReal') {
-            trigger_error('`beReal` matcher is deprecated, please use `beFloat`', E_USER_DEPRECATED);
-        }
-
         $expected = lcfirst(substr($name, 2));
         if ($expected == 'boolean') {
             return 'is_bool';
+        }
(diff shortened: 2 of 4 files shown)


src/PhpSpec/Matcher/ScalarMatcher.php
```diff
@@ -127,6 +127,10 @@ private function getCheckerName(string $name)
             return false;
         }
 
+        if (PHP_VERSION_ID >= 70400 && $name === 'real') {
```
stof: This is a BC break, as specs will suddenly stop passing. I suggest we trigger a deprecation notice instead (we can even do it for all PHP versions) and switch to `is_float` internally (as that's what the alias is internally).

stof:
To be sure that we support PHP 7.4, we would also need to add some Behat scenarios (running only on PHP 7.4+) that would use phpspec with classes using typed properties (we should check it for both the specced class and the collaborator class).

drupol:
Reading this thread makes me think to an extension I've been working on lately.

https://github.com/friends-of-phpspec/phpspec-requires

It let user `requires` a specific PHP version for tests.

Feel free to give some feedback.

stof:
I don't think using third-party phpspec extensions in the suite of phpspec itself is a wise idea (using a tool to test itself is already enough to create weird case sometimes, but relying on third-party extensions becomes really complex to manage, as the extension will never react to PHPSpec changes before they are done...)

stof:
btw, if my suggestion is followed, most of the checks will become unnecessary.

drupol:
> I don't think using third-party phpspec extensions in the suite of phpspec itself is a wise idea (using a tool to test itself is already enough to create weird case sometimes, but relying on third-party extensions becomes really complex to manage, as the extension will never react to PHPSpec changes before they are done...)

I definitely agree and suggesting to use an extension was not my intention.

src/PhpSpec/Matcher/ScalarMatcher.php
ddziaduch: Sounds like a great idea !

ddziaduch: I have just added `trigger_error()` with the deprecation but it is catches by PhpSpec so the specs will still fail

src/PhpSpec/Matcher/ScalarMatcher.php
```diff
@@ -127,6 +127,10 @@ private function getCheckerName(string $name)
             return false;
         }
 
+        if ($name == 'beReal') {
+            trigger_error('`beReal` matcher is deprecated, please use `beFloat`', E_USER_DEPRECATED);
```
stof: use `@trigger_error('...', E_USER_DEPRECATED);` to follow the opt-in deprecation reporting that the Symfony community defined (after breaking the whole ecosystem in 2.7.0 when trying to use it without silencing, for similar reasons than what you have here)

ddziaduch: Sorry, but how this is supposed to work? I trigger notice and immediately silence it. How the users will know that this matcher is deprecated?

ddziaduch: @stof OK I found something about it

stof: an error handler is still able to handle silenced errors, but they can know whether what they handle is silenced or no.

And well-behaving error handlers which are converting errors to exceptions would be doing it only for non-silenced errors, which is what makes the deprecation warning opt-in

ddziaduch: Thanks. I've fixed that. But I have no idea how to test this using PhpSpec. What about integration test?

ddziaduch: @stof can you review my changes again please? :)

gep:
@stof @ddziaduch can we merge it since requested changes are resolved?

composer.json
```diff
@@ -23,7 +23,7 @@
 
     "require": {
         "php":                      "^7.2, <7.4",
```
gep: Do we need to change it to smth like `"^7.2, 7.4.0",` in order to run on PHP 7.4.0?

ciaranmcnulty: just `^7.2 <7.5`

ddziaduch:
> @stof @ddziaduch can we merge it since requested changes are resolved?
> Or we need to get #1266 done before?

Have no idea 🙂

composer.json
gep: I think just `^7.2` is even better.

ahilles107:
With 7.4 release tagged - can we speedup somehow merging of this PR? Looks like all tests passed.

composer.json
ciaranmcnulty: @gep no, PHP doesn't use server (otherwise this PR wouldn't have been necessary)

composer.json
```diff
@@ -22,8 +22,8 @@
     ],
 
     "require": {
-        "php":                      "^7.2, <7.4",
-        "phpspec/prophecy":         "^1.7",
+        "php":                      "^7.2",
```
ciaranmcnulty:
```suggestion
        "php":                      "^7.2, <7.5",
```

src/PhpSpec/Matcher/ScalarMatcher.php
```diff
@@ -131,6 +131,10 @@ private function getCheckerName(string $name)
         if ($expected == 'boolean') {
             return 'is_bool';
         }
+        if ($expected == 'real') {
+            @trigger_error('`beReal` matcher is deprecated, please use `beFloat`', E_USER_DEPRECATED);
```
ciaranmcnulty: I'd rather not trigger the error
```suggestion
```

ciaranmcnulty:
Thanks for this work @ddziaduch - if you address my suggestions this can be pushed out pretty fast

ddziaduch:
@ciaranmcnulty it's done 🙂

ciaranmcnulty:
Oh I misunderstood and thought that this was putting the 7.4 tests into the Travis grid

ciaranmcnulty:
Looks like 7.4 Travis on master is failing, I'll take a look later today

Issue #1217, opened 2018-08-16
No line number in syntax error


jleaders:
At command line I run
```
$ bin/phpspec run                                                                                                                                                                          

```

And the output is:
```
      PSM\Business\Client

 119  ! Loading specification
        exception [err:ParseError("syntax error, unexpected '}', expecting ';'")] has been thrown.

----  broken examples

        PSM/Business/Client
 119  ! Loading specification
        exception [err:ParseError("syntax error, unexpected '}', expecting ';'")] has been thrown.


1 specs
1 examples (1 broken)
5ms

```

I'm sure my code has a syntax error, however this bug is showing that there is no line number showing in the ParseError. **How will I ever track down errors without a line number?**

I tried with & without formatter: pretty

**phpspec.yml**
```
suites:
    default:
        namespace: PSM\Business
        psr4_prefix: PSM\Business
#formatter.name: pretty
```

nightlinus:
Does `phpspec run -v` solves the issue?

jleaders:
Unfortunately I moved away from PHPSpec towards a library that didn't require verbose output for usage, so I can't test anymore easily. 
Perhaps if the exception type is a parse error it could be more verbose, or at least suggest to add `-v`?

ddziaduch:
@jleaders what library did you chose?

DonCallisto:
`-v` (or `-vv` or `-vvv` there's no difference from PHPSpec prospective) will show file and line.
We build the exception this way https://github.com/phpspec/phpspec/blob/9d592837d0fdbbdfd53168af0c9ef92bd200e6bc/src/PhpSpec/Exception/ErrorException.php#L20-L23

We can enhance message with file and code line, otherwise we can stay with verbose output.

PR #1271, opened 2019-10-21
Augments parse error output with file and line

gquemener:
See https://github.com/phpspec/phpspec/issues/1217

I believe that it enhances DX to always (not only with verbosity on) display file and line number in case of a parsing error.

WDYT?

### DIFF
--- a/spec/PhpSpec/Formatter/Presenter/Value/BaseExceptionTypePresenterSpec.php
+++ b/spec/PhpSpec/Formatter/Presenter/Value/BaseExceptionTypePresenterSpec.php
@@ -28,4 +28,17 @@ class BaseExceptionTypePresenterSpec extends ObjectBehavior
         $this->present(new ErrorException(new \Error('foo')))
             ->shouldReturn('[err:Error("foo")]');
     }
+
+    function it_should_present_a_parse_error_with_file_and_line_number()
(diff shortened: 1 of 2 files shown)


spec/PhpSpec/Formatter/Presenter/Value/BaseExceptionTypePresenterSpec.php
```diff
@@ -28,4 +28,17 @@ function it_should_present_an_error_as_a_string()
...
+    {
+        $this->present(new ErrorException(new class() extends \ParseError {
+            public function __construct()
+            {
+                $this->message = 'Something is not correct';
+                $this->file = '/app/some/file.php';
+                $this->line = 42;
+            }
+        }))
+            ->shouldContain('Something is not correct in "/app/some/file.php" on line 42');
```
gquemener: Another suggestion would be `Something is not correct in /app/some/file.php:42`.

Some cool terminal might allow to quick open the file (by clicking on the path) in one's favorite editor. I've tried unsuccessfully with `guake` though => https://guake.readthedocs.io/en/stable/user/features.html#quick-open

ciaranmcnulty:
Looks like a nice enhancement, thanks

Issue #1284, opened 2019-11-30
Regression on 6.1.0 for private constructors


tseho:
When upgrading from 6.0.0 to 6.1.0, one of our test is now failing when using a private constructor.

Given a class like this:
```
class Option
{
    private $value;

    private function __construct(string $value)
    {
        $this->value = $value;
    }

    public static function create(string $value)
    {
        return new self($value);
    }
}
```

And the related spec:
```
class OptionSpec extends ObjectBehavior
{
    public function it_is_initializable()
    {
        $this->shouldHaveType(Option::class);
    }

    public function it_can_be_constructed_through_create()
    {
        $this->beConstructedThrough('create', ['foo']);
        $this->shouldBeAnInstanceOf(Option::class);
    }
}
```

We now have, in 6.1.0, the error `exception [exc:ReflectionException("Access to non-public constructor of class Option")] has been thrown`.

I don't understand what changed but I created a repo for reproducing the problem:
with phpspec 6.0.0 : https://github.com/tseho/phpspec-regression-6.1.0/tree/6.0.0
with phpspec 6.1.0 : https://github.com/tseho/phpspec-regression-6.1.0/tree/6.1.0

EDIT:
I overlooked the fact we don't specify the factory in the `it_is_initializable`, so our test was bad but it was passing somehow.

ciaranmcnulty:
This is still an issue IMO if it's breaking behaviour in a minor

ciaranmcnulty:
It's caused by https://github.com/phpspec/phpspec/pull/1267/commits/33efa6a641c52a327906064014fab70673c406c9

Previously `shouldHaveType` wasn't triggering instantiation (which is weird behaviour, see #1233)

I'm up for changing this behaviour for 7.0.0 but we should:
a) add an acceptance test for this case
b) revert the behaviour

I could revert all of PR #1267 for now, unless @ddziaduch wants to take a look?

tseho:
Ok, I haven't been far enough and I didn't know it was previously using `newInstanceWithoutConstructor`.
I agree this change should be included in a major.

ddziaduch:
> I overlooked the fact we don't specify the factory in the it_is_initializable, so our test was bad but it was passing somehow.

Ah yes, when I was working on the fix I got the same thing. The PhpSpec itself was silencing this error. Seems that my fix fix that so your specs are now failing 🙈

@ciaranmcnulty I think we should revert that and put that into major version (7?)

tkotosz:
[...]

PR #1267, opened 2019-10-08
#1183 Potentially undefined variable

ddziaduch:
Make sure that the variable is defined to prevent raw php errors.

Thix fixes #1183

PR #1285, opened 2019-12-02
Fix regression around triggering instantiation during type checks

ciaranmcnulty:
Changing the catch from Exception to Error in #1267 seems to have broken this behaviour, see discussion in #1284 

Possibly it should be Throwable?

Added an acceptance test around the behaviour that should be preserved and reverted the key change

### DIFF
--- a/spec/PhpSpec/Wrapper/Subject/Expectation/ConstructorDecoratorSpec.php
+++ b/spec/PhpSpec/Wrapper/Subject/Expectation/ConstructorDecoratorSpec.php
@@ -22,35 +22,20 @@ class ConstructorDecoratorSpec extends ObjectBehavior
 
     function it_rethrows_php_errors_as_phpspec_error_exceptions(Subject $subject, WrappedObject $wrapped)
     {
-        $subject->__call('getWrappedObject', [])->willThrow(new Error());
-        $this->shouldThrow(ErrorException::class)->duringMatch('be', $subject, array(), $wrapped);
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
ping @ddziaduch can you review?

spec/PhpSpec/Wrapper/Subject/Expectation/ConstructorDecoratorSpec.php
```diff
@@ -22,35 +22,20 @@ function let(Expectation $expectation)
 
     function it_rethrows_php_errors_as_phpspec_error_exceptions(Subject $subject, WrappedObject $wrapped)
     {
-        $subject->__call('getWrappedObject', [])->willThrow(new Error());
-        $this->shouldThrow(ErrorException::class)->duringMatch('be', $subject, array(), $wrapped);
+        $subject->__call('getWrappedObject', array())->willThrow('PhpSpec\Exception\Example\ErrorException');
```
ddziaduch: @ciaranmcnulty why `array()` instead of `[]`?

spec/PhpSpec/Wrapper/Subject/Expectation/ConstructorDecoratorSpec.php
```diff
@@ -22,35 +22,20 @@ function let(Expectation $expectation)
 
     function it_rethrows_php_errors_as_phpspec_error_exceptions(Subject $subject, WrappedObject $wrapped)
```
ddziaduch: I think you have to amend the title because you are throwing now PhpSpec exception instead of Php Error

PR #1290, opened 2019-12-21
Switch travis to bionic and use stable PHP 7.4

GrahamCampbell:


ciaranmcnulty:
Build seems broken on master too so ignore the failure :-/

GrahamCampbell:
I think it might have been caused by the phpspec/prophecy 1.9 to 1.10 upgrade, looking at the difference in dependencies installed between the most recent passing build, and the failing build just after it. The diff in installed deps is here: https://www.diffchecker.com/kSyK73OT.

GrahamCampbell:
Indeed, this seems to be the case. I've sent a PR that pins the version to 1.9.x: https://github.com/phpspec/phpspec/pull/1291.

We should later update the version to `~1.9.0 || ^1.10.1` if a fix is implemented in v1.10.1, for example (once we establish what has gone wrong - I've not looked at any of the detail, only the fact that "the phpspec tests don't pass with 1.10.0").

GrahamCampbell:
Looks like phpspec's test suite still fails with prophecy 1.10.1.

GrahamCampbell:
What's the status here? :)

PR #1291, opened 2019-12-22
Don't allow prophecy 1.10

GrahamCampbell:
Prophecy 1.10.0 breaks the tests, as we can see from the travis run of the 6.1.1 release commit. The commit just before it ran with 1.9.0, and the tests passed.

I propose that we disallow 1.10+ for now, until it's established what the issue is, and what fixes are required.

---

See https://github.com/phpspec/phpspec/pull/1290#issuecomment-568295565 where this was pointed out.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
 
     "require": {
         "php":                      "^7.2, <7.5",
-        "phpspec/prophecy":         "^1.9",
+        "phpspec/prophecy":         "~1.9.0",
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Looks like phpspec's test suite still fails with prophecy 1.10.1.

ciaranmcnulty:
Yep, probably needs fixing upstream then - I've not got time to dig into why it's failing right now due to family stuff

PR #1295, opened 2020-02-06
Convert docs to Hugo

jaylinski:


drupol:
Good idea :)

jaylinski:
@drupol We'll see :sweat_smile:  (it isn't quite ready yet)

ciaranmcnulty:
We currently host with readthedocs.io, does that support Hugo?

GrahamCampbell:
No. Hugo is a static site builder? It would be to replace readthedocs.io as the "thing that renders the pages" I think.

jaylinski:
I'm using GitHub actions to automatically build the docs with Hugo and push them to the `gh-pages` branch.

WIP: https://jaylinski.github.io/phpspec/

Hugo is not specifically made for docs, so it lacks some features. I'll explain it in a new PR so you can decide if you want to migrate to Hugo or stick with readthedocs (which seems broken).

drupol:
There's this very new theme that could fit for documentation: https://themes.gohugo.io/compose/

drupol:
> WIP: https://jaylinski.github.io/phpspec/

Amazing wip, nice work !

I definitely vote for Hugo :-)
That said, it would be nice to know why readthedocs is broken, I never had any issue with ReadTheDocs and it's working well for my own projects.

PR #1309, opened 2020-06-18
Move PHAR build to makefile and use newer version of box

ciaranmcnulty:
Moving to the newer version of box with a view to scoping the dependencies in future

It struck me this isn't adequately tested so I've executed some of the behat tests against the PHAR - we could add more via a `@phar` tag

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@
 behat.yml
 vendor
 composer.lock
-phpspec.phar
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ box.phar:
 
 phpspec.phar: box.phar vendor
 	./box.phar compile
+	vendor/bin/behat --profile=phar --format=progress || rm phpspec.phar
 
 vendor: composer.phar
 	./composer.phar install
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,16 @@
+box.phar:
+	curl -Lls https://github.com/humbug/box/releases/download/3.8.4/box.phar -o box.phar
+	chmod +x box.phar
+
+phpspec.phar: box.phar vendor
+	./box.phar compile
+
+vendor: composer.phar
+	./composer.phar install
+
+composer.phar:
+	curl -Lls https://getcomposer.org/installer | php
+
+.PHONY: clean
+clean:
+	rm -f phpspec.phar box.phar composer.phar
--- a/box.json
+++ b/box.json
@@ -1,5 +1,4 @@
 {
-    "chmod": "0755",
     "directories": [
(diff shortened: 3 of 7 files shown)


jakzal:
Summoning @theofidry

ciaranmcnulty:
@jakzal Not sure how to test I didn't break the deploy step in travis

.gitignore
```diff
@@ -3,4 +3,3 @@
 behat.yml
 vendor
 composer.lock
-phpspec.phar
```
theofidry, line 6: shouldn't it remain ignored?

Makefile
```diff
@@ -0,0 +1,17 @@
+box.phar:
```
theofidry, line 1: you probably want to add `touch $@` at the end of each to ensure the file timestamp is properly updated

Makefile
```diff
@@ -0,0 +1,17 @@
...
+	curl -Lls https://github.com/humbug/box/releases/download/3.8.4/box.phar -o box.phar
```
theofidry, line 2: is it well understood here that if you change the link it is not enough to invalidate the existing `box.phar`?

Makefile
```diff
@@ -0,0 +1,17 @@
...
+	chmod +x box.phar
+
+phpspec.phar: box.phar vendor
```
theofidry, line 5: I think it should also depend on the project source code, e.g. `$(shell find bin/ src/ -type f)` (see https://github.com/infection/infection/blob/master/Makefile#L208)

box.json
```diff
@@ -1,5 +1,4 @@
 {
-    "chmod": "0755",
     "directories": [
```
theofidry, line 2: might be worth checking but you can move the current `files` to `files-bin` and remove completely `directories` and `finder`

.gitignore
ciaranmcnulty: There's *.phar earlier in the file

box.json
ciaranmcnulty: Thanks, I'll try it out

ciaranmcnulty:
Thanks for the comments @theofidry - I'm not sure why invalidation is important in this case, as we're always making from a clean travis. Are you thinking about the developer who's made it locally and isn't doing `make clean` each time?

ciaranmcnulty:
@theofidry And thanks for that hard work on box etc ;-)

theofidry:
>Are you thinking about the developer who's made it locally and isn't doing make clean each time?

Yes: you'll likely have to create one and check it one locally at some point so having to clean it manually between each change might unnecessarily annoying

You're welcome :)

PR #1310, opened 2020-06-18
Scope the Phar dependencies

ciaranmcnulty:
Tests seem to be passing, but is more coverage needed?

### DIFF
--- a/box.json
+++ b/box.json
@@ -12,5 +12,8 @@
             "in": "vendor"
         }
     ],
-    "output": "phpspec.phar"
+    "output": "phpspec.phar",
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
I may just #yolo this. Worst that can happen is a PHAR user says it's broken

GrahamCampbell:
👍 Best way to test this is to see who complains about stuff breaking. This is pretty safe, since this package is a test runner, and not something run on every request in someone's app, by comparison. :)

GrahamCampbell:
Call it the integration testing. :P

Issue #1304, opened 2020-06-16
Rename the branches


ciaranmcnulty:
There is some discussion around the branch naming in the community, mostly around what `master` sounds like. This has made me realise that the name doesn't really reflect what we use it for.

I tend to juggle the branches around when there's a new release and this is a good time.

We tend to have 2 active branches: one that's receiving new features and one that's receiving bug fixes (that get rolled forward into the other branch). We then have to document that `5.1` is the bug branch and `master` is the feature branch - it'd be easier if the names reflected that.

What's a good name for this? `bugfix` and `active?`

GrahamCampbell:
I think 5.1 and master are fine branch names, and people are used to that naming scheme, since it is what packagist expects branches to be named (other custom names require branch aliases (including master, but master is the convention among many projects inc phpunit, laravel, symfony)).

ciaranmcnulty:
Feedback I've had is that `master` isn't great. There's not anything particular about the name that makes me want to stick with it because it's neither mastering anything in the sense of controlling, and it's not really a master in the sense of something you take copies of.

If there's a better term I'd switch. Would indeed need to see the impact on packagist but TBH if people are using `dev-master` they can expect breakage

drupol:
I also think that this should not change, even if I understand the reasons. Reasons make senses, but people are used to this naming conventions.

IMHO, the amount of issue this would create doesn't worth the change.

GrahamCampbell:
Maybe if you make 5.1 the default branch, there will be less confusion?

drupol:
> Maybe if you make 5.1 the default branch, there will be less confusion?

That would be a good idea.

ciaranmcnulty:
5.1 is the bugfix branch... should that really be default?

drupol:
I guess he meant `6.0`

GrahamCampbell:
> I guess he meant 6.0

No, there is no 6.0 branch, and nor should there be. I really did mean 5.1. master is 6.1 at the moment (not 6.0).

GrahamCampbell:
Though, tbh, I don't see how there can be much confusion, if it is documented? Maybe add an issue template, like Laravel does.

drupol:
> 
> 
> > I guess he meant 6.0
> 
> No, there is no 6.0 branch, and nor should there be. I really did mean 5.1. master is 6.1 at the moment (not 6.0).

Sorry you're right.

I would vote for keeping `master` and having a `5.0` or `5.1` branch.

PR #1312, opened 2020-06-23
Update contributing notes to reflect branch rename

ciaranmcnulty:
As discussed in #1304 I'd like to rename the branches so that the contributing docs don't need to reflect version numbers

If this is merged I'll do the rename at the same time as doing the initial `7.0` release and EOL-ing `5.1`

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,9 +7,12 @@ feel free to do this, but remember to follow this few simple rules:
 Branching strategy
 -------------------
 
-- For new features, or bugs that only affect the 6.x versions, base your changes on the `master` branch and open PRs against `master`
-- For bugs that affect 5.1.x versions, base your changes on the `5.1` branch and open PRs agains `5.1`
-- Bugs in previous versions are not going to be fixed, upgrade to `5.1` minimum.
+At any given point there are three active branches:
+
+* Bug fixes that apply to old versions should target the current bugfix branch, which will be named after the last minor 
+version supported (e.g. `5.3`, `6.2`)
+* New features, refactoring and general cleanup should target the `main` branch and maintain backward compatibility
+* Any changes or refactoring that would introduce a backward incompatibility should target the `next` branch
 
 Coverage
 --------


CONTRIBUTING.md
```diff
@@ -7,9 +7,11 @@ feel free to do this, but remember to follow this few simple rules:
...
 -------------------
 
-- For new features, or bugs that only affect the 6.x versions, base your changes on the `master` branch and open PRs against `master`
-- For bugs that affect 5.1.x versions, base your changes on the `5.1` branch and open PRs agains `5.1`
-- Bugs in previous versions are not going to be fixed, upgrade to `5.1` minimum.
+There are three active branches, `previous`, `current` and `next`. To see how they correspond to version numbers check the [roadmap](https://github.com/phpspec/phpspec/wiki#roadmap)
+
+- Bug fixes should be against the `previous` branch, assuming the bug is present in that version
+- New features should be added to the `current` branch unless there is no way to implement them without breaking backwards compatibility
+- Breaking changes should only be made to the `next` branch
```
greg0ire, line 14: What about pedantic changes such as documentation, CI changes, cs changes, etc. ?

eugenekurasov:
Hi, @ciaranmcnulty 

I will not comment about change master to other name. But what reason for rename branches from [semversion](https://semver.org/) to `next` and `previous`?

For example if you have 5.1.x - you know that branch should include fix or small improve exist functional.
And when you use (composer) packages it's can be some diffuse with version and branch.

ciaranmcnulty:
@eugenekurasov tags would still be `7.1` etc

CONTRIBUTING.md
ciaranmcnulty: Thanks, I'll update

ciaranmcnulty:
An alternative would be to (at the next major) have `6.x`, `7.x`, `8.x` branches with `7.x` as the default branch?

This would mean documenting (maybe in the wiki as it lives inside VCS) which version is current, next, prev at any point

eugenekurasov:
My opinion naming default branch: `main` or `default` will be fine.

I can't understand why is 7.x/6.x/5.x bad branches?

ciaranmcnulty:
@eugenekurasov

> I can't understand why is 7.x/6.x/5.x bad branches?

Only because the contributing docs would need continuous update

eugenekurasov:
Hm, but if you upgrade major version you most probably be update docs (on change major version you most probably will broken BC).

PR #1313, opened 2020-06-26
Remove deprecated Reflection method calls

ciaranmcnulty:
PHP8 deprecates some of the reflection API

there are similar issues in dependent libraries (like Prophecy) so we won't be compatible until those are fixed too

### DIFF
--- a/src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
+++ b/src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
@@ -143,7 +143,7 @@ final class CollaboratorsMaintainer implements Maintainer
 
     private function isUnsupportedTypeHinting(\ReflectionParameter $parameter): bool
     {
-        return $parameter->isArray() || $parameter->isCallable();
+        return $parameter->getType() && in_array($parameter->getType()->getName(), ['array', 'callable']);
     }
 
     /**
@@ -201,7 +201,7 @@ final class CollaboratorsMaintainer implements Maintainer
     private function getParameterTypeFromReflection(\ReflectionParameter $parameter): string
     {
         try {
-            if (null === $class = $parameter->getClass()) {
+            if (null === $class = $parameter->getType()) {
                 return '';
             }
 


src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
```diff
@@ -143,7 +143,7 @@ private function generateCollaborators(CollaboratorManager $collaborators, \Refl
 
     private function isUnsupportedTypeHinting(\ReflectionParameter $parameter): bool
     {
-        return $parameter->isArray() || $parameter->isCallable();
+        return $parameter->getType() && in_array($parameter->getType()->getName(), ['array', 'callable']);
```
GrahamCampbell, line 146: This isn't the same? `isCallable` would return true if the type name is `Closure` too.

GrahamCampbell: Also, I think this is only deprecated in PHP 8. If you are looking for proper PHP 8 support then you will also need to check that `getType` returns an object which has `getName` (`instance of ReflectionNamedType`). Union typed objects only have a `getNames` method.

GrahamCampbell: https://github.com/laravel/framework/blob/v7.17.2/src/Illuminate/Support/Reflector.php#L10-L31 might be a helpful example RE the getName stuff.

GrahamCampbell: > This isn't the same? isCallable would return true if the type name is Closure too.

There could also be a typed class that is invokable. You'd need to check that too (and also the special case of `self`, which doesn't get resolved to the real class name).

stof: IIRC, `isCallable` for parameter is only about the `callable` typehint

GrahamCampbell: Oh, so not the same as `is_callable`?

ciaranmcnulty: @GrahamCampbell thanks, we'll need to consider how union types will even work I guess!

ciaranmcnulty: @GrahamCampbell No, `isCallable` is just about `callable`. To evaluate whether a string or array or object was callable would require autoloading and all sorts of runtime stuff

https://3v4l.org/oktBd

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
```diff
@@ -201,7 +201,7 @@ private function getParameterTypeFromIndex(\ReflectionClass $classRefl, \Reflect
     private function getParameterTypeFromReflection(\ReflectionParameter $parameter): string
     {
         try {
-            if (null === $class = $parameter->getClass()) {
+            if (null === $class = $parameter->getType()) {
```
GrahamCampbell, line 204: This is actually not the same as before. A reflection exception will not be thrown if the class doesn't exist. You need to do a class_exists check and throw the collaborator not found yourself.

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
GrahamCampbell: Yep. You are right. That line looks great. 👍 

NB: I'm unable to mark this comment as resolved for some reason.

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
ciaranmcnulty: oh that explains why that exception is caught elsewhere too - thanks

GrahamCampbell:
See https://github.com/ciaranmcnulty/phpspec/pull/3.

src/PhpSpec/Runner/Maintainer/CollaboratorsMaintainer.php
GrahamCampbell: > A reflection exception will not be thrown if the class doesn't exist.

By that I mean, `getClass` would throw an exception, but `getType` does not. :)

Issue #1318, opened 2020-07-08
Switch to GitHub Actions


ciaranmcnulty:
See #1317 

@drupol talk here please

drupol:
Sure, maybe we can get inspired by what has been done here ?

https://github.com/friends-of-phpspec/phpspec-code-coverage/pull/25

I can give it a try tonight, I'll catch you on Slack when I start working on this.

PR #1320, opened 2020-07-08
Remove scrutinizer config

ciaranmcnulty:
It's rarely looked at and we can use Actions for tools instead (#1318)


drupol:
I don't think Github actions can do what Scrutinizer do.

I like Scrutinizer to tell me the percentage of code coverage and the code quality. AFAIK, Github Actions doesn't do that.

ciaranmcnulty:
I mean using open-source tools inside actions

jakzal:
@ciaranmcnulty what about the badge in README?

stof:
@drupol Scrutinizer was not setup for code coverage anyway, only for static analysis

PR #1323, opened 2020-07-14
Remove type hint for result parameter in getFailureExceptionFor method

naomigo:
Removing the type hint ensures that calls to the `shouldBe...()` and `shouldHave...()` methods fail gracefully, providing representitive error output.


ciaranmcnulty:
Thanks!

PR #1324, opened 2020-07-26
Add support for the "fake" option for ObjectStateMatcher

naomigo:
This PR makes it possible to run the `--fake` option when running tests which use the object state matcher.

Previously this was not supported because any methods that begin with the word `should` are automatically skipped by the `Subject::__call` method. This meant that the runner never ended up in the `MethodReturnedNullListener::afterMethod` method, which then defined the `lastMethodCallEvent` property which was essential for the following code in the `afterSuite` method, which is where the prompt for faking occurs.

The changes introduced are:
- Add support for defining the subject and method called in `NotEqualException`.
- Modify `ObjectStateMatcher` to throw `NotEqualException` rather than it's parent `FailureException`, passing the subject and method properties
- Have `MethodReturnedNullListener` check for the existence of subject and method in the exception object when `lastMethodCallEvent` has not been defined.


features/code_generation/developer_generates_return_constant.feature
```diff
@@ -314,3 +314,195 @@ Feature: Developer generates a method returning a constant
...
+
+      class MyFeature
+      {
+          public function hasAvailability()
+          {
+              return false;
+          }
+      }
+
+      """
```
ciaranmcnulty, line 508: Please remove this line break

src/PhpSpec/Matcher/ObjectStateMatcher.php
```diff
@@ -120,15 +120,15 @@ public function getPriority(): int
...
      * @param mixed    $result
      *
-     * @return FailureException
+     * @return NotEqualException
      */
-    private function getFailureExceptionFor(callable $callable, bool $expectedBool, $result): FailureException
+    private function getNotEqualExceptionFor(callable $callable, bool $expectedBool, $result): NotEqualException
     {
-        return new FailureException(sprintf(
+        return new NotEqualException(sprintf(
```
ciaranmcnulty, line 127: I feel like this is overloading NotEqualException TBH, it might be better to have a new exception type that's handled in MethodReturnedNullListener?

features/code_generation/developer_generates_return_constant.feature
ciaranmcnulty, line 508: Please add a line break

ciaranmcnulty (commented):
Sorry this wasn't reviewed earlier, very minor points

ciaranmcnulty:
@naomigo I appreciate you had to wait a while, but are you able to address the comments?

features/code_generation/developer_generates_return_constant.feature
naomigo: Added in 9e4a84afd4038b8911fe216409ccdd92f99ef557

src/PhpSpec/Matcher/ObjectStateMatcher.php
naomigo: I have added a MethodFailureException class in a82ffef0842e1ba0744276dd7fbed6bc03105a2f - hope this is what you were meaning?

naomigo:
@ciaranmcnulty Sorry - one of those things I had completely forgotten about! Have posted some updates.

ciaranmcnulty:
Thanks!

PR #1325, opened 2020-09-16
Support the new tokens in PHP8 for namespaces

ciaranmcnulty:
Rather than looping through the string tokens building the namespace we can grab it as one string

### DIFF
--- a/src/PhpSpec/Locator/PSR0/PSR0Locator.php
+++ b/src/PhpSpec/Locator/PSR0/PSR0Locator.php
@@ -297,7 +297,8 @@ class PSR0Locator implements ResourceLocator, SrcPathLocator
         for ($i = 0; $i < $count; $i++) {
             if ($tokens[$i][0] === T_NAMESPACE) {
                 for ($j = $i + 1; $j < $count; $j++) {
-                    if ($tokens[$j][0] === T_STRING || $tokens[$j][0] === 314 || $tokens[$j][0] === 312) {
+                    if ($tokens[$j][0] === T_STRING
+                        || \PHP_VERSION_ID >= 80000 && ($tokens[$j][0] === T_NAME_FULLY_QUALIFIED || $tokens[$j][0] === T_NAME_QUALIFIED)) {
                         $namespace .= $tokens[$j][1].'\\';
                     } elseif ($tokens[$j] === '{' || $tokens[$j] === ';') {
                         break;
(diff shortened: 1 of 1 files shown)


src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -297,9 +297,10 @@ private function findSpecClassname($path)
         for ($i = 0; $i < $count; $i++) {
             if ($tokens[$i][0] === T_NAMESPACE) {
                 for ($j = $i + 1; $j < $count; $j++) {
-                    if ($tokens[$j][0] === T_STRING) {
+                    if ($tokens[$j][0] === T_STRING || $tokens[$j][0] === 314 || $tokens[$j][0] === 312) {
                         $namespace .= $tokens[$j][1].'\\';
-                    } elseif ($tokens[$j] === '{' || $tokens[$j] === ';') {
+                    }
+                    elseif ($tokens[$j] === '{' || $tokens[$j] === ';') {
```
stof, line 303: this change looks wrong to me

ciaranmcnulty: yeah I'll revert it

src/PhpSpec/Locator/PSR0/PSR0Locator.php
```diff
@@ -297,9 +297,10 @@ private function findSpecClassname($path)
         for ($i = 0; $i < $count; $i++) {
             if ($tokens[$i][0] === T_NAMESPACE) {
                 for ($j = $i + 1; $j < $count; $j++) {
-                    if ($tokens[$j][0] === T_STRING) {
+                    if ($tokens[$j][0] === T_STRING || $tokens[$j][0] === 314 || $tokens[$j][0] === 312) {
```
ciaranmcnulty, line 300: Can't use the token names here because undefined under PHP7

Could use token_name to be more explicit, or just leave a comment?

stof: Please either add a comment or change this to `(\PHP_VERSION_ID >= 80000 && ($tokens[$j][0] === T_FULLY_QUALIFIED_NAME || $tokens[$j][0] === T_QUALIFIED_NAME))` (please check the constant name, I wrote them from memory).
The advantage of the second alternative is that it still uses the actual constants (and we would remove the version check in a future version of phpspec dropping support for PHP 7 by looking for version checks) while still resolving the version check at compile time to avoid any overhead (that's why I'm qualifying the constant)

ciaranmcnulty: Short-circuiting the if is a great suggestion thank you

stof (approved):

PR #1332, opened 2020-10-06
Cope with different T_COMMENT tokenisation in PHP 8

ciaranmcnulty:
T_COMMENT no longer captures the following newline so we need to look further back in the tokens when deciding whether to add a newline


ciaranmcnulty:
Thanks @GrahamCampbell for the link to the PHP change

PR #1337, opened 2020-10-26
Configure the build on github actions

jakzal:
To start I only aimed to reproduce what's currently done on Travis.

# Matrix

| PHP | Composer flags | Composer extras | Stability |
|--|--|--|--|
| 7.3 |  | | stable |
| 7.3 | --prefer-lowest | | stable |
| 7.3 |  | symfony/event-dispatcher ^3.4 symfony/contracts ^2.0 | stable |
| 7.4 | | | stable |
| 7.4 | | | dev |

image

# Notes

* The PHAR is built on each job, but it's only published as a build artifact on PHP 7.3 stable (so it's only published once).
* The PHAR is attached to a release if the build is triggered by a release.
* The only thing I couldn't get to work is "allow failures". It's documented here, but doesn't seem to work: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error
* Behat tests are currently disabled as they're failing.

# Example builds

Example build: https://github.com/jakzal/phpspec/actions/runs/331115517
Example build for a release: https://github.com/jakzal/phpspec/actions/runs/331116603
Example release: https://github.com/jakzal/phpspec/releases/tag/0.0.1-test

# Build times

## Github Actions

image

## Travis CI

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,90 @@
+name: Build
+
+on:
+    push:
+        branches: [main, next, master, 2.5, 3.4, 4.3, 5.1, 6.2]
+    pull_request:
+    release:
+        types: [created]
+
+jobs:
+    tests:
+        runs-on: ubuntu-latest
+        name: Build and test
+        strategy:
+            matrix:
+                php: [7.3, 7.4]
+                composer-flags: [""]
+                composer-extras: [""]
+                stability: ["stable"]
+                include:
+                    -   php: 7.3
+                        publish-phar: true
+                    -   php: 7.3
+                        composer-flags: --prefer-lowest
+                    -   php: 7.3
+                        composer-extras: "symfony/event-dispatcher ^3.4 symfony/contracts ^2.0"
+                    -   php: 7.4
+                        stability: "dev"
+
+        env:
+            COMPOSER_ROOT_VERSION: dev-master
+
+        steps:
+            -   uses: actions/checkout@v2
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
+                with:
+                    php-version: "${{ matrix.php }}"
+                    ini-values: "phar.readonly=0"
+
+            -   name: Install composer extras
+                if: matrix.composer-extras != ''
+                run: composer require ${{ matrix.composer-extras }} --no-update
+
+            -   name: Update the minimum stability
+                if: matrix.stability == 'dev'
+                run: composer config minimum-stability ${{ matrix.stability }}
+
+            -   name: Install dependencies
+                run: composer update ${{ matrix.composer-flags }}
+
+            -   name: Run tests (phpspec)
+                run: bin/phpspec run --format=dot
+
+            -   name: Run tests (phpunit)
+                run: ./vendor/bin/phpunit -v
+
+            -   name: Build the PHAR
+                run: |
+                    composer config platform.php 7.3.0 &&
+                    ln -s `which composer` composer.phar &&
+                    make phpspec.phar
+
+            -   uses: actions/upload-artifact@v1
+                name: Publish the PHAR
+                if: matrix.publish-phar
+                with:
+                    name: phpspec.phar
+                    path: phpspec.phar
+
+    publish-phar:
+        runs-on: ubuntu-latest
+        name: Publish the PHAR
+        needs: tests
+        if: github.event_name == 'release'
+        steps:
+            -   uses: actions/download-artifact@v1
+                with:
+                    name: phpspec.phar
+                    path: .
+            -   name: Upload phpspec.phar
+                uses: actions/upload-release-asset@v1
+                env:
+                    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                with:
+                    upload_url: ${{ github.event.release.upload_url }}
+                    asset_path: phpspec.phar
+                    asset_name: phpspec.phar
+                    asset_content_type: application/zip


.github/workflows/build.yml
```diff
@@ -0,0 +1,59 @@
+name: Build
+
+on:
+    push:
+        branches: [main, next, master, 2.5, 3.4, 4.3, 5.1, 6.2]
+    pull_request:
+    release:
+        types: [created]
+    schedule:
+        -   cron: '0 4 * * *'
```
ciaranmcnulty, line 10: Every day seems a lot?

.github/workflows/build.yml
```diff
@@ -0,0 +1,59 @@
...
+            -   name: Run tests (phpspec)
+                run: bin/phpspec run --format=dot
+
+            -   name: Run tests (phpunit)
+                run: ./vendor/bin/phpunit -v
+
+            -   name: Run tests (behat)
+                run: ./vendor/bin/behat --format=progress
+
+            -   name: Build PHAR
```
ciaranmcnulty, line 50: Is this going to upload the PHAR from every build?

ciaranmcnulty:
How come it doesn't do the checks in this PR (as a travis change would?)

stof:
@ciaranmcnulty that's because there is no workflow file in the target branch, so the workflow is not triggered (that would be the same in Travis if you don't enable the webhook before adding the config file).

.github/workflows/build.yml
```diff
@@ -0,0 +1,59 @@
...
+                run: |
+                    composer config platform.php 7.3.0 &&
+                    ln -s `which composer` composer.phar &&
+                    make phpspec.phar
+
+            -   uses: actions/upload-artifact@v1
```
stof, line 56: this should run only for release runs

stof: also, probably not for all versions of the matrix. Instead, we should have a `build phar` job which depends on all tests jobs and deals with the phar when tests are successful.

.github/workflows/build.yml
jakzal: Ok, Disabled for now.

.github/workflows/build.yml
jakzal: Yes, I planned to only upload it for 7.3 (or whatever the current minimum is).

I think it should run on every build, not just on release builds. It's useful to be able to download the phar for any build.

Attaching the file to the release is still left to do.

drupol:
If you're looking for some inspirations: https://github.com/loophp/collection/blob/master/.github/workflows/continuous-integration.yml

* Run on Mac, Windows and Linux
* It caches the `composer install`
* It fix Git LF
* Building on Lower/high deps should be quite easy to integrate

.github/workflows/build.yml
jakzal: Actually, I think it's worth to build the PHAR on every PHP version to make sure it continues building on all of them, but only publish the phar on 7.3.

.github/workflows/build.yml
```diff
@@ -0,0 +1,94 @@
...
+                composer-flags: [""]
+                composer-extras: [""]
+                stability: ["stable"]
+                include:
+                    -   php: 7.3
+                        publish-phar: true
+                    -   php: 7.3
+                        composer-flags: --prefer-lowest
+                    -   php: 7.3
+                        composer-extras: "symfony/event-dispatcher ^3.4 symfony/contracts ^2.0"
```
drupol, line 26: Shouldn't it be written as such?

```suggestion
                        composer-extras: "symfony/event-dispatcher^3.4 symfony/contracts^2.0"
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,94 @@
...
+                run: composer update ${{ matrix.composer-flags }}
+
+            -   name: Run tests (phpspec)
+                run: bin/phpspec run --format=dot
+
+            -   name: Run tests (phpunit)
+                run: ./vendor/bin/phpunit -v
+
+            -   name: Run tests (behat)
+                if: false
```
drupol, line 60: How about replacing this with

```suggestion
                continue-on-error: true
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,94 @@
...
+                        stability: "dev"
+
+        env:
+            COMPOSER_ROOT_VERSION: dev-master
+
+        steps:
+            -   uses: actions/checkout@v2
+
+            -   name: Setup PHP
+                uses: shivammathur/setup-php@v2
```
drupol, line 37: How about caching this?

.github/workflows/build.yml
jakzal: I kept what was there in the travis config.

.github/workflows/build.yml
jakzal: This should go away once behat tests are fixed.

.github/workflows/build.yml
jakzal: There was no caching on travis. I suggest we reproduce the same build first, and then see if caching is worth it.

Build on Github takes slightly above a minute. Caching might actually slow it down.

.github/workflows/build.yml
drupol: Fair enough!

.github/workflows/build.yml
drupol: Ok

.github/workflows/build.yml
ciaranmcnulty: Yes I misunderstood that it'd be uploading to the release. If it's to the GH actions page that's handy

.github/workflows/build.yml
jakzal: @drupol I was wrong, Travis is configured to use caching:

```
cache:
  directories:
    - $HOME/.composer/cache
```

I'm still opting to do it once this PR is merged. Perhaps you'd be up for it? ;)

drupol: I'm ok to take care of it, but it's very easy to implement, see this comment: https://github.com/phpspec/phpspec/pull/1337#issuecomment-716707082

The file linked in there add steps to enable caching:

```
            - name: Get Composer Cache Directory
              id: composer-cache
              run: echo "::set-output name=dir::$(composer config cache-files-dir)"

            - name: Cache dependencies
              uses: actions/cache@v2
              with:
                  path: ${{ steps.composer-cache.outputs.dir }}
                  key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
                  restore-keys: ${{ runner.os }}-composer-
```

jakzal: Yeah, I know how to do it, I just don't think it's worth it! :)

drupol: Up to you to decide, I'm just proposing :-)

.github/workflows/build.yml
```diff
@@ -0,0 +1,94 @@
...
+            -   name: Install dependencies
+                run: composer update ${{ matrix.composer-flags }}
+
+            -   name: Run tests (phpspec)
+                run: bin/phpspec run --format=dot
+
+            -   name: Run tests (phpunit)
+                run: ./vendor/bin/phpunit -v
+
+            -   name: Run tests (behat)
```
ciaranmcnulty, line 59: Just take this out if it's not working; we can add it in with the PR fixing the tests

jakzal: Done 👍

ciaranmcnulty:
Thanks @jakzal!

Ideally we'll get the behat tests working and remove .travis.yml before the next release, or we'll have two PHARs

drupol:
Nice!

PR #1341, opened 2020-10-29
Remove superfluous PHPDoc tags.

drupol:
Hello,

While doing a new extension ([loophp/phpspec-time](https://github.com/loophp/phpspec-time)) and fixing Psalm issues, I noticed this issue:

```
ERROR: UndefinedDocblockClass - src/Matcher/TakeInBetweenMatcher.php:23:73 - Docblock-defined class or interface PhpSpec\Matcher\Boolean does not exist (see https://psalm.dev/200)
    public function supports(string $name, $subject, array $arguments): bool
```

The supports method is defined in [Matcher.php](https://github.com/phpspec/phpspec/blob/main/src/PhpSpec/Matcher/Matcher.php#L29).

As we can see, Psalm is (often) right and `Boolean` is not a PHP type nor a Class.

We should either replace it with `bool` or remove it because it's superfluous and already announced in the signature of the method.

This PR:

* [x] Remove all the superluous tags from the source code with PHPCSFixer
* [x] Add the `bool` return type to `features/bootstrap/Fake/ReRunner.php` (_that's the only manual edit in the PR_)

To remove those superfluous tags, I used PHP CS Fixer:  

`php-cs-fixer fix --rules=no_superfluous_phpdoc_tags,no_empty_phpdoc src/`

### DIFF
--- a/features/bootstrap/Fake/ReRunner.php
+++ b/features/bootstrap/Fake/ReRunner.php
@@ -8,10 +8,7 @@ class ReRunner implements BaseReRunner
 {
     private $hasBeenReRun = false;
 
-    /**
-     * @return boolean
-     */
-    public function isSupported()
+    public function isSupported(): bool
     {
         return true;
     }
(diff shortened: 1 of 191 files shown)


ciaranmcnulty:
It’s annoying because Boolean was correct symfony coding style when the code was written :)

drupol:
Up to you mate!

I think those rules are now used in modern versions of Symfony.

drupol:
Thanks!

PR #1342, opened 2020-11-01
Allow testing and install under PHP 8

ciaranmcnulty:


composer.json
```diff
@@ -22,7 +22,7 @@
     ],
 
     "require": {
-        "php":                      "^7.2, <7.5",
+        "php":                      "^7.2 || 8.0.*",
```
GrahamCampbell, line 25: Does `>=7.2, <8.1` work? I'm in particular thinking about how this constraint will look when 8.1 comes around. ;)

GrahamCampbell (changes requested):

composer.json
ciaranmcnulty: It'd need to be `^7.2 || ^8.0, <8.1` I think.

Maybe let's cross the bridge when we come to it?

PR #1345, opened 2020-11-03
Cache the composer folder during checks

ciaranmcnulty:
Use `composer.json` as a cache key

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,8 +48,19 @@ jobs:
                 if: matrix.stability == 'dev'
                 run: composer config minimum-stability ${{ matrix.stability }}
 
-            -   name: Install dependencies
-                run: composer update ${{ matrix.composer-flags }}
+            -   name: Resolve composer dependencies
+                run: composer update ${{ matrix.composer-flags }} --no-install
+
+            -   name: Cache composer
+                uses: actions/cache@v2
+                env:
+                    cache-name: cache-composer
+                with:
+                    path: ~/.composer
+                    key: ${{ hashFiles('composer.lock') }}
+
+            -   name: Install composer dependencies
+                run: composer install
 
             -   name: Run tests (phpspec)
                 run: bin/phpspec run --format=dot
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,19 +48,16 @@ jobs:
                 if: matrix.stability == 'dev'
                 run: composer config minimum-stability ${{ matrix.stability }}
 
-            -   name: Resolve composer dependencies
-                run: composer update ${{ matrix.composer-flags }} --no-install
-
-            -   name: Cache composer
+            -   name: Cache composer dependencies
                 uses: actions/cache@v2
                 env:
                     cache-name: cache-composer
                 with:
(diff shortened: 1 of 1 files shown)


drupol:
Reference documentation: https://github.com/shivammathur/setup-php#cache-composer-dependencies

drupol:
Wouldn't it be better to use composer.json instead as composer.lock is not committed?

ciaranmcnulty:
@drupol thanks for the documentation

I think maybe you're right about the JSON - in the case where newer dependencies are available we would get at least partial caching. I'll amend

drupol:
You can get inspired here: https://github.com/loophp/collection/blob/master/.github/workflows/continuous-integration.yml

I'm using that for all my libs, works pretty good :-)

jakzal (approved):

.github/workflows/build.yml
```diff
@@ -48,7 +48,19 @@ jobs:
...
+                id: composercache
+                run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+            -   name: Cache composer dependencies
+                uses: actions/cache@v2
+                env:
+                    cache-name: cache-composer
+                with:
+                    path: ${{ steps.composercache.outputs.dir }}
+                    key: ${{ hashFiles('composer.json') }}
```
jakzal, line 61: Should we include the PHP version in the cache key?

drupol: Nice catch, if we're using multiple PHP versions to tests, then this is strongly recommended!

stof: here, we're caching the composer files cache, not the vendor folder. So it might be fine to share that cache between PHP versions.

@ciaranmcnulty have you compared the time it takes with Composer 2 ? This would make the job download 1 big archive for the persistent cache, vs downloading several small archives in parallel (with similar network latencies as the packages also come from github servers). the parallelisation of downloads might make this caching useless on the CI server (the cache for packagist metadata is probably useless already as it will likely be outdated for next builds, and they are small files now)

drupol: > here, we're caching the composer files cache, not the vendor folder. So it might be fine to share that cache between PHP versions.

True as well!

ciaranmcnulty:
So @Stof asked about performance.

Before:
Install dependencies: 19s

After:
Cache restore: 1s
Install dependencies: 13s

The speed difference on the install is much less than I'd have expected, TBH. It seems to spend a lot of time extracting separate ZIPs from the cache folder and I wonder if there's a way to speed that up

.github/workflows/build.yml
ciaranmcnulty: Yes I guess the cache folder can store multiple versions of the resolved dependencies. It makes me wonder whether we really do want to invalidate it when composer.json changes?

jakzal: Won't parallel runs overwrite each other's caches?

ciaranmcnulty: No idea! Good question though

jakzal: Looking at the last build, composer re-downloaded 7.2 dependencies. For other PHP versions dependencies were extracted from cache.

PR #1346, opened 2020-11-05
GitHub-formatted actions

ciaranmcnulty:
When running in a Github Action, we can output errors in a specific format so that they show inline in the code

This implements it in a fairly trivial way, I'd like feedback on:

* Should this be automatic based on environment, or opt-in/opt-out via config?
* Should it be in PhpSpec, or an extension?
* Is it something people want?

To-do
- [ ] Escape the messages properly (needs testing/research)
- [ ] Find a way to use the line number of the failing matcher (not present in stack trace currently)

Screenshot 2020-11-05 at 14 36 16

### DIFF
--- a/src/PhpSpec/Console/ContainerAssembler.php
+++ b/src/PhpSpec/Console/ContainerAssembler.php
@@ -606,8 +606,18 @@ final class ContainerAssembler
             } catch (\InvalidArgumentException $e) {
                 throw new \RuntimeException(sprintf('Formatter not recognised: "%s"', $formatterName));
             }
+
             $c->set('event_dispatcher.listeners.formatter', $formatter, ['event_dispatcher.listeners']);
         });
+
+        if (getenv('GITHUB_ACTIONS')) {
+            $container->define('event_dispatcher.listeners.github', function(IndexedServiceContainer $c) {
+                return new SpecFormatter\GithubFormatter(
+                    $c->get('console.io'),
+                    getcwd()
+                );
+            }, ['event_dispatcher.listeners']);
+        }
     }
 
     
--- a/features/bootstrap/ApplicationContext.php
+++ b/features/bootstrap/ApplicationContext.php
@@ -48,6 +48,8 @@ class ApplicationContext implements Context
     {
         StreamWrapper::register();
 
+        putenv('GITHUB_ACTIONS=0');
+
         $this->application = new Application('2.1-dev');
         $this->application->setAutoExit(false);
         $this->setFixedTerminalDimensions();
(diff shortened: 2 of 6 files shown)


drupol:
Wonderful idea :-)

mheki:
👍 for this idea, I'm using phpspec with github actions in several repositories.
Looks handy, but IMO it should not be a part of a core but an extension (maybe included in composer suggest)

ciaranmcnulty:
@mheki I like to keep the core 'pure' but the discoverability would be better than if it was an extension

stof:
>     * Escape the messages properly (needs testing/research)

here is the implementation of the necessary escaping in the official JS toolkit: https://github.com/actions/toolkit/blob/e7eb2c741847a05686a1b8df93dea4b244640b17/packages/core/src/command.ts#L80-L94

iambrosi:
@ciaranmcnulty 

> Should this be automatic based on environment, or opt-in/opt-out via config?

IMO this should be opt-in/opt-out via config. I see it a a new output formatter, and as such configurable via CLI option or config file. 

> Should it be in PhpSpec, or an extension?

I have mixed feelings about this. This formatter would only be useful in a specific running environment, and for this, I would say it should be integrated as an extension. A caveat to this is that it becomes an extra library to be maintained, to keep up to date in projects, and to keep in sync with PhpSpec development.

On the other hand, Github Actions is gaining popularity and is being used more broadly, so allowing developers to just use a built-in formatter for this environment would make things easier for many people.

> Is it something people want?

This is something I would definitively use. I found this kind of messages in Github PRs really helpful.

src/PhpSpec/Console/ContainerAssembler.php
```diff
@@ -606,8 +606,18 @@ function (IndexedServiceContainer $c) {
             } catch (\InvalidArgumentException $e) {
                 throw new \RuntimeException(sprintf('Formatter not recognised: "%s"', $formatterName));
             }
+
             $c->set('event_dispatcher.listeners.formatter', $formatter, ['event_dispatcher.listeners']);
         });
+
+        if (getenv('GITHUB_ACTIONS')) {
```
GrahamCampbell, line 613: The standard place to read environment variables from is actually `$_SERVER`. `getenv` and `putenv` are not threadsafe.

GrahamCampbell (changes requested):

src/PhpSpec/Console/ContainerAssembler.php
jakzal: Surely we're safe on a single-threaded CLI app?

stof: I'm quite sure some crazy devs have created a web runner for phpspec...

jakzal: Even then, surely `GITHUB_ACTIONS` is set to the same value on all threads? ;)

ciaranmcnulty:
Closing, the extension can be found at https://github.com/ciaranmcnulty/phpspec-githubactions

PR #1352, opened 2020-12-15
Handle cases where prophecy reports wrong classname

### LINKED ISSUE #1350: Unknown methods on collaborators don't prompt method generation in some cases
In cases like this, we prompt to add the method signature to the interface:

```php
function it_does_something(Collab $collab)
{
    $collab->someMethod()->willReturn();
}
```

however when a method is called inside the class being specced, we don't catch the error and prompt:

```php

class  MyClassSpec extends ObjectBehaviour
{
    function it_does_something(Collab $collab)
    {
        $this->foo($collab);
    }
}
```
```php
class  MyClass
{
    public function foo(Collab $collab)
    {
         $collab->someMethod();
    }
}
```

I think this may just be the case of handling additional Throwables inside `CollaboratorMethodNotFoundListener`

ciaranmcnulty:
Fixes #1350  on the phpspec side

I have PRed a fix on https://github.com/phpspec/prophecy/pull/514 but this will fix it for older versions

### DIFF
--- a/features/bootstrap/FilesystemContext.php
+++ b/features/bootstrap/FilesystemContext.php
@@ -82,6 +82,7 @@ class FilesystemContext implements Context
 
     /**
      * @Given the class file :file contains:
+     * @Given the interface file :file contains:
      * @Given the trait file :file contains:
      */
     public function theClassOrTraitFileContains($file, PyStringNode $contents)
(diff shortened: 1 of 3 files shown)


ciaranmcnulty:
This is fixed in Prophecy 1.12.2

However, should the hacky fix be merged?

PR #1357, opened 2020-12-31
Remove backward compat layer for older symfony event dispatcher

ciaranmcnulty:
The next major removes older event dispatcher support so this is no longer needed

### DIFF
--- a/src/PhpSpec/Event/BaseEvent.php
+++ b/src/PhpSpec/Event/BaseEvent.php
@@ -2,17 +2,8 @@
 
 namespace PhpSpec\Event;
 
-use Symfony\Component\EventDispatcher\EventDispatcher;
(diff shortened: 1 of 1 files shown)


src/PhpSpec/Event/BaseEvent.php
```diff
@@ -2,17 +2,8 @@
 
 namespace PhpSpec\Event;
 
-use Symfony\Component\EventDispatcher\EventDispatcher;
-use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
-use Symfony\Component\EventDispatcher\Event as OldEvent;
 use Symfony\Contracts\EventDispatcher\Event as ContractEvent;
```
ciaranmcnulty, line 5: Should we include contracts in our composer dependencies?

PR #1360, opened 2021-01-04
Normalize composer.json with ergebnis/composer-normalize

drupol:
Hello,

Recently I saw that the README file of [ergebnis/composer-normalize](https://github.com/ergebnis/composer-normalize) has been updated and it takes the composer.json file of phpspec as example.

This gave me the idea to submit a PR to fix and normalize the composer.json using that tool.

Here it is.

To reproduce:

1. `composer require --dev ergebnis/composer-normalize`
2. `composer normalize`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,73 +1,74 @@
 {
-    "name":         "phpspec/phpspec",
-    "description":  "Specification-oriented BDD framework for PHP 7.1+",
-    "keywords":     ["BDD", "SpecBDD", "TDD", "spec", "specification", "tests", "testing"],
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
I don't see how the effect on VCS traceability is worth having them in sorted order?

drupol:
This is just a way to normalize the `composer.json` file. There is absolutely no impact on the package itself.

ciaranmcnulty:
I'd rather not add to the diffs for something trivial like this; thanks for the effort though

PR #1366, opened 2021-01-13
Psalm level 4 and associated fixes

ciaranmcnulty:


src/PhpSpec/ObjectBehavior.php
```diff
@@ -110,7 +110,7 @@ public function getWrappedObject()
      *
      * @param int|string $key
      */
-    public function offsetExists($key): Subject
+    public function offsetExists($key): bool
```
ciaranmcnulty, line 113: Is this a BC issue? It's clearly wrong...

GrahamCampbell: I'd say it's fine. Often bug fixes are formally BC breaks. So, any BC break that is in itself strictly a bug fix, is totally fine.

drupol: This is indeed a bug, see #1383

Issue #1367, opened 2021-01-16
Matching array of floats


TiMESPLiNTER:
Is there any convenient possibility to match an array of floating point numbers in phpspec 7.x?

Example:

```php
final class TestSpec extends ObjectBehavior
{
    public function it_calculates_points(): void
    {
        $values = [
            627.88,
            679.65,
            766.47,
            498.09,
            1041.99,
            543.06,
            779.02,
            546.83,
            701.85,
            465.8,
            650.84,
            994.04,
            637.77,
            1145.74,
            1741,
            1992,
            804.75,
            1621.28,
            848.64,
        ];

        $expectedResult = [
            669.70166666667,
            708.46,
            731.615,
            741.20166666667,
            705.49666666667,
            649.435,
            623.69666666667,
            588.82833333333,
            654.86166666667,
            732.22166666667,
            843.36666666667,
            1050.3433333333,
            1400.5416666667,
            1569.415,
            1492.63,
            1282.1166666667,
        ];

        $this::calculatePoints($values)->shouldBe($expectedResult);
    }
}
```

This returns: 

> expected [array:11], but got [array:11].

The problem is the proximity of the floats which I can solve with `shouldBeApproximately()`. But then I would have to loop over every array entry and call `shouldBeApproximately()`. That way I loose the nice array diff in case of an error.
Sometimes I'm lucky and the precision doesn't matter and the `shouldBe()` works fine on some floating point arrays.

What's the best practice on that? Could support for a float proximity somehow be added to the `shouldBe()` matcher in case of an array?

Maybe we could add a note in the docs to the `shouldBeApproximately()` section where it tells what the best way is to match floating number arrays.

ciaranmcnulty:
You can write your own matcher, but I wonder if we could support a generic way to apply a matcher to an iterable?

TiMESPLiNTER:
> You can write your own matcher, but I wonder if we could support a generic way to apply a matcher to an iterable?

Exactly, that's basically my question: should phpspec not support a generic way to do this? 

I guess my the matcher would be rather complex if I want to preserve the array diff view on a failed test but only do a approximate match. But I would have to look into it in more detail.

ciaranmcnulty:
Actually for your case we could extend ApproximatelyMatcher to support arrays just by modifying the logic in its matches() method to handle arrays as a special case (and incidentally its supports()) could do more type checking

PR #1368, opened 2021-01-17
Allow ApproximatelyMatcher to compare arrays

ciaranmcnulty:
As suggested in #1367 

My only concern is that the Differ doesn't know the precision the Matcher used, so in some case the verbose diff could be misleading

### DIFF
--- a/features/matchers/developer_uses_approximately_matcher.feature
+++ b/features/matchers/developer_uses_approximately_matcher.feature
@@ -6,36 +6,68 @@ Feature: Developer uses approximately matcher
   @issue581
   Scenario: "Approximately" alias matches using the approximately matcher
     Given the spec file "spec/Matchers/FloatApproximatelyExample1/GeoCoordSpec.php" contains:
-    """
-    <?php
+      """
+      <?php
 
-    namespace spec\Matchers\FloatApproximatelyExample1;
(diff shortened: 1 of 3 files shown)


TiMESPLiNTER:
Regarding your concern: that's definitely not ideal, but the current message you get if you compare an array of floats with `shouldBe()` is misleading anyway. It says:

> expected [array:11], but got [array:11].

But if you see the use of `shouldBeApproximately()` in a test case you're immediately kind of aware the you're dealing with some sort of precision problem, right?

ciaranmcnulty:
@TiMESPLiNTER If you pass the `-v` option you will see a diff

TiMESPLiNTER:
> @TiMESPLiNTER If you pass the `-v` option you will see a diff

True, but not if you compare an array with floats with `shouldBe()`. Then I just got the message mentioned above without a diff. Maybe a bug in the array diff?

ciaranmcnulty:
Odd, maybe open a new issue with a reproducer

Screenshot 2021-01-18 at 10 09 32

drupol:
@TiMESPLiNTER What's the status of this?

TiMESPLiNTER:
Somehow lost track of it and never ended up having this problem again because I never dealt with floating arrays since then. I guess the problem still exists though.

Issue #1375, opened 2021-06-02
New failures since upgrade to PHP 8


drupol:
Hi,

I upgraded to PHP8 and I'm having issues with some tests now. Those issue were not existing in PHP 7.4.19.

See more context here: https://github.com/loophp/collection/discussions/103

I think it's definitely an issue with PHPSpec.
Look, I'm currently working for work and I'm having the same kind of issue, see:

![image](https://user-images.githubusercontent.com/252042/120457489-c9052380-c396-11eb-9a88-320c2f6f06c9.png)

The test can be found here: https://github.com/ecphp/cas-bundle/blob/master/spec/EcPhp/CasBundle/Security/Core/User/CasUserProviderSpec.php#L72

kelunik:
The issue is https://github.com/phpspec/phpspec/blob/eb0d939b235f6091a8439a2a2fc144c9e0d5f569/src/PhpSpec/Runner/Maintainer/ErrorMaintainer.php#L95 not checking properly for error suppression. It needs to use a bit mask as documented on PHP.net, as PHP 8 does no longer suppress fatal errors with @.

kelunik:
See https://www.php.net/manual/en/function.set-error-handler

drupol:
I'll submit a PR in 1 or 2 hours. We need this for work.

PR #1376, opened 2021-06-02
WIP: Fix `error_reporting()` with PHP >= 8

drupol:
This PR:

- [x] Needs #1377 in order to work

Relates to #1375

### DIFF
--- a/spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
+++ b/spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
@@ -18,7 +18,7 @@ class ErrorMaintainerSpec extends ObjectBehavior
         $this->shouldHaveType(ErrorMaintainer::class);
     }
 
-    function it_return_false_when_error_suppresed_or_no_error_reporting()
+    function it_return_false_when_error_suppressed_or_no_error_reporting()
     {
         $oldLevel = error_reporting(0);
         $this->errorHandler(0, 'error message', 'file', 1)->shouldBe(false);
--- a/spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
+++ b/spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
@@ -33,7 +33,8 @@ class ErrorMaintainerSpec extends ObjectBehavior
 
     function it_throws_error_exception_when_message_not_match()
     {
-        $this->shouldThrow(ExampleException\ErrorException::class)
-             ->during('errorHandler', [0, 'error message', 'file', 1]);
+        $this
+            ->shouldThrow(ExampleException\ErrorException::class)
+            ->during('errorHandler', [-1, 'error message', 'file', 1]);
     }
 }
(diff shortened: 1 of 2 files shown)


spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
```diff
@@ -30,10 +30,4 @@ function it_return_true_when_recoverable_level_and_message_match()
         $msg = 'Argument 1 passed to ' . self::class . '::test() must be an instance of string, string given';
         $this->errorHandler(E_RECOVERABLE_ERROR, $msg, 'file', 1)->shouldBe(true);
     }
-
-    function it_throws_error_exception_when_message_not_match()
```
ciaranmcnulty, line 34: Why is this deleted?

drupol: I think this is not needed anymore in PHP 8.

Could be nice to have confirmation from @alexandrugg or @kelunik ?

AlexandruGG: are you saying it's not possible to test the exception path anymore?

AlexandruGG: I think we need a couple of extra tests to show the behaviour we want from this actually

drupol: At least in PHP 8, yes.

AlexandruGG: if I test that with `32767 E_ALL` it does throw the exception in both php 7 and 8:
```PHP
function it_throws_error_exception_when_message_not_match()
    {
        $this->shouldThrow(ExampleException\ErrorException::class)
             ->during('errorHandler', [E_ALL, 'error message', 'file', 1]);
    }
```


this being said I can't claim to fully understand this code change 😅 .

Would it be useful to have a test where the `@` [operator](https://www.php.net/manual/en/language.operators.errorcontrol.php) is used to suppress something that would've caused the exception to be thrown in PHP 8, and now with this change it's no longer thrown? Because that's kind of what we're looking to accomplish with this

kelunik: I don't see a reason to delete any test.

drupol: I deleted the test, maybe I was wrong.

Anyway, despite that, I don't understand the meaning of this test.

If anyone could help, I would be very grateful :)

AlexandruGG: I'll try some stuff so I can better wrap my head around this

ciaranmcnulty:
It'd be good to get more coverage of different error scenarios - maybe in the behat features

spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
AlexandruGG: I opened a separate PR here cause I didn't think I could push to yours: https://github.com/phpspec/phpspec/pull/1378

drupol:
Closing in favor of #1378

PR #1377, opened 2021-06-02
Fix tests by fixing Psalm issues.

drupol:


ciaranmcnulty:
What's the reason for these changes?

drupol:
Psalm is complaining otherwise.

I guess this is because of newer version of psalm, nothing to worry about.

drupol:
@ciaranmcnulty Should I do something else on this ?

drupol:
Thanks!

PR #1378, opened 2021-06-04
Fix `error_reporting()` with PHP >= 8

AlexandruGG:
This PR improves on #1376 with a Behat feature and clearer specs.

Solves #1375.

The code change is based on:
- https://www.php.net/manual/en/function.set-error-handler (see Example 1)
- https://github.com/phpspec/phpspec/issues/1375#issuecomment-852935131
- https://www.php.net/manual/en/language.operators.errorcontrol.php

In a nutshell this is the main issue:
`Warning
Prior to PHP 8.0.0, the value of the severity passed to the custom error handler was always 0 if the diagnostic was suppressed. This is no longer the case as of PHP 8.0.0.`

The code change allows use of the error control operator `@` in PHP 8 without causing PHPSpec to fail.

### DIFF
--- a/features/exception_handling/developer_can_suppress_errors.feature
+++ b/features/exception_handling/developer_can_suppress_errors.feature
@@ -1,0 +1,80 @@
+Feature: Developer uses error control operator "@"
+  As a Developer
+  I want to be able to use the error control operator "@"
+  In order to suppress certain errors and still have my specs passing
+
+  Scenario: An unsuppressed error will cause a valid spec failure
+    Given the spec file "spec/Runner/ErrorSuppression1/ErrorControlSpec.php" contains:
+      """
+      <?php
+
+      namespace spec\Runner\ErrorSuppression1;
+
+      use PhpSpec\ObjectBehavior;
+
+      class ErrorControlSpec extends ObjectBehavior
+      {
+          function it_returns_string()
+          {
+              $this->notSuppressing()->shouldBe('it works!');
+          }
+      }
+
+      """
+    And the class file "src/Runner/ErrorSuppression1/ErrorControl.php" contains:
+      """
+      <?php
+
+      namespace Runner\ErrorSuppression1;
+
+      class ErrorControl
+      {
+          public function notSuppressing(): string
+          {
+              trigger_error('Nope!', E_USER_WARNING);
+
+              return 'it works!';
+          }
+      }
+
+      """
+    When I run phpspec
+    Then I should see "1 broken"
+
+  Scenario: A suppressed error will not cause a spec failure
+    Given the spec file "spec/Runner/ErrorSuppression2/ErrorControlSpec.php" contains:
+      """
+      <?php
+
+      namespace spec\Runner\ErrorSuppression2;
+
+      use PhpSpec\ObjectBehavior;
+
+      class ErrorControlSpec extends ObjectBehavior
+      {
+          function it_returns_string()
+          {
+              $this->suppressing()->shouldBe('it works!');
+          }
+      }
+
+      """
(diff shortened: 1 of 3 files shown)


features/exception_handling/developer_can_suppress_errors.feature
```diff
@@ -0,0 +1,80 @@
...
+          {
+              @trigger_error('Nope!', E_USER_WARNING);
+
+              return 'it works!';
+          }
+      }
+
+      """
+    When I run phpspec
+    Then the suite should pass
```
AlexandruGG, line 80: this Behat scenario will fail in PHP 8 without the code change

spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
```diff
@@ -18,22 +18,51 @@ function it_is_initializable()
...
+
         $msg = 'Argument 1 passed to ' . self::class . '::test() must be an instance of string, string given';
         $this->errorHandler(E_RECOVERABLE_ERROR, $msg, 'file', 1)->shouldBe(true);
+
+        error_reporting($oldLevel);
     }
 
     function it_throws_error_exception_when_message_not_match()
     {
+        $oldLevel = error_reporting(E_ALL);
```
AlexandruGG, line 61: adding these in the specs removes the hidden dependency on the value set in php.ini

AlexandruGG:
hey @ciaranmcnulty, this is ready for review on my side. Let me know if there's anything else you'd like to see on here

drupol:
This PR is indeed what we should do. Thanks for this!
Going to close the other one.

spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
```diff
@@ -18,22 +18,51 @@ function it_is_initializable()
...
+        error_reporting($oldLevel);
+    }
+
+    /**
+     * In PHP 8 error_reporting() will not return 0 anymore when
+     * and error is suppressed with the error control operator "@":
+     *
+     * @link https://www.php.net/manual/en/language.operators.errorcontrol.php
+     */
+    function it_returns_false_when_error_suppressed_in_php_8()
```
kelunik, line 36: No need to have `_in_php_8` in the name, because the same behavior should be present for other versions.

AlexandruGG: I think it's useful to see the distinction because I'm making error reporting return this code, which is specific to php 8:

`error_reporting(E_ERROR | E_PARSE | E_CORE_ERROR | E_COMPILE_ERROR | E_USER_ERROR | E_RECOVERABLE_ERROR)`

The spec above it uses `error_reporting(0)`, which is the behaviour when suppressing errors in php < 8

AlexandruGG:
@ciaranmcnulty hey, could you give this a review please? Cheers

spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
drupol: @kelunik Are you ok with this?

drupol:
Hi all,

It would be nice to have some feedback here.

We had to add [a specific condition](https://github.com/loophp/collection/blob/master/spec/loophp/collection/CollectionSpec.php#L234) in our tests in order to prevent failures with PHP 8.

Is there anything I can do to help, feel free to let me know, I guess @AlexandruGG share the same feelings.

Thanks!

ciaranmcnulty:
Thanks for this @AlexandruGG, it looks great

AlexandruGG:
@ciaranmcnulty thank you for reviewing and getting this merged!

PR #1383, opened 2021-07-05
Fix implementation with arrayaccess

### LINKED ISSUE #1382: Regression: Testing an implementation extending `ArrayAccess` doesn't work since 7.1
Hi,

I'm having issue with some existing tests in [ecphp/cas-bundle](https://github.com/ecphp/cas-bundle) and specifically this [test](https://github.com/ecphp/cas-bundle/blob/master/spec/EcPhp/CasBundle/Configuration/SymfonySpec.php#L19).

With phpspec/phpspec 7.0.1 is works, however, with 7.1.0 is doesn't work anymore and fails:

image

I haven't made the bisect between those two version, but this is something that I'm planning to do, I'm posting this here as a reference for later.

drupol:
This PR:

- [x] Fix #1382


.gitattributes
```diff
@@ -2,3 +2,4 @@
 *.bat  eol=crlf
 
 docs     export-ignore
+/tests   export-ignore
```
ciaranmcnulty, line 5: This should be a separate PR

psalm.xml
```diff
@@ -5,59 +5,12 @@
...
     xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+    errorBaseline="psalm-baseline.xml"
 >
     <projectFiles>
         <directory name="src" />
         <ignoreFiles>
             <directory name="vendor" />
         </ignoreFiles>
     </projectFiles>
-    <issueHandlers>
```
ciaranmcnulty, line 15: I'd rather keep this here with explanatory comments than use a baseline file

spec/tests/spec/PhpSpec/TestWithArrayAccessSpec.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace spec\tests\spec\PhpSpec;
+
+use PhpSpec\ObjectBehavior;
+use tests\spec\PhpSpec\TestWithArrayAccess;
+
+class TestWithArrayAccessSpec extends ObjectBehavior
```
ciaranmcnulty, line 8: Specs are for testing the classes in the codebase

Please move this test into a feature file, and if possible add some tests that use `$this['foo']`-type access to check we didn't break it

src/PhpSpec/ObjectBehavior.php
```diff
@@ -110,7 +110,7 @@ public function getWrappedObject()
      *
      * @param int|string $key
      */
-    public function offsetExists($key): bool
+    public function offsetExists($key): Subject
```
ciaranmcnulty, line 113: This is incompatible with the parent interface, which may become stricter. It's perhaps better to remove the type hint

src/PhpSpec/Wrapper/Subject.php
```diff
@@ -202,9 +202,9 @@ public function getFromWrappedObject(string $property)
     /**
      * @param int|string $key
      */
-    public function offsetExists($key): bool
+    public function offsetExists($key): Subject
```
ciaranmcnulty, line 205: This is incompatible with the parent interface, which may become stricter. It's perhaps better to remove the type hint

ciaranmcnulty (changes requested):

.gitattributes
drupol: Fixed.

psalm.xml
drupol: Fixed.

spec/tests/spec/PhpSpec/TestWithArrayAccessSpec.php
drupol: Fixed.

src/PhpSpec/ObjectBehavior.php
drupol: Fixed.

src/PhpSpec/Wrapper/Subject.php
drupol: Fixed.

drupol:
Dear @ciaranmcnulty ,

This is ready for another round of review.

driesvints:
@drupol no. Look at the discussion in https://github.com/phpspec/phpspec/pull/1397. @ciaranmcnulty is thinking of removing the feature in a next major release.

drupol:
Thanks!

ciaranmcnulty:
I've totally lost track of what we're doing here 😅 

Is 7.2.0 working or do we need to do something?
Is there a BC break we need to make before 8.0.0?

PR #1389, opened 2021-08-01
Relax `require` and `require-dev` dependencies.

drupol:
Hi there,

I tried this morning to test a package of mine against PHP 8.1 and it was not working.

We are currently unable to get PHPSpec working on PHP 8.1 because of constraint in the `composer.json` file.

This PR propose to adopt the Symfony and PHPunit way to constraint PHP in the `composer.json` file.

See for Symfony here: https://github.com/symfony/symfony/pull/36876
See for PHPUnit here: https://packagist.org/packages/phpunit/phpunit#9.5.8

Let me know what you think.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,23 +22,23 @@
     ],
 
     "require": {
-        "php":                      "^7.3 || 8.0.*",
+        "php":                      ">= 7.3",
         "phpspec/prophecy":         "^1.9",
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
I don't want to have any phpspec versions that claim to support versions of PHP or other packages that we haven't tested against

drupol:
Fair enough, thanks!

PR #1391, opened 2021-09-14
Add basic TeamCity formatter

AyrtonRicardo:
To have a better overview about the tests during TeamCity builds.

Extracted from: https://github.com/sebastianbergmann/phpunit/blob/master/src/Logging/TeamCityLogger.php
TeamCity docs: https://www.jetbrains.com/help/teamcity/service-messages.html#Reporting+Messages+for+Build+Log

This also helps to easily identify failed tests and its reasons:
image

**Question**
Are you willing to accept a PR where I add a multi-formatter to phpspec core?
_I would like to be able to output to TeamCity and at the same time output to a JUnit file._

### DIFF
--- a/CHANGES-v7.md
+++ b/CHANGES-v7.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [7.2.0]
+### Added
+- TeamCityFormatter added 
+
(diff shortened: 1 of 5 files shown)


ciaranmcnulty:
Thanks, this looks useful!

PHPUnit is `BSD-3-Clause` so I don't know what the status of 'extracted' code is? It looks fairly dissimilar on a line-by-line basis

@sebastianbergmann what is your opinion here? Is there a way we can resolve this with PhpSpec's MIT licence?

ciaranmcnulty:
Thank you @sebastianbergmann !

ciaranmcnulty:
Thanks @AyrtonRicardo !

PR #1396, opened 2021-10-14
Fix Psalm issues

driesvints:
The int type cast seems redundant here: https://psalm.dev/docs/running_psalm/issues/RedundantCast. This should make most of the build pass (except the PHP 8.1 failures).


driesvints:
@ciaranmcnulty I'm already sending this in separately so we can focus the PHP 8.1 fixes in a separate PR.

PR #1397, opened 2021-10-14
PHP 8.1 fixes

driesvints:


ciaranmcnulty:
Looking again at the RFC, we may be able to just add the `#[ReturnTypeWillChange]` attribute, at least until PHP 9.0

driesvints:
@ciaranmcnulty seems like Psalm is borking. I'm sure the syntax for the attribute is correct.

driesvints:
@ciaranmcnulty can you maybe do that? I've never worked with Psalm myself I'm afraid.

GrahamCampbell:
Psalm also has a baseline functionality like phpstan, if you wanna quickly generate and add to that.

ciaranmcnulty:
@GrahamCampbell I want to keep the exclusions specific

GrahamCampbell:
Can generate it and then move the generated code into the other file. :trollface:

GrahamCampbell:
psalm also has a `@psalm-suppress` annotation you can add in phpdoc. maybe that's preferable here.

ciaranmcnulty:
Hm, there are more issues...

ciaranmcnulty:
I guess those tests all need proper stubs

jaylinski:
I fixed the remaining issues in #1407, please have a look.

PR #1401, opened 2021-11-22
Fix error with unset properties on ThrowMatcher

jmleroux:
The ThrowMatcher could crash when class properties are unset. Alas, the `isInitialized` method only exists since PHP 7.4 and the fix will not work in PHP 7.3.

This behavior was detected after this Symfony PR: https://github.com/symfony/symfony/pull/44037 related to the use of dynamic properties.

### DIFF
--- a/features/exception_handling/developer_specifies_exceptions.feature
+++ b/features/exception_handling/developer_specifies_exceptions.feature
@@ -45,3 +45,99 @@ Feature: Developer specifies exception behaviour
       """
     When I run phpspec
     Then the suite should pass
+
+  Scenario: Throwing an exception with unset properties
+    Given the spec file "spec/Runner/ExceptionExample/UnsetPropertyClassSpec.php" contains:
+      """
+      <?php
+
+      namespace spec\Runner\ExceptionExample;
+
+      use PhpSpec\ObjectBehavior;
+
+      class UnsetPropertyClassSpec extends ObjectBehavior
+      {
+          function it_throws_an_exception_with_unset_property()
+          {
+              $this->shouldThrow(new \Runner\ExceptionExample\ChildException('exception message'))
+                  ->duringDoSomething();
+          }
+      }
+
+      """
+    And the class file "src/Runner/ExceptionExample/UnsetPropertyClass.php" contains:
+      """
+      <?php
+
+      namespace Runner\ExceptionExample;
+
+      class UnsetPropertyClass
+      {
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Is there a test that could accompany this?

jmleroux:
> Is there a test that could accompany this?

:+1: 

I'll try to add one, I was waiting for the Travis CI activation.

jmleroux:
Would you accept the addition of a Dockerfile and/or docker-compose file in this PR? 

It could facilitate the tests on different PHP versions for example

ciaranmcnulty:
I'd rather keep a dockerfile out of it; we manage ok on Github Actions testing different versions and don't want things in two places

jmleroux:
FYI, I cannot yet reproduce the case with a simple test when I still can reproduce it on Akeneo.
Work in progress.

jmleroux:
I added a Behat scenario to test the modification.
For a reason I still not identify, the previous ThrowMatcher will still pass the test in PHP 7.4, but will fail on PHP 8.0. PHP 8.0 may be more strict with uninitialized reflection values.
That's why it took me so much time to finalize this PR because I was trying to validate my test with PHP7.4

jmleroux:
Hello,
Is there a problem with my PR?
I'd like to see it running on the real CI  and waiting for it because i'm stuck with my specs on one private project after migrating to PHP 8.0 :wink:

shouze:
> Hello,
> Is there a problem with my PR?
> I'd like to see it running on the real CI and waiting for it because i'm stuck with my specs on one private project after migrating to PHP 8.0 😉

I would like too ❤️

jmleroux:
> I would like too

Hello @shouze ,
Did you have the opportunity to validate it on your side?

jmleroux:
Hello guys,
Is there a problem with this PR or is this inactivity normal?

ciaranmcnulty:
@jmleroux no problem with the PR, looks good thank you

ciaranmcnulty:
Weirdly this is failing on main for the older PHP versions... taking a look

jmleroux:
Weird, I think I tested it. I even added a dummy PR to test without the fix: https://github.com/phpspec/phpspec/pull/1409
Do not hesitate to ping me if I can help.

Note: I don't have several PHP version locally and didn't have access to Travis for this PR, so I tested with several docker files.

PR #1402, opened 2021-12-01
Symfony 6 support (without Flex dependency)

loic425:


ciaranmcnulty:
These Psalm issues may have been resolved in the #1397 branch, and that commit could be cherry-picked

loic425:
@ciaranmcnulty I was looking the errors on the CI and I was wondering if you want to keep the support of php < 8.0.
Cause Symfony 6 has dropped support for PHP 7.4 and you override some Symfony console methods.

webignition:
Symfony 6 supports PHP >= 8.0.2, support for Symfony 6 requires dropping support for PHP < 8.0.2. 

It's not really a question of whether you want to drop support for PHP < 8 but more of a realisation that this is, as far as I understand, dropping support for PHP < 8 is an unavoidable requirement.

In order to support Symfony 6, is there any other choice other than to drop support for PHP < 8?

loic425:
@webignition Yes I think this is ok with keeping PHP 7 support. I think I was wrong with the issue encountered.

loic425:
@ciaranmcnulty Could you approve running the CI build please?

src/PhpSpec/Console/Command/RunCommand.php
```diff
@@ -31,7 +32,7 @@
  */
 final class RunCommand extends Command
 {
-    public function getApplication() : Application
+    public function getApplication() : ?BaseApplication
```
webignition, line 35: This method needs to return an instance of `PhpSpec\Console\Application` instead of `Symfony\Component\Console\Application` due to needing to subsequently call `getContainer()` which is a method of the former but not a method of the latter.

Is this change necessary? I'm not saying it isn't, just curious given the impact this change makes.

webignition: I've tried this out in [my own attempt to get Symfony 6.0 working](https://github.com/webignition/phpspec) and I can't see the need for this.

loic425: Maybe I have to try again without these change...
But this is a different return typehint than the Symfony one and I'm not sure it can work.
Maybe just an assertion can be enough on phpspec code.

loic425:
CI should be fixed before with [that PR](#1407) and [Psalm error too](#1408)

src/PhpSpec/Console/Command/RunCommand.php
webignition: Returning `PhpSpec\Console\Application` not only works but is necessary for the command to call `PhpSpec\Console\Application::getContainer()`.

You are correct in saying that the return type is different. An instance of `PhpSpec\Console\Application` is acceptable to return here as it is a child type of (i.e. extends) `Symfony\Component\Console\Application`. Technically speaking, this is a covariant return type and support was introduced in PHP 7.4.

loic425: Yes you're right. But "we" have to drop support for PHP 7.3 then.

loic425: And I think we can drop support for Symfony 3 too... It is not maintained anymore.

loic425: @ciaranmcnulty Are you ok to drop both PHP 7.3 and Symfony 3 support?

webignition: The PHP 7.3 concern is a valid point that I completely failed to consider and that does complicate things somewhat. But with PHP 7.3 now being EOL, I don't think it's fully unwise to consider dropping support.

shouze:
@webignition @ciaranmcnulty do you think guys this PR can get merged soon, I can help if any other blocker :angel:

loic425:
@webignition I would add the [PHP 8.1 support PR](#1406 ) between 1 and 2 step

loic425:
@ciaranmcnulty I will work on this PR this week now you merged the other needed PRs. Thx a lot

loic425:
@ciaranmcnulty Do you really want maintain PHP 7.3 support?
Please look at this thread https://github.com/phpspec/phpspec/pull/1402#discussion_r773738908.

loic425:
@ciaranmcnulty Could you run the CI on this one please?
Let's move forward an bring that support to whole Symfony community.

ciaranmcnulty:
I don't mind dropping 7.3 support, it'll just mean a new major (only because that's the convention we established) - it should drop unmaintained Symfony too

loic425:
> I don't mind dropping 7.3 support, it'll just mean a new major (only because that's the convention we established) - it should drop unmaintained Symfony too

Great, in this current PR? CI is green for other PHP versions.

ciaranmcnulty:
I've updated the `next` branch to reflect the minimum PHP and Symfony versions we'll support in 8.0.0. 

Can you target this PR against that branch please?

ciaranmcnulty:
I got another chance to think about this; I think we should support PHP 7.3. The next major will be 8.0+ and that will leave Symfony users who want to use 6.0 but are stuck on PHP 7.4 without a version to use.

(commented inline in code how we can support 7.3)

src/PhpSpec/Console/Command/RunCommand.php
ciaranmcnulty: I think we should keep this returning `Application`, then do a runtime assertion in the code that it's a `BaseApplication` we got back

loic425: I think we can keep the existing code code, cause on PHP lower than 8.0 we have not symfony 6 downloaded so this still works with no return type on symfony lower than 6.

loic425:
> I got another chance to think about this; I think we should support PHP 7.3. The next major will be 8.0+ and that will leave Symfony users who want to use SF 6.0 but are stuck on PHP 7.4 without a version to use.
> 
> (commented inline in code how we can support 7.3)

Do you have an Idea to solve the last error. It seems to need some investigation but this is not related to this PR imho.

ciaranmcnulty:
> Do you have an Idea to solve the last error. It seems to need some investigation but this is not related to this PR imho.

It's fixed on main, please rebase

loic425:
> It's fixed on main, please rebase

done

ciaranmcnulty:
Thank you, I plan to do a release of PhpSpec 7 and then 8 (from `next` branch)

loic425:
Thank you too @ciaranmcnulty 🎉

PR #1407, opened 2021-12-15
Fix PHP 8.1 errors

jaylinski:
Fixes PHP 8.1 errors in CI pipeline.

Closes #1397.


src/PhpSpec/ObjectBehavior.php
```diff
@@ -130,6 +130,7 @@ public function offsetGet($key): Subject
      *
      * @param int|string $key
      */
+    #[\ReturnTypeWillChange]
```
jaylinski, line 133: As mentioned in https://github.com/phpspec/phpspec/pull/1397#issuecomment-944081760, this comment was added instead of the `void` return type in order to preserve backwards compatibility.

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -198,7 +198,9 @@ private function extractTypehints(array &$tokens, int $index, array $token): voi
 
     private function haveNotReachedEndOfTypeHint($token) : bool
     {
-        if ($token == '|' || $token == '&') {
+        // PHP 8.1 returns the intersection token `&` as an array,
+        // while previous versions return it as a string.
+        if ($token == '|' || $token == '&' || (is_array($token) && $token[1] == '&')) {
```
jaylinski, line 203: This will break if PHP returns a token-array with only one item. Should I check for it, or can I count on the array to always have a `1`-index? :thinking:

webignition: I'd say that the explicit check isn't hard to add but might make the code somewhat harder to understand:

```php
if ($token == '|' || $token == '&' || (is_array($token) && ($token[1] ?? '') == '&')) {
    //...
}
```

It could be more readable by separating out the expression into independent booleans:
```php
$isFoo = is_string($token) && $token == '|' || $token == '&';
$isBar = is_array($token) && ($token[1] ?? '') == '&';

if ($isFoo || $isBar) {
    return false;
}
```

Best of luck with naming `$isFoo` and `$isBar` correctly ...

shouze:
> @ciaranmcnulty If I can help you to review this PR in any way, please let me know.

FYI I've forked and applied all the patches in the suggested order, and I can confirm that by using this fork we've been able to finish our Symfony 6 migration and to upgrade from php 8 to 8.1. Everything works as expected since few days :ok_hand: 

https://github.com/gogaille/phpspec/tree/php-81-and-symfony-6

lchrusciel:
We can also confirm that it works - https://github.com/loic425/Sylius/runs/5037334478 (thanks @loic425 for branch).

Zales0123:
Hello, it's a friendly 2-weeks reminder about this PR 😄 👋  @ciaranmcnulty do you think it's possible to be merged in the predictable future? Maybe someone else from PHPSpec Organization has some info? @MarcelloDuarte @stof? 🚀

PR #1411, opened 2022-02-17
Fixed support for PHP 8.1

iambrosi:
What's included:

* Added missing typehints to classes that implement `ArrayAccess`
* Ensure `preg_match` never receives a `null` subject
* Take into account ampersands used for intersection types when rewriting tokens (PHP 8.1 only)
* Added missing returns to collaborators in specs

### DIFF
--- a/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
+++ b/spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
@@ -5,7 +5,7 @@ namespace spec\PhpSpec\Locator\PSR0;
 use PhpSpec\ObjectBehavior;
 use PhpSpec\Util\Filesystem;
 
-use SplFileInfo;
+use Symfony\Component\Finder\SplFileInfo;
 
 class PSR0LocatorSpec extends ObjectBehavior
 {
--- a/src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
+++ b/src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
@@ -52,6 +52,10 @@ final class TokenizedTypeHintRewriter implements TypeHintRewriter
             $this->typehintTokens[] = T_NAME_FULLY_QUALIFIED;
             $this->typehintTokens[] = T_NAME_QUALIFIED;
         }
+
+        if (\PHP_VERSION_ID >= 80100) {
+            $this->typehintTokens[] = T_AMPERSAND_NOT_FOLLOWED_BY_VAR_OR_VARARG;
+        }
     }
 
     public function rewrite(string $classDefinition): string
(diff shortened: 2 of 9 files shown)


src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
```diff
@@ -52,6 +52,10 @@ public function __construct(TypeHintIndex $typeHintIndex, NamespaceResolver $nam
             $this->typehintTokens[] = T_NAME_FULLY_QUALIFIED;
             $this->typehintTokens[] = T_NAME_QUALIFIED;
         }
+
+        if (\PHP_VERSION_ID >= 80100) {
+            $this->typehintTokens[] = T_AMPERSAND_NOT_FOLLOWED_BY_VAR_OR_VARARG;
```
stof, line 57: if we need to add the `&` here, don't we also need to add the `|` for union types on PHP 8+ ?

spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
```diff
@@ -5,7 +5,7 @@
 use PhpSpec\ObjectBehavior;
 use PhpSpec\Util\Filesystem;
 
-use SplFileInfo;
+use Symfony\Component\Finder\SplFileInfo;
```
stof, line 8: why this change ?

src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php
iambrosi: This is already handled in the line below, as `token_get_all` returns `|` instead of a three element array for this. Please, check the position 7 of the array in this output https://3v4l.org/vkq1E#v8.0.16

https://github.com/phpspec/phpspec/blob/ec649c1b79c7e6041fbefa56e8998260a85cd40c/src/PhpSpec/CodeAnalysis/TokenizedTypeHintRewriter.php#L205

spec/PhpSpec/Locator/PSR0/PSR0LocatorSpec.php
iambrosi: In PHP 8.1, the following error started to make the test suite fail, which I think is related to https://github.com/phpspec/prophecy/issues/414:

```
PhpSpec/Locator/PSR0/PSR0Locator
 137  - it finds all resources from tracked specPath
      8192: SplFileInfo::__construct(): Passing null to parameter #1 ($filename) of type string is deprecated in
      /phpspec/vendor/phpspec/prophecy/src/Prophecy/Doubler/Generator/ClassCreator.php(49) : eval()'d code line 6
```

Given that [Filesystem](https://github.com/phpspec/phpspec/blob/b6ef0b627f212910c6e3dfd5b9ad2630b778fff4/src/PhpSpec/Util/Filesystem.php#L55-L61) class already leverages Symfony's Finder to find for spec files in a path, which return instances of `Symfony\Component\Finder\SplFileInfo`, I figured it wouldn't affect the testeability of this class to rely on Symfony's own implementation of `SplFileInfo`.

iambrosi:
Ping @ciaranmcnulty to know his thoughts on this PR.

ciaranmcnulty:
These changes look ok to me - like you say we already use symfony/filesystem

ciaranmcnulty:
Thanks @iambrosi !

PR #1416, opened 2022-03-09
[Fix] Do not break on deprecations

loic425:
```php
<?php

declare(strict_types=1);

namespace App;

final class User implements \Serializable
{
    public function serialize(): string
    {
        return '';
    }

    public function unserialize($data): void
    {
    }
}

```

```php
<?php

namespace spec\App;

use App\User;
use PhpSpec\ObjectBehavior;

class UserSpec extends ObjectBehavior
{
    {
        $this->shouldHaveType(User::class);
    }
}

```

Capture d’écran 2022-03-09 à 10 18 56

### DIFF
--- a/spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
+++ b/spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
@@ -18,7 +18,7 @@ class ErrorMaintainerSpec extends ObjectBehavior
         $this->shouldHaveType(ErrorMaintainer::class);
     }
 
-    function it_returns_false_when_error_suppresed_or_no_error_reporting()
+    function it_returns_false_when_error_suppressed_or_no_error_reporting()
     {
         $oldLevel = error_reporting(0);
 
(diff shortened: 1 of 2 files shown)


spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
```diff
@@ -46,6 +46,19 @@ function it_returns_false_when_error_suppressed_in_php_8()
...
+    {
+        /**
+         * Based on manual testing and:
+         * @link https://www.php.net/manual/en/language.operators.errorcontrol.php#125938
+         */
+        $oldLevel = error_reporting(E_DEPRECATED);
+
+        $this->errorHandler(E_DEPRECATED, 'error message', 'file', 1)->shouldBe(false);
+
+        error_reporting($oldLevel);
```
GrahamCampbell, line 59: you you need to put this in a try-finally.

GrahamCampbell (changes requested):

spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php
loic425: what do you mean? we are on the spec file here and this is the same way as on the method just above.

https://github.com/phpspec/phpspec/blob/main/spec/PhpSpec/Runner/Maintainer/ErrorMaintainerSpec.php#L44

ciaranmcnulty:
This is a pretty opinionated change, how could a user opt out?

Currently you can set `runner.maintainers.errors.level` in the yaml (which is a bit messy but at least works)

loic425:
> Currently you can set `runner.maintainers.errors.level` in the yaml (which is a bit messy but at least works)

I didn't know that. Yes it works with these following configuration

```yaml
# phpspec.yaml
runner.maintainers.errors.level: 22527 # E_ALL ^ E_STRICT ^ E_DEPRECATED
```

It's on on documentation here:
http://www.phpspec.net/en/stable/cookbook/configuration.html#options

But what do you think about changing that on the ContainerAssembler:
```php
$c->getParam('runner.maintainers.errors.level', E_ALL ^ E_STRICT ^ E_DEPRECATED)
```

Cause I don't think we need an error for deprecations by default.

ciaranmcnulty:
It sort of 'accidentally' works because it's a parameter; it's a little verbose for the config file - I'd be happy changing the default (in next major)

ciaranmcnulty:
What is a sensible default value for this in 8.0.0? Remove E_DEPRECATED and E_USER_DEPRECATED?

MarcelloDuarte:
Closing because it is very old and a workaround was presented in https://github.com/phpspec/phpspec/pull/1416#issuecomment-1062990482

PR #1423, opened 2022-03-16
Psalm level 2 changes

ciaranmcnulty:
Definitely includes some BC breaks so aimed at `next`


src/PhpSpec/Wrapper/Subject/Caller.php
```diff
@@ -137,7 +137,8 @@ public function getWrappedObject(): mixed
             return $this->wrappedObject->getInstance();
         }
 
-        if (!$className || !class_exists($className)) {
+        /** @var string $className don't trust it's a class-string, we still want to check */
```
GrahamCampbell, line 140: I don't think this is needed. `class-string` need not be the name of a class. it could be an interface, in which case the class_exists check would fail.

GrahamCampbell (changes requested):

src/PhpSpec/Wrapper/Subject/Caller.php
ciaranmcnulty: I'm a little unsure whether it should be class-string at all, as in some cases it's provided by the user

ciaranmcnulty: That is I think it's provided by the user but we force it to class-string :/ Will address later

PR #1438, opened 2022-12-13
Psalm 5 Support

rogervila:
Hello,

This PR adds Psalm 5 Support.
It also fixes some Psalm errors necessary to get green, but there are other issues remaining.

PR #1439, opened 2022-12-14
Fix psalm issues on new branch

ciaranmcnulty:
Merges #1438 into the `next` branch, and fixes some extra issues

### DIFF
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -70,7 +70,10 @@ final class Application extends BaseApplication
             $this->add($command);
         }
 
-        $this->setDispatcher($this->container->get('console_event_dispatcher'));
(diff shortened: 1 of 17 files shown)


ciaranmcnulty:
@rogervila  can you review maybe?

src/PhpSpec/Console/Application.php
```diff
@@ -120,6 +123,7 @@ protected function getDefaultInputDefinition(): InputDefinition
 
     /**
      * @throws \RuntimeException
+     * @return void
```
rogervila, line 126: the return definition on DocBlock is redundant here

rogervila (changes requested):

src/PhpSpec/Console/Application.php
rogervila: A part of that everything looks good @ciaranmcnulty :)

PR #1440, opened 2022-12-15
Refactoring for PHP 8.0

ciaranmcnulty:
This uses rector to do a bunch of refactoring

The only deliberate potential BC break is the addition of visibility to class constants, hence this targets the `next` branch

### DIFF
--- a/src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
+++ b/src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
@@ -3,6 +3,7 @@
 namespace PhpSpec\Matcher\Iterate;
 
 use PhpSpec\Formatter\Presenter\Presenter;
+use function is_iterable;
 
 final class IterablesMatcher
 {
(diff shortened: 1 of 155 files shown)


src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
```diff
@@ -3,14 +3,13 @@
 namespace PhpSpec\Matcher\Iterate;
 
 use PhpSpec\Formatter\Presenter\Presenter;
+use function is_iterable;
+use function is_iterable as is_iterable1;
```
jdreesen, line 7: This is strange.

src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
```diff
@@ -21,11 +20,11 @@ public function __construct(Presenter $presenter)
      */
     public function match(mixed $subject, mixed $expected, bool $strict = true): void
     {
-        if (!$this->isIterable($subject)) {
+        if (!is_iterable1($subject)) {
```
jdreesen, line 23:
```suggestion
        if (!is_iterable($subject)) {
```

src/PhpSpec/Matcher/Iterate/IterablesMatcher.php
ciaranmcnulty: Ah I didn't spot that, thanks!

PR #1443, opened 2023-02-03
Allow sebastian/export v5

arendjantetteroo:
v5 is required for phpunit 10, which phpspec now blocks.  v5 drops support for PHP 7.3, PHP 7.4 and PHP 8.0 but has no other BC breaks so should work with phpspec.

PR #1444, opened 2023-02-23
Allow PHPUnit 10

Jean85:
This is a continuation of #1443 to try and resolve the blockers, /cc @ciaranmcnulty

I expect the CI to have issues due to https://github.com/ramsey/composer-install/issues/241. I also have locally detected some failures, but they are only on test code:

In Behat:
```
001 Scenario: Generating a spec for class with namespace containing reserved keyword # features/code_generation/developer_generates_spec.feature:184
      And there should be no file "spec/Namespace/ClassExample1/MarkdownSpec.php"    # features/code_generation/developer_generates_spec.feature:187
        Fatal error: Call to undefined method PHPUnit\Framework\Assert::assertFileNotExists() (Behat\Testwork\Call\Exception\FatalThrowableError)
```
In PHPSpec:
```
PhpSpec/Formatter/Presenter/Differ/ObjectEngine                                                                                                                                                            
  17  - it is a differ engine                                                                                                                                                                              
      could not reflect class SebastianBergmann\Exporter\Exporter as it is marked final.                                                                                                                   
                                                                                                                                                                                                           
PhpSpec/Formatter/Presenter/Differ/ObjectEngine                                                                                                                                                            
  22  - it does not support scalars                                                                                                                                                                        
      could not reflect class SebastianBergmann\Exporter\Exporter as it is marked final.                                                                                                                   
                                                                                                                                                                                                           
PhpSpec/Formatter/Presenter/Differ/ObjectEngine                                                                                                                                                            
  27  - it only supports objects                                                                                                                                                                           
      could not reflect class SebastianBergmann\Exporter\Exporter as it is marked final.                                                                                                                   
                                                                                                                                                                                                           
PhpSpec/Formatter/Presenter/Differ/ObjectEngine                                                                                                                                                            
  32  - it converts objects to string and diffs the result                                                                                                                                                 
      could not reflect class SebastianBergmann\Exporter\Exporter as it is marked final.                                                                                                                   
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
     "require-dev": {
         "behat/behat":           "^3.3",
         "symfony/filesystem":    "^3.4 || ^4.0 || ^5.0 || ^6.0",
-        "phpunit/phpunit":       "^8.0 || ^9.0",
+        "phpunit/phpunit":       "^8.0 || ^9.0 || ^10.0",
         "vimeo/psalm": "^4.3 || ^5.2"
     },
 
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php":                      "^7.3 || 8.0.* || 8.1.* || 8.2.*",
         "phpspec/prophecy":         "^1.9",
         "phpspec/php-diff":         "^1.0.0",
-        "sebastian/exporter":       "^3.0 || ^4.0",
+        "sebastian/exporter":       "^3.0 || ^4.0 || ^5.0",
         "symfony/console":          "^3.4 || ^4.4 || ^5.0 || ^6.0",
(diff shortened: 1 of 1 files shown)


Jean85:
Ehi @ciaranmcnulty, how do you want to proceed? We have 3 PRs in a row, in a deadlock due to circular dependency_
 * this one
 * https://github.com/phpspec/prophecy/pull/585
 * https://github.com/phpspec/prophecy-phpunit/pull/45

As I said in https://github.com/phpspec/phpspec/pull/1443#issuecomment-1445498910, your proposal of an "allow failure" job is not possible. Would you be ok with a job that forcibly required PHPUnit 10 and fails on all 3 repos, and subsequent PRs to fix those failures?

Issue #1431, opened 2022-06-22
ObjectStateMatcher prevails over ScalarMatcher since PHP 8.0


bsod85:
As pointed out in issue https://github.com/phpspec/phpspec/issues/1403 the matchers order changed since PHP 8.0.

In our case a problem arises because ObjectStateMatcher takes precedence now over ScalarMatcher when we use `$object->shouldNotBeNull()`.

Then test breaks with the error `method [array:2] not found.` and phpspec prompts us to create a method with the message:
```Do you want me to create                                                      
  `OurClass::isNull()` for you?```
  
Our solution was simply to increment the ScalarMatcher priority to 51. If this is the best solution I can open a PR, otherwise depending of what the solution is I'l be glad if I can be of any help.

Lehren:
+1, very annoying suddenly

ciaranmcnulty:
Yes please do open a PR if they're currently equal priority and PHP's sort order has changed.

PR #1447, opened 2023-05-02
Restore ScalarMatcher's priority order for PHP 8.0+

bsod85:
Fixes issue https://github.com/phpspec/phpspec/issues/1431

We already use this change with a composer patch since June 2022 without any side effect


newscientistgeeks:
hi, any blocker on merging this one? otherwise it would be great, breaks simple checks for array collections etc on php8.2

ciaranmcnulty:
Thanks @bsod85!

PR #1463, opened 2024-01-23
Add stricter argument, property, and return types

ciaranmcnulty:
Used existing docblocks to populate types, and added obvious narrower types where there was no docblock. Used wide types where it was not clear what the real types should be.

Did not attempt to fix types in this PR, just to tried to ensure that everywhere has a declared type of some sort.

### DIFF
--- a/src/PhpSpec/Wrapper/Subject/WrappedObject.php
+++ b/src/PhpSpec/Wrapper/Subject/WrappedObject.php
@@ -13,6 +13,7 @@
 
 namespace PhpSpec\Wrapper\Subject;
 
+use Closure;
 use PhpSpec\Factory\ObjectFactory;
 use PhpSpec\Formatter\Presenter\Presenter;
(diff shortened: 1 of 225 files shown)


src/PhpSpec/Wrapper/Subject/WrappedObject.php
```diff
@@ -89,10 +74,7 @@ public function beConstructedWith(array $args): void
         $this->beAnInstanceOf($this->classname, $args);
     }
 
-    /**
-     * @param null|callable|string $factoryMethod
-     */
-    public function beConstructedThrough($factoryMethod, array $arguments = array()): void
+    public function beConstructedThrough(callable|array|string|null $factoryMethod, array $arguments = array()): void
```
stof, line 77: why adding `array` here, when it was not in the phpdoc type (neither for the argument nor for the property) ?

PR #1466, opened 2024-01-26
Add Token utility object

ciaranmcnulty:
While trying to make the code using tokens clearer I found myself adding a lot of types and spotting repeated logic

This PR introduces some small utility value objects that encapsulate this repetitive token handling

### DIFF
--- a/src/PhpSpec/CodeAnalysis/TokenizedNamespaceResolver.php
+++ b/src/PhpSpec/CodeAnalysis/TokenizedNamespaceResolver.php
@@ -38,7 +38,7 @@ final class TokenizedNamespaceResolver implements NamespaceResolver
 
         $tokens = Token::getAll($code);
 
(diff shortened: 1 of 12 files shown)


src/PhpSpec/CodeAnalysis/TokenizedNamespaceResolver.php
```diff
@@ -34,56 +36,59 @@ public function analyse(string $code): void
         $this->currentUseGroup = '';
         $this->uses = [];
 
-        $tokens = token_get_all($code);
+        $tokens = Token::getAll($code);
 
-        foreach ($tokens as $index => $token) {
+        foreach ($tokens as $_ => $token) {
```
jdreesen, line 41: Why not just:

```suggestion
        foreach ($tokens as $token) {
```

PR #1467, opened 2024-01-31
Exclude service container calls from type checks as it's currently no…

ciaranmcnulty:
…n-type-safe


ciaranmcnulty:
Ah, this opens the door to all sorts of other errors currently excluded by psalm level 3 :/

PR #1472, opened 2024-08-21
PHP 8.4 | Fix implicitly nullable parameters (7.x)

jrfnl:
PHP 8.4 deprecates implicitly nullable parameters, i.e. typed parameter with a `null` default value, which are not explicitly declared as nullable.

This commit fixes all found instances of that in the PHPSpec codebase.

Includes updating the documentation to match (where relevant, i.e. only existing documentation has been touched).

Ref: https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

### DIFF
--- a/src/PhpSpec/Console/ConsoleIO.php
+++ b/src/PhpSpec/Console/ConsoleIO.php
@@ -111,12 +111,12 @@ class ConsoleIO implements IO
         return $this->lastMessage;
     }
 
-    public function writeln(string $message = '', int $indent = null): void
+    public function writeln(string $message = '', ?int $indent = null): void
     {
         $this->write($message, $indent, true);
     }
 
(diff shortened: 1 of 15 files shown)


stof (approved):

andypost:
It needs to get rid of `E_STRICT` and the branch 7.5 could declare compatibility https://wiki.php.net/rfc/deprecations_php_8_4#remove_e_strict_error_level_and_deprecate_e_strict_constant

```
PHP Deprecated:  Constant E_STRICT is deprecated in /mnt/vendor/phpspec/phpspec/src/PhpSpec/Console/ContainerAssembler.php on line 647

Deprecated: Constant E_STRICT is deprecated in /mnt/vendor/phpspec/phpspec/src/PhpSpec/Console/ContainerAssembler.php on line 647
```

Issue #1470, opened 2024-08-07
Documentation site is down


ash-m:
https://www.phpspec.net is reporting:

> # Error 1014
> Ray ID: 8af9be107bf87c1c • 2024-08-07 19:41:08 UTC
> 
> ## CNAME Cross-User Banned
> 
> What happened?
> You've requested a page on a website that is part of the [Cloudflare](https://www.cloudflare.com/5xx-error-landing/) network. The host is configured as a CNAME across accounts on Cloudflare, which is not allowed by Cloudflare's security policy.
> 
> What can I do?
> If this is an R2 custom domain, it may still be initializing. If you have attempted to manually point a CNAME DNS record to your R2 bucket, you must do it using a custom domain. Refer to [R2's documentation](https://developers.cloudflare.com/r2/data-access/public-buckets/#connect-your-bucket-to-a-custom-domain) for details.
> 
> [Visit our website](https://www.cloudflare.com/5xx-error-landing/) to learn more about Cloudflare.

ash-m:
Not sure who is administering the DNS for this, maybe @mdavis1982 (no idea, but authoritative origin has 'matt' in it :P)

Can someone look into this please? @ciaranmcnulty @MarcelloDuarte 

Thanks

mdavis1982:
@ash-m It's not me, I'm afraid. I think @ciaranmcnulty was last managing the project, but also @MarcelloDuarte may know!

jdreesen:
It works without the `www.`, see: #1475

ash-m:
@jdreesen I see; weirdly, in-private browsing redirects `phpspec.net/` (with a trailing slash) to `www`. I guess it's just a caching problem. 

Thank you for the update.

PR #1475, opened 2024-08-28
Fix documenation link

BackEndTea:
The `www` domain gave the following error. It might also be nice to just redirect the `www` domain to the non www version. And while i was at it i updated the link to be https

image

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,7 +1,7 @@
 phpspec
 =======
 
-The main website with documentation is at `http://www.phpspec.net <http://www.phpspec.net>`_.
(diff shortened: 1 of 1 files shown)


jdreesen:
Fixes #1470

W0rma (approved):

chadyred:
Hi @BackEndTea as it is resolved, can you close you PR, and you will be the best :)

jdreesen:
@ciaranmcnulty the URL in the repo description is still with `www.` and doesn't work:
grafik

PR #1483, opened 2024-11-26
No ticket. Support PHP Unit 11.

evs-xsarus:
With these changes, phpspec will install next to PHP Unit 11.

PR #1484, opened 2025-02-07
Allow Psalm 6

Jean85:
Unlocks #1483 

[EDIT] I preferred to avoid any code changes and to put all new reported issues in the Psalm baseline. Future PRs can address those.


Jean85:
@stof can we proceed with this? It unlocks #1483, which in turn unlocks PHPUnit 12 for Prophecy: https://github.com/phpspec/prophecy-phpunit/pull/67

PR #1488, opened 2025-04-29
[7.x] Allow PHPUnit 11

Jean85:


Jean85:
Thanks to , PHPUnit 11 is actually running in CI! 🎉 

https://github.com/phpspec/phpspec/actions/runs/14753370765/job/41415748000?pr=1489#step:11:10

I'll just drop an unneeded commit and this is good to merge!

PR #1491, opened 2025-05-03
Upmerge 7.x into main

Jean85:
This is as agreed with @MarcelloDuarte

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,10 +25,6 @@ jobs:
                         publish-phar: true
                     -   php: "8.2"
                         composer-flags: --prefer-lowest
-                    -   php: "8.3"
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -25,10 +25,6 @@ jobs:
                         publish-phar: true
                     -   php: "8.2"
                         composer-flags: --prefer-lowest
-                    -   php: "8.3"
-                        operating-system: windows-latest
-                    -   php: "8.2"
-                    -   php: "8.3"
```
Jean85, line 31: These are unneeded, the normal matrix already tests for these combinations.

MarcelloDuarte:
Thanks, @Jean85 👍

PR #1502, opened 2025-11-14
Add missing callable type for Subject::beConstructedThrough

Kozzi11:
FIX issue https://github.com/phpspec/phpspec/issues/1501

### DIFF
--- a/src/PhpSpec/Wrapper/Subject.php
+++ b/src/PhpSpec/Wrapper/Subject.php
@@ -131,7 +131,7 @@ class Subject implements ArrayAccess, ObjectWrapper
         $this->wrappedObject->beConstructedWith(\func_get_args());
     }
 
-    public function beConstructedThrough(string|array $factoryMethod, array $arguments = array()): void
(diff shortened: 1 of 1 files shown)


src/PhpSpec/Wrapper/Subject.php
```diff
@@ -131,7 +131,7 @@ public function beConstructedWith(): void
         $this->wrappedObject->beConstructedWith(\func_get_args());
     }
 
-    public function beConstructedThrough(string|array $factoryMethod, array $arguments = array()): void
+    public function beConstructedThrough(string|array|callable $factoryMethod, array $arguments = array()): void
```
stof, line 134:
```suggestion
    public function beConstructedThrough(string|callable $factoryMethod, array $arguments = array()): void
```

the only arrays that are supported in WrappedObjects are objects describing a callable.

MarcelloDuarte:
Closing because there is a workaround via array callable (PhpSpec >=8 was object design only, so function constructed was not supported)

PR #1506, opened 2025-12-29
Wrong version in new release

### LINKED ISSUE #1504: The PhpSpec PHAR version is incorrect
The version displayed for 8.1.0 is 8.0.0:

```shell
# Download 8.1.0
wget --quiet https://github.com/phpspec/phpspec/releases/download/8.1.0/phpspec.phar --output-document=phpspec.phar
chmod +x phpspec.phar
./phpspec.phar # 8.0.0
```

MarcelloDuarte:
Fixes #1504

cc @theofidry

### DIFF
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -67,7 +67,7 @@ final class Application extends BaseApplication
         $this->loadConfigurationFile($input, $this->container);
 
         foreach ($this->container->getByTag('console.commands') as $command) {
-            $this->add($command);
+            $this->addCommand($command);
         }
 
         $dispatcher = $this->container->get('console_event_dispatcher');
--- a/src/PhpSpec/Console/Application.php
+++ b/src/PhpSpec/Console/Application.php
@@ -67,7 +67,11 @@ final class Application extends BaseApplication
         $this->loadConfigurationFile($input, $this->container);
 
         foreach ($this->container->getByTag('console.commands') as $command) {
-            $this->addCommand($command);
+            if (method_exists($this, 'addCommand')) {
+                $this->addCommand($command);
+            } else {
+                $this->add($command);
+            }
         }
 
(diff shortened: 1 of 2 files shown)


src/PhpSpec/Console/Application.php
```diff
@@ -67,7 +67,11 @@ public function doRun(InputInterface $input, OutputInterface $output): int
         $this->loadConfigurationFile($input, $this->container);
 
         foreach ($this->container->getByTag('console.commands') as $command) {
-            $this->add($command);
+            if (method_exists($this, 'addCommand')) {
+                $this->addCommand($command);
+            } else {
+                $this->add($command);
+            }
```
jakubtobiasz, line 74: I believe this part should be resolved [here](https://github.com/phpspec/phpspec/pull/1503) and should not be a part of this PR.

roukmoute: For my part, I feel that the current proposal is the right approach. Simply replacing it with `addCommand` would risk breaking compatibility with Symfony 5.4, 6.x, and 7.x.

Using `if (method_exists())` seems appropriate to me to ensure backward compatibility between versions. Alternatively, we could use `addCommands([$command])`, which is available everywhere, even if it is indeed a little less elegant. What do you think?

roukmoute:
Hello @stof, would you have a moment to take a look at this PR :) ?

src/PhpSpec/Console/Application.php
Chris53897: I agree with @roukmoute for his solution to support more symfony versions.
But i think it should be moved to a seperate PR.

Chris53897: Update: It is already merged in a sperate PR. Looks like addCommand() is supported since 5.4 accoring to test-runs.

torchello:
Fixed in https://github.com/phpspec/phpspec/pull/1509

PR #1508, opened 2026-01-19
Fix `Call to undefined method PhpSpec\Console\Application::add()` in `Application`

mspirkov:
There is such an error here: https://github.com/phpspec/prophecy/actions/runs/21114065815/job/60726462700
This is because the `add` method was removed in version 8.0 of `symfony/console`.


mspirkov:
Duplicate: https://github.com/phpspec/phpspec/pull/1503

PR #1511, opened 2026-03-15
Proof of concept for PhpSpec 9

MarcelloDuarte:
## Summary

This is a proof of concept for PhpSpec 9 — a ground-up modernisation of the framework.

### Why PhpSpec 9?

PhpSpec has served the PHP community well for over a decade, but the ecosystem has moved on:

- **PHP itself has evolved dramatically.** Enums, fibers, readonly classes, first-class callables, named arguments — modern PHP enables patterns that weren't possible when PhpSpec was designed. The current architecture can't take full advantage of these.

- **The dependency surface has grown fragile.** Maintaining compatibility across Prophecy, Doctrine Instantiator, Sebastian Exporter, four Symfony packages, and their cascading version matrices is an ever-growing maintenance burden. Every upstream release risks breakage. PhpSpec 9 reduces runtime dependencies from 9 packages to 2 (symfony/console + symfony/yaml).

- **The `ObjectBehavior` pattern creates friction.** Developers must learn a DSL that maps `$this->shouldReturn()` to assertions and `$this->beConstructedWith()` to constructors. It's powerful but opaque. The `describe`/`it`/`expect` pattern (used by virtually every other testing framework across languages) is immediately familiar.

- **Story BDD shouldn't require a separate tool.** Currently phpspec users need Behat for feature-level testing. PhpSpec 9 has built-in Gherkin support with `given()`/`when()`/`then()` step definitions, unifying Spec BDD and Story BDD in a single tool.

- **AI-assisted development is the new reality for BDD.** AI coding assistants can generate classes, methods, and entire modules in seconds — but speed without direction is dangerous. An AI will happily produce code that looks correct but doesn't match the intended behaviour. This is exactly the problem BDD was designed to solve: a spec is a machine-readable contract that validates behaviour, whether a human or an AI writes the implementation. BDD becomes *more* important with AI, not less. PhpSpec 9 embraces this by making AI a first-class participant in the BDD cycle, not an afterthought.

### What changes

**Syntax** — Jasmine/RSpec-style closures replace ObjectBehavior:

```php
// PhpSpec 8
class CalculatorSpec extends ObjectBehavior
{
    function it_adds_numbers()
    {
        $this->add(2, 3)->shouldReturn(5);
    }
}

// PhpSpec 9
describe(Calculator::class, function () {
    it('adds numbers', function () {
        expect((new Calculator())->add(2, 3))->toBe(5);
    });
});
```

**Mocking** — Built-in, no Prophecy:

```php
it('logs messages', function (Logger $logger) {   // auto-mocked
    allow($logger->info())->toReturn(true);
    // ... exercise SUT ...
    expect($logger->info())->toBeCalled()->once();
});
```

**Story BDD** — Built-in Gherkin support:

```gherkin
Feature: Calculator
  Scenario: Adding numbers
    Given I have a calculator
    When I add 2 and 3
    Then the result should be 5
```

```php
given('I have a calculator', function () {
    $this->calculator = new Calculator();
});
```

**Dependencies** — Dramatically reduced:

| | PhpSpec 8 | PhpSpec 9 |
|---|---|---|
| Runtime packages | 9 | 2 |
| Prophecy | Required | Removed |
| Doctrine Instantiator | Required | Removed |
| Sebastian Exporter | Required | Removed |
| Symfony packages | 5 (Console, EventDispatcher, Process, Finder, Yaml) | 2 (Console, Yaml) |

### AI-powered pair programming

PhpSpec 9 includes three AI commands — `pair`, `next`, and `refactor` — that embed the AI inside the BDD cycle rather than running alongside it. All AI features are **opt-in**: they require a `suggest` dependency (`papi-ai/papi-core` + a provider package) and an `ai:` section in your config. Without them, PhpSpec works exactly as it does without AI — zero runtime cost, no API keys needed.

#### Why AI belongs in the BDD tool

AI assistants generate code fast, but they don't know *what behaviour is correct*. Specs do. By placing the AI inside PhpSpec, every generated class and method is immediately validated against specs. The BDD cycle becomes the contract that keeps both human and AI-generated code honest:

```
Feature scenario (failing)
  → AI generates step definitions
    → Spec (failing)
      → AI generates class/method
        → Spec (green) ← the spec decides what "correct" means
      → Step (green)
    → Scenario (green)
```

The AI accelerates the cycle; the specs guarantee correctness. Neither replaces the other.

#### `pair` — Interactive BDD REPL

Start an interactive pair programming session:

```bash
bin/phpspec pair
```

Or send a single prompt:

```bash
bin/phpspec pair --prompt "write a spec for a Calculator that adds two numbers"
```

The pair command launches a REPL where built-in commands (`describe`, `exemplify`, `run`) work without AI, and free-form input routes to the AI assistant when configured. Smart routing detects intent:

```
> describe App/Calculator                       # runs describe command
> describe what the Loader class does            # routes to AI (natural language)
> run spec/App                                   # runs specs
> run my specs and explain the failures          # routes to AI
```

**The AI is agentic** — it has tools to generate specs, features, step definitions, write/update files, run specs, and read project files. It receives your project's directory structure, existing step definitions, the full DSL reference, and matcher/mock syntax as context, so it writes code that matches your project's patterns and reuses existing steps.

A typical pairing flow:

```
> write a feature scenario for user registration

  ✓ Generated features/scenarios/user_registration.feature
  ✓ Generated features/steps/user_registration.steps.php

> run features/

  ✗ Class App\UserRepository not found

> describe App/UserRepository

  ✓ Generated spec/App/UserRepository.spec.php
  ✓ Generated src/App/UserRepository.php

> now add a findByEmail method that returns a User or null

  ✓ Updated spec/App/UserRepository.spec.php
  ✓ Updated src/App/UserRepository.php

> run

  ✓ All specs pass
  ✓ All scenarios pass
```

Conversation history is maintained across the session — the AI remembers what you've been working on and builds on previous exchanges. All interactions are logged to `.phpspec/pair.log`.

#### `next` — What should I work on?

```bash
bin/phpspec next
```

Scans your project's source, specs, and features, then suggests the single most impactful next step. It follows the scenario-first workflow — recommends a feature scenario before a spec, a spec before an implementation:

```
  Analysing project...

  Write a feature scenario for user registration.

  Your project has a UserRepository class with a spec, but no feature
  scenario covering the registration flow. Adding a scenario first will
  drive the step definitions and any missing specs.
```

#### `refactor` — AI-powered, behaviour-preserving refactoring

```bash
bin/phpspec refactor "App\Calculator"
bin/phpspec refactor "App\Calculator::sum"
```

The command:

1. Runs specs to establish a green baseline (refuses to refactor broken code)
2. Sends source + spec to the AI, which identifies a single baby-step refactoring
3. Applies the change and re-runs specs
4. If specs pass → keeps the change and shows a diff. If specs fail → rolls back automatically

```
Technique: Extract Method
Description: Extracted validation logic into a validateInput() method

   1   <?php
   2   namespace App;
   3 - class Calculator {
   3 + class Calculator {
   4       public function add(int $a, int $b): int {
   5 -         if ($a < 0 || $b < 0) {
   6 -             throw new \InvalidArgumentException('Negative');
   7 -         }
   5 +         $this->validateInput($a, $b);
              return $a + $b;
          }
  10 +
  11 +     private function validateInput(int $a, int $b): void {
  12 +         if ($a < 0 || $b < 0) {
  13 +             throw new \InvalidArgumentException('Negative');
  14 +         }
  15 +     }
      }

Specs still pass.
```

#### AI configuration

Add an `ai:` section to `phpspec.yaml`:

```yaml
ai:
  provider: anthropic          # or google, openai
  model: claude-sonnet-4-20250514  # optional, sensible defaults per provider
  api_key: YOUR_API_KEY
```

Supported providers: **Google** (Gemini), **Anthropic** (Claude), **OpenAI** (GPT). The provider abstraction is behind an interface (`ProviderInterface`) — adding new providers is straightforward.

### What's included

- Full spec suite: 86 specs, 1030+ examples, 91%+ coverage
- Feature scenarios for all major commands
- Code generation (specs, classes, interfaces, methods)
- Formatters: Pretty, Dot, TAP, JUnit XML
- Parallel execution via PHP Fibers
- AI pair programming, next-step suggestion, and automated refactoring (opt-in)
- Documentation in `docs/`

### Status

This is a **proof of concept** for discussion. It demonstrates the direction, not a final implementation. Feedback on the approach, syntax choices, and migration path is welcome.

### Core

- [ ] `composer install`
- [ ] `php bin/phpspec run` — all specs pass
- [ ] `php bin/phpspec run features/` — all feature scenarios pass
- [ ] `vendor/bin/php-cs-fixer fix --dry-run` — code style clean
- [ ] `vendor/bin/phpstan analyse -l 1 src/` — static analysis clean

### AI pair programming (optional — requires API key)

To try the AI flow, install a provider and add config:

```bash
composer require papi-ai/papi-core papi-ai/google   # or papi-ai/anthropic, papi-ai/openai
```

```yaml
# phpspec.yaml
ai:
  provider: google       # or anthropic, openai
  api_key: YOUR_API_KEY
```

Then experiment with the BDD cycle:

- [ ] `php bin/phpspec pair` — enter the interactive REPL; try built-in commands (`describe App/Greeter`, `run`) without AI first
- [ ] In pair mode, ask the AI to generate a feature: `write a feature scenario for greeting users` — verify it creates `.feature` and `.steps.php` files
- [ ] In pair mode, run the feature and let the AI drive the spec/class generation cycle: `run features/` → `describe App/Greeter` → ask the AI to add methods → `run` until green
- [ ] `php bin/phpspec pair --prompt "write a spec for a Calculator"` — single-prompt mode, verify it generates a spec file
- [ ] `php bin/phpspec next` — verify it scans the project and suggests a next step following the scenario-first workflow
- [ ] `php bin/phpspec refactor "App\Greeter"` — verify it runs specs, applies a refactoring, and re-runs specs (or reports no refactoring needed)


MarcelloDuarte:
@torchello @Jean85 @stof @everzet @ciaranmcnulty I also put together a website with the new concept

https://phpspec-site.fly.dev/

Jean85:
Questions:
1. would this mean rewriting it with AI? What about license concerns?
2. Such a hard breaking change would stop a lot of users from upgrading; can we ease the transition in some way?

stof:
This looks like a totally different test framework. Maybe it should be released as a *different* package instead of a new major version of `phpspec/phpspec` to allow projects to use both at the same time to adopt the new tool little by little.

.github/workflows/build.yml
```diff
@@ -2,143 +2,106 @@ name: Build
...
+                composer-flags: ['']
+                include:
+                    - php: '8.2'
+                      operating-system: ubuntu-latest
+                      publish-phar: true
+                    - php: '8.2'
+                      operating-system: ubuntu-latest
+                      composer-flags: --prefer-lowest
 
-            -   name: Run static analysis (psalm)
```
stof, line 123: Why removing static analysis from the CI setup ?

.github/workflows/build.yml
```diff
@@ -2,143 +2,106 @@ name: Build
...
-            -   name: Upload phpspec.phar
-                uses: actions/upload-release-asset@v1
-                env:
-                    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-                with:
-                    upload_url: ${{ github.event.release.upload_url }}
-                    asset_path: phpspec.phar
-                    asset_name: phpspec.phar
-                    asset_content_type: application/zip
+            - uses: actions/download-artifact@v4
```
stof: this should not downgrade the action

composer.json
```diff
@@ -1,78 +1,69 @@
...
-        "phpspec/nyan-formatters": "Adds Nyan formatters"
-    },
-
-    "conflict": {
-        "sebastian/comparator" : "<1.2.4"
+        "php": "^8.2 || ^8.3 || ^8.4 || ^8.5",
+        "symfony/console": "^7.0 || ^8.0",
+        "symfony/yaml": "^7.0 || ^8.0",
+        "ext-dom": "*",
+        "ext-curl": "*",
```
stof: why making `ext-curl` a mandatory requirement ?

phpstan.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+    level: 1
```
stof: level 1 of phpstan is basically useless. A new codebase should actually *start* at level 8 or higher.

src/PhpSpec/StoryBDD/GherkinParser.php
```diff
@@ -0,0 +1,333 @@
...
+ */
+
+namespace PhpSpec\StoryBDD;
+
+/**
+ * Parses .feature file content into a FeatureNode data structure.
+ * Supports Background, Scenario, Scenario Outline with Examples tables, DataTables,
+ * Doc strings, tags, and And/But step keywords.
+ */
+final class GherkinParser
```
stof, line 37: Please don't write another Gherkin parser. Use `cucumber/gherkin` instead to use the official parser that is tested against the Gherkin shared testsuite (or use `behat/gherkin` if you prefer, but that would be better to use the official parser to be future-proof).

The current parser implementation is **not** spec compliant.

src/PhpSpec/EventDispatcher/Dispatcher.php
```diff
@@ -0,0 +1,188 @@
...
+
+use PhpSpec\Specification\ExampleRegistry;
+
+/**
+ * Static event bus that coordinates DSL functions, spec tree building, and result collection.
+ *
+ * Manages a scope stack to track the current context/example registry during spec
+ * file loading, and dispatches events to registered subscribers and listeners.
+ */
+final class Dispatcher
```
stof, line 29: is it expected that the dispatcher is all based on global state while not being internal ?

src/PhpSpec/Ai/AiTools.php
```diff
@@ -0,0 +1,88 @@
...
+            description: 'Read the contents of a file from the project.',
+            parameters: [
+                'path' => [
+                    'type' => 'string',
+                    'description' => 'Absolute or relative path to a file',
+                ],
+            ],
+            handler: function (array $args) use ($filesystem) {
+                $path = $args['path'];
+                if (!str_starts_with($path, '/')) {
```
stof: this won't work on Windows. Absolute paths don't always start with `/`

src/PhpSpec/Ai/AiTools.php
```diff
@@ -0,0 +1,88 @@
...
+                    $path = getcwd() . '/' . $path;
+                }
+
+                if (!$filesystem->exists($path)) {
+                    return "File not found: {$args['path']}";
+                }
+
+                return $filesystem->read($path);
```
stof, line 51: shouldn't this protect against directory traversal like `../../../etc/password` or `/etc/password` which are outside the project ?

src/PhpSpec/Browser/functions.php
```diff
@@ -0,0 +1,171 @@
...
+function delete(string $path, array $options = []): Response
+{
+    return BrowserRegistry::client()->request('DELETE', $path, $options);
+}
+
+/**
+ * Static registry holding the Browser Client instance.
+ * Lazily creates the client from Configuration on first access.
+ */
+class BrowserRegistry
```
stof: shouldn't this class be registered in its own file following PSR-4 ?

stof: thus, it is currently defined outside the phpspec namespace, which is even worse.

src/PhpSpec/CodeGeneration/ClassGenerator.php
```diff
@@ -0,0 +1,104 @@
...
+
+use PhpSpec\Filesystem;
+use PhpSpec\RealFilesystem;
+use RuntimeException;
+
+/**
+ * Generates PHP class files from a fully qualified class name.
+ * Creates the directory structure and writes a minimal class skeleton.
+ */
+final class ClassGenerator
```
stof, line 26: Most of these classes should probably be tagged as `@internal`, to properly define the public API covered by semver.

src/PhpSpec/CodeGeneration/ClassGenerator.php
```diff
@@ -0,0 +1,104 @@
...
+    public static function resolveFqcn(string $fqcn, string $srcPath): array
+    {
+        $pieces = explode('\\', $fqcn);
+        $shortName = $pieces[count($pieces) - 1];
+        $namespace = '';
+        $filePath = getcwd() . DIRECTORY_SEPARATOR . $srcPath . DIRECTORY_SEPARATOR;
+
+        if (count($pieces) > 1) {
+            $namespace = "\n\nnamespace " . implode('\\', array_slice($pieces, 0, count($pieces) - 1)) . ';';
+            $filePath .= implode(DIRECTORY_SEPARATOR, array_slice($pieces, 0, -1)) . DIRECTORY_SEPARATOR . $shortName . '.php';
```
stof: this seems to support only PSR-0 structures (or rather, the common subset of PSR-0 and PSR-4 as it has no special handling for `_` in global namespace classes), while almost all projects use PSR-4 now. This should support configuring which namespace is mapped to the `src` path.

src/PhpSpec/Coverage/CloverReport.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+namespace PhpSpec\Coverage;
+
+use DOMDocument;
+use DOMException;
+
+/**
+ * Renders a Clover XML code coverage report, compatible with CI tools like Jenkins and Codecov.
+ * Produces a standard Clover coverage XML document with per-file line-level data and project metrics.
```
stof, line 23: On which clover standard is this based ? PHPUnit has added a new output format due to several competing definitions of "clover" and not being fully compatible.

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+        // create a unique name for the double class
+        $mockClassName = $className . '_Double_' . uniqid();
+
+        // register the mapping from mock class name to original FQCN
+        self::$doubleRegistry[$mockClassName] = $class;
+
+        // generate the methods for the double
+        $reflectionClass = new ReflectionClass($class);
+
+        if (!$reflectionClass->isInterface() && ($reflectionClass->isFinal() || $reflectionClass->isReadOnly())) {
```
stof, line 95: this should probably reject traits and enums as well.

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+     * @throws ReflectionException
+     */
+    public static function getInstance(string $class): object
+    {
+        if (isset(self::$classCache[$class])) {
+            $cached = self::$classCache[$class];
+            return self::generateDouble($cached['reflectionClass'], $cached['mockClassName']);
+        }
+
+        // get the class base name (e.g. A\B\C → C) for the double class name
```
stof: why using only the short class name ? It is not unique at all.

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+        $className = substr($class, strrpos($class, '\\') + 1);
+
+        // create a unique name for the double class
+        $mockClassName = $className . '_Double_' . uniqid();
```
stof: It might make sense to put the generated mocks in a namespace like `PhpspecDouble` instead, to reduce the risk of conflicts.

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+        }
+
+        // put it all together and evaluate the double class template
+        $mockClassCode = <<<PHP
+
+class $mockClassName $extends $prefixedClass {
+    private \$stack;
+    private \$stubbedReturns = [];
+    private \$stubbedThrows = [];
+    private \$stubbedReturnCallbacks = [];
```
stof: those should probably be prefixed, to avoid conflicts with properties of the parent class (the current implementation does not support mocking a class that has a protected `$stack` property with a type declared)

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+
+    $methods
+
+    private function ______PhpSpecWasCalledWith(\$method, \$args) {
```
stof, line 138: should this code be defined in a trait to make it easier to maintain it instead of being part of the generated class ?

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+                continue;
+            }
+
+            // Intersection types: try to mock first type
+            if ($parameterType instanceof ReflectionIntersectionType) {
+                $types = $parameterType->getTypes();
+                $first = $types[0];
+                if ($first instanceof ReflectionNamedType && !$first->isBuiltin()) {
+                    try {
+                        $dependencies[] = Double::getInstance($first->getName());
```
stof, line 245: this won't satisfy an intersection type. It must be a double implementing *all* interfaces.

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+        }
+
+        if ($type instanceof ReflectionUnionType) {
+            $parts = [];
+            foreach ($type->getTypes() as $t) {
+                $name = $t->getName();
+                if ($t instanceof ReflectionNamedType && !$t->isBuiltin()) {
+                    $name = '\\' . $name;
+                }
+                $parts[] = $name;
```
stof: this is broken. A `ReflectionUnionType` can contain `ReflectionIntersectionType`. This should use `formatTypeSyntax` recursively instead (and check for necessary braces).

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+            // format return type syntax using helper
+            $returnedTypeSyntax = '';
+            $isClassReturn = false;
+            $simpleReturnName = self::getSimpleReturnTypeName($returnedType);
+            $isNullable = self::isNullableType($returnedType);
+
+            if ($returnedType) {
+                $returnedTypeSyntax = ': ' . self::formatTypeSyntax($returnedType);
+                $isClassReturn = self::isClassReturnType($returnedType);
+                if ($simpleReturnName && !($returnedType instanceof ReflectionNamedType && $returnedType->isBuiltin())) {
```
stof: Why returning a type *name* in this variable and having to redo all the work instead of returning `?ReflectionNamedType` and then reusing `formatTypeSyntax` ?

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+            if ($isClassReturn && !$isEnum && $depth === 0) {
+                $returnedTypeReflected = new ReflectionClass($returnedType);
+                $templateMethods = '';
+
+                if (count($returnedTypeReflected->getMethods())) {
+                    $templateMethods = self::generateMethods($returnedTypeReflected, $depth + 1);
+                }
+                $template = '';
+                if ($returnedTypeReflected->isInterface() && $returnedTypeReflected->getName() !== 'Throwable') {
+                    $className = 'MockedMethodFor' . ucfirst($methodName) . uniqid();
```
stof: this class name is does not seem to account for the fact that multiple mocked classes might use the same method name.

src/PhpSpec/Mock/Double.php
```diff
@@ -0,0 +1,678 @@
...
+            $isEnum = $isClassReturn && $returnedType && enum_exists(ltrim($returnedType, '\\'));
+            if ($isClassReturn && !$isEnum && $depth === 0) {
+                $returnedTypeReflected = new ReflectionClass($returnedType);
+                $templateMethods = '';
+
+                if (count($returnedTypeReflected->getMethods())) {
+                    $templateMethods = self::generateMethods($returnedTypeReflected, $depth + 1);
+                }
+                $template = '';
+                if ($returnedTypeReflected->isInterface() && $returnedTypeReflected->getName() !== 'Throwable') {
```
stof: what about `Traversable` or `DatetimeInterface` which are also special in core ?

src/PhpSpec/Mock/functions.php
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Creates a test double for the given class or interface.
+ *
+ * @param class-string<object> $class fully qualified class or interface name
+ * @return object the generated test double instance
+ * @throws InvalidArgumentException if the class or interface does not exist
+ * @throws ReflectionException
+ */
+function mock(string $class): object
```
stof, line 30: I think all those functions should be in a namespace instead of being in the global namespace.

src/PhpSpec/Mock/MockedMethod.php
```diff
@@ -0,0 +1,87 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace PhpSpec\Mock;
+
+/**
+ * Records a single method invocation on a test double, tracking the subject,
+ * method name, arguments, and a mutable call count used for verification.
+ */
+/* readonly */ final class MockedMethod
```
stof: why this commented keyword ? This class is *not* readonly as it has a mutable call count.

src/PhpSpec/Report/Formatter/Pretty/views/specification.errors.view.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+?>
+<?php foreach ($specification->getResults() as $exampleResult): ?>
+<?php if ($exampleResult->isContext() && $exampleResult->isError() === false): ?>
+<?php $this->render('context.errors', [
+    'context' => $exampleResult,
+    'specification' => $specification->getTitle() . ' > ' . $exampleResult->getTitle(),
+]) ?>
+<?php elseif ($exampleResult->isContext() && $exampleResult->isError()): ?>
+<?php $this->write(PHP_EOL . '  <fg=red>• ' . $specification->getTitle() . ' > ' . $exampleResult->getTitle() . '</>' . PHP_EOL) ?>
```
stof: why using separate PHP tags when *all* the content is in PHP tags ? It only makes the file harder to read IMO, without any of the benefits of actual PHP templates.

stof: those templates could actually be replaced by *functions* having access to a writer (to replace `$this->write()`) and a renderer (to replace `$this->render()`, unless extensibility is not needed and it could use a function call directly) and having a well-defined signature, which would even allow type-checking the code properly, with the same way to write the code.

src/PhpSpec/Report/Formatter/Tap.php
```diff
@@ -0,0 +1,137 @@
...
+        return $examples;
+    }
+
+    /**
+     * Escapes a string for safe inclusion in YAML diagnostic output.
+     */
+    private function yamlEscape(string $value): string
+    {
+        if (str_contains($value, "\n") || str_contains($value, '"') || str_contains($value, ':')) {
+            return '"' . addcslashes($value, "\"\\\n") . '"';
```
stof: there are more special characters in YAML double-quoted strings.

src/PhpSpec/Report/Formatter/Tap.php
```diff
@@ -0,0 +1,137 @@
...
+
+        return $examples;
+    }
+
+    /**
+     * Escapes a string for safe inclusion in YAML diagnostic output.
+     */
+    private function yamlEscape(string $value): string
+    {
+        if (str_contains($value, "\n") || str_contains($value, '"') || str_contains($value, ':')) {
```
stof: A string starting with a single quote will also need escaping, to avoid being parsed as a single-quoted string.

composer.json
```diff
@@ -1,78 +1,69 @@
...
-    "suggest": {
-        "phpspec/nyan-formatters": "Adds Nyan formatters"
-    },
-
-    "conflict": {
-        "sebastian/comparator" : "<1.2.4"
+        "php": "^8.2 || ^8.3 || ^8.4 || ^8.5",
+        "symfony/console": "^7.0 || ^8.0",
+        "symfony/yaml": "^7.0 || ^8.0",
+        "ext-dom": "*",
```
stof: isn't `ext-dom` needed only for the JUnit formatter ? If yes, it should be an optional dependency IMO.

CHANGES-v9.md
```diff
@@ -0,0 +1,35 @@
+# Changelog — PhpSpec 9
+
+## [9.0.0-poc] — Proof of Concept
+
+### Breaking Changes
```
stof: another breaking change: breaks compat with all existing phpspec extensions, as it is basically a different project.

CHANGES-v9.md
```diff
@@ -0,0 +1,35 @@
+# Changelog — PhpSpec 9
```
stof: why using a separate changelog file ?

check-release.php
```diff
@@ -1,49 +0,0 @@
-<?php
-
-/**
- * Sanity-checks a release for consistency
```
stof, line 4: Why ius this script gone ?

stof:
@MarcelloDuarte the website using gray text on a black background is almost unreadable. Please change the website styles to use good contrasting colors.

src/PhpSpec/Mock/Double.php
stof: and what about other core classes that require special handling when mocking them ? Prophecy has more special cases that were built over time.

MarcelloDuarte:
> Questions:
> 
> 1. would this mean rewriting it with AI? What about license concerns?
> 2. Such a hard breaking change would stop a lot of users from upgrading; can we ease the transition in some way?

@Jean85 thanks for the feedback! 🙏

On the AI question — the code is original, written with AI as a development tool (same as using an IDE with autocomplete). No third-party code was incorporated, so no license concerns. I kept Konstantin in the copyright and added Ciaran for his longstanding contributions as a maitainer.

On the migration — we'd provide a migration tool to mechanically convert ObjectBehavior specs to the new syntax, and maintain 8.x with security/compatibility fixes during a transition period. Were you thinking of anything beyond that?

MarcelloDuarte:
@stof thanks for your feedback. I will go over one by one.

> This looks like a totally different test framework. Maybe it should be released as a _different_ package instead of a new major version of `phpspec/phpspec` to allow projects to use both at the same time to adopt the new tool little by little.

That is an excellent idea. Releasing this as phpspec/phpspec-9 or phpspec/phpspec-next makes the migration story much better. It would help with the issues raised by @Jean85 as well.

The downside is it fragments the brand/community a little, but for adoption it's arguably the pragmatic choice. If the adoption takes off, we could move to phpspec/phpspec on 9.1 to counter that.

stof:
Maybe using `phpspec/runner` could avoid the need to rename the package in the future by having an acceptable name.

.github/workflows/build.yml
MarcelloDuarte: I had stan locally, didn't realise we had psalm until merging. Restored 7e23af25

.github/workflows/build.yml
MarcelloDuarte: Restored to v5 [9a9d91b](https://github.com/phpspec/phpspec/pull/1511/commits/9a9d91bad300bfed52b6dd39988027a92214f39b)

composer.json
MarcelloDuarte: Moving it to suggest. It is used with some optional AI features 7fecb99f

phpstan.neon
MarcelloDuarte: restored psalm anyway

src/PhpSpec/StoryBDD/GherkinParser.php
MarcelloDuarte: Agreed. [03d5dd0](https://github.com/phpspec/phpspec/pull/1511/commits/03d5dd0b81669fbd93444ee7e41956f9c95c1d13)

.github/workflows/build.yml
stof: Well, I'm fine with using phpstan (this is what I'm using in all my projects, as I find it superior these days, especially when it comes to extensions). but it should be part of the CI

src/PhpSpec/StoryBDD/GherkinParser.php
```diff
@@ -0,0 +1,173 @@
...
+        $parser = new CucumberParser(
+            includeSource: false,
+            includeGherkinDocument: true,
+            includePickles: false,
+        );
+
+        $document = null;
+        $errors = [];
+
+        foreach ($parser->parseString('inline.feature', $content) as $envelope) {
```
stof: shouldn't you pass the actual file name to the parser, to ensure that the right file name is used in error messages ?

src/PhpSpec/StoryBDD/GherkinParser.php
```diff
@@ -0,0 +1,173 @@
...
+    private function mapFeature(Feature $feature): FeatureNode
+    {
+        $background = null;
+        $scenarios = [];
+
+        foreach ($feature->children as $child) {
+            if ($child->background !== null) {
+                $background = $this->mapBackground($child->background);
+            }
+            if ($child->scenario !== null) {
```
stof, line 119: This does *not* handle `Rule` children.

Maybe you should compile the feature to pickles and convert pickles to your own data model (or process pickles directly). In cucumber, pickles are representing what needs to be executed, abstracting away all the Gherkin features related to organizing the feature file.

src/PhpSpec/StoryBDD/functions.php
```diff
@@ -0,0 +1,167 @@
...
+function beforeStep(Closure $fn): void
+{
+    StoryBDDRegistry::$hooks->addBeforeStep($fn);
+}
+
+/**
+ * Static registry holding the global StepRegistry and HookRegistry instances.
+ * Initialized on file load and used by the DSL functions (given/when/then/etc.).
+ */
+class StoryBDDRegistry
```
stof: this class should not be defined in the global namespace either.

src/PhpSpec/StoryBDD/functions.php
```diff
@@ -0,0 +1,167 @@
...
+     * @return void
+     */
+    public static function restoreState(array $state): void
+    {
+        self::$steps = $state['steps'];
+        self::$hooks = $state['hooks'];
+    }
+}
+
+StoryBDDRegistry::init();
```
stof: I would suggest adding static methods in `StoryBDDRegistry` instead, to allow keeping this initialization lazy. Running such initialization when loading the `vendor/autoload.php` file of composer (even when phpspec is not running) does not make sense.

src/PhpSpec/EventDispatcher/DispatcherRegistry.php
```diff
@@ -0,0 +1,40 @@
...
+
+    public static function set(Dispatcher $dispatcher): void
+    {
+        self::$instance = $dispatcher;
+    }
+
+    public static function get(): Dispatcher
+    {
+        if (self::$instance === null) {
+            self::$instance = new Dispatcher();
```
stof: should this create a dispatcher or should it throw saying that this is not intended to be used outside a context where a runner has set a dispatcher ?

src/PhpSpec/EventDispatcher/Dispatcher.php
MarcelloDuarte: It is pretty common for function based testing frameworks to keep the state somewhere global. However we can use a slightly more OO approach for that [f997733](https://github.com/phpspec/phpspec/pull/1511/commits/f99773353102142015e2527e654563283a0ea94c)

.github/workflows/build.yml
MarcelloDuarte: Ok, reverting it to stan (I prefer it also), restored  7e1284b

phpstan.neon
MarcelloDuarte: [Set it to level 8](https://github.com/MarcelloDuarte/phpspec/blob/7e1284b900fb28175c796ea1859d6bd353a38940/phpstan.neon#L2)

src/PhpSpec/Ai/AiTools.php
MarcelloDuarte: Fixed [6c41347](https://github.com/phpspec/phpspec/pull/1511/commits/6c41347d65d5a1cd94256bfdfbc3612cb6412a95)

src/PhpSpec/Ai/AiTools.php
MarcelloDuarte: Added a [wrapper](https://github.com/MarcelloDuarte/phpspec/blob/2a7dfb59d3c294a85329851e21608324a22153a6/src/PhpSpec/Ai/AiTools.php#L98)

src/PhpSpec/Browser/functions.php
MarcelloDuarte: Fixed 5337cfbf

src/PhpSpec/CodeGeneration/ClassGenerator.php
MarcelloDuarte: Fixed f66698d

src/PhpSpec/CodeGeneration/ClassGenerator.php
MarcelloDuarte: Fixed 663d76ac

src/PhpSpec/Coverage/CloverReport.php
MarcelloDuarte: oh, just plain old Atlassian clover. Wasn't aware people started writing variations of it...

src/PhpSpec/Mock/functions.php
MarcelloDuarte: The DSL functions (allow(), mock(), any(), type(), callback(), arrayIncluding()) are intentionally in the global namespace for developer experience — they're the primary API users interact with in every spec file. Requiring use function PhpSpec\Mock\allow, PhpSpec\Mock\mock,... at the top of every spec would add friction to the workflow that the closure-based DSL was designed to eliminate.

This follows the same approach as the core DSL (describe(), it(), expect(), let()) — all global for the same reason. The tradeoff is deliberate: a small global namespace footprint in exchange for zero-boilerplate specs.

That said, for testing libraries with functions of the same name, and to avoid collision, we could consider offering a phpspec.yaml option to opt into namespaced functions (with a bootstrap that registers global aliases by default). What do you think? I'd say that is not a must, but a nice to have...

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 45d83cb9 and 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 1d8377af — now rejects all internal interfaces, not just Throwable

src/PhpSpec/Mock/MockedMethod.php
MarcelloDuarte: Fixed ba659030

src/PhpSpec/Mock/Double.php
MarcelloDuarte: Fixed 60c94ae2 — extracted a `MatchableDoubleBehaviour` trait used by all generated wrapper classes

src/PhpSpec/Report/Formatter/Pretty/views/specification.errors.view.php
MarcelloDuarte: Fixed 619fd36f — replaced all view templates with typed static methods in PrettyViews class

MarcelloDuarte: Fixed 619fd36f — templates are now functions (static methods) with typed signatures, OutputInterface writer, and direct method calls instead of $this->render()

src/PhpSpec/Report/Formatter/Tap.php
MarcelloDuarte: Fixed 9027ce91

composer.json
MarcelloDuarte: Fixed 671aa6e5

CHANGES-v9.md
MarcelloDuarte: Yes, this is intentional — the extension API has changed significantly (event dispatcher, formatter interface, etc.). We'll document a migration guide for extension authors. The new extension interfaces (ListenerExtension, MatcherExtension, FormatterExtension, CommandExtension) are simpler than the current system. Releasing as a separate package (per your earlier suggestion) also avoids breaking existing extensions that target phpspec/phpspec ^8.

CHANGES-v9.md
MarcelloDuarte: Fixed ef8d11b6 — consolidated into CHANGES.md, removed CHANGES-v9.md

check-release.php
MarcelloDuarte: Fixed ef8d11b6 — restored check-release.php adapted for the v9 structure

src/PhpSpec/StoryBDD/GherkinParser.php
MarcelloDuarte: Fixed 5bf133e9

src/PhpSpec/StoryBDD/GherkinParser.php
MarcelloDuarte: Fixed b8ed5177 — refactored to use Cucumber pickles. Handles Rule children, background merging, and Scenario Outline expansion automatically. Step keywords preserved via AST node ID lookup.

src/PhpSpec/StoryBDD/functions.php
MarcelloDuarte: Fixed d215cd02

src/PhpSpec/StoryBDD/functions.php
MarcelloDuarte: Fixed f26e8785 — registry is now lazy-initialized via steps()/hooks() accessors. No side effects on autoload.

src/PhpSpec/EventDispatcher/DispatcherRegistry.php
MarcelloDuarte: The lazy creation is intentional — DSL functions (describe(), it(), etc.) call DispatcherRegistry::dispatcher() during spec file loading, which happens before the Application has a chance to set a dispatcher. Throwing here would break the spec loading flow. The lazy-created dispatcher works correctly because all callers go through the same registry singleton.

MarcelloDuarte:
Ok, let me know if we have more questions, feedback. I can redirect this to another repository either:
 - phpspec/runner --->  _**Stof's suggestion**_
 - phpspec/future  --->  _**Similar to Python future library (where devs can experiment future features)**_
 - phpspec/phpspec9  --->  **_My preference. I don't want to distance too much from the brand and want to move in the direction of the final merge._**

torchello:
Hey folks! Sorry, I missed the PR, catching up....

@MarcelloDuarte First: this is impressive work. I need a few days to go through the code properly, but I think this direction is genuinely exciting and worth pushing forward.

## New package
I agree this should be released as a new package, and `phpspec/phpspec9` sounds good to me.

## New style
I'm personally not a fan of describe/it style:
```php
describe(Calculator::class, function () {
    it('adds numbers', function () {
        expect((new Calculator())->add(2, 3))->toBe(5);
    });
});
```
Mostly because `it('adds numbers', function () { ... })` feels less natural to me than `function it_adds_numbers()`.
That said, personal taste is secondary here. Does this new approach significantly simplify the implementation/maintenance of the framework, or is it mostly a syntax/style choice?

## AI features
I’m very curious to try this and hear real feedback. AI-oriented tooling fits the current moment and can bring a lot of attention to the framework.

One side note: I expect AI to take over more of the spec and implementation loop over time, so I’d also like us to think about agent-driven workflows, not only human-authored ones. Since the AI features are optional, that feels like a good foundation for experimenting without forcing a particular direction too early.

MarcelloDuarte:
Hey @torchello,

Good to hear from you!

> I agree this should be released as a new package, and phpspec/phpspec9 sounds good to me.

👍

> Does this new approach significantly simplify the implementation/maintenance of the framework?

The attempt here was to normalise phpspec a bit more with other spec frameworks like jasmin and rspec. And yes, `it(...)` is a simpler construct to `function it_...`.  This approach removes some of the magic in the previous implementation, e.g. people not understanding `$this` referred to the SUS was an entry barrier to some.

> AI-oriented tooling fits the current moment

Exactly right! And phpspec goals has always been make test a natural part of the development. If people naturally interacts with AI nowadays, this will fit in with current styles and could potentially integrate with IDEs plugins/extensions as well.

> I’d also like us to think about agent-driven workflows

Yep. That would be a nice direction to explore.

MarcelloDuarte:
hi @Jean85

I created a tool for migration at https://github.com/phpspec/migrate.

I am merging this to a 9.x branch for now. Later master, if all is well. I will release a v.9.0.0-beta.

Jean85:
Hi @MarcelloDuarte, thanks. I'm not actively using PHPSpec right now, just Prophecy... Would I be impacted by this change? Does Prophecy need any work? Does any Prophecy-only user need any work? Does the migration script has any use if you only use Prophecy?

stof:
@Jean85 no, you won't be impacted. Prophecy does not depend on phpspec (except as a dev dependency for its own testsuite)

PR #1512, opened 2026-04-13
Prepare for 8.3.0 release

stof:


stof:
Failing tests are because of an issue in Prophecy, which got undetected because phpspec itself was preventing us from testing with latest PHPUnit dependencies. It will be resolved in prophecy.

PR #1555, opened 2026-07-26
Consult the model for step bodies and existing features, and follow up on /generate chooser notes

MarcelloDuarte:
Two dogfood bugs from the same transcript.

**`generate the body of the steps in X.feature` answered with a no-op pending() scaffold.** The steps wording plus a named feature trips the deterministic scaffolder, which appends missing step definitions; with every step already defined it emitted a whitespace-only "MODIFIED" diff without ever consulting the model (the capture doc for such an exchange shows `request: null`). The rule is now state-based, not wording-based: a scaffold that adds nothing meaningful is not an answer, so the deterministic path declines and the model is asked, grounded in both the feature and the current steps file (named path tokens now bring their files into context, and a named .feature brings its steps file along). The same grow-not-overwrite rule now applies to `write_feature`: an existing feature file is grown via `propose_edit`, never re-scaffolded. The write-steps instructions and tool descriptions say so explicitly.

**A chooser note on the `/generate` confirm vanished.** "3. No, the body is still pending()! Replace that." left the REPL waiting. The note is now a course correction: it becomes the follow-up instruction for another round through the pipeline (declined: "the human declined, saying ... address that"; accepted: "applied, the human then asked ..."), bounded at three rounds. Because the follow-up re-resolves the step, a decline note can even redirect the artifact type.

Covered outside-in: a generate.feature scenario reproducing the transcript, registry examples for the no-op guards (trailing-newline normalisation included, which is exactly the phantom `14 +` line in the transcript), dispatcher examples proving the note reaches round two and the second proposal lands, and eval E13 replaying the exchange end to end: scaffold declines, model consulted with pending() bodies in context, propose_edit with real implementations wins.

Verified: 1639 examples and 1089 steps green on PHP 8.2.30, 8.3.16, 8.4.4, and 8.5.3; coverage 91.8%; phpstan and php-cs-fixer clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Fixed
+ - `generate` places a bare filename under its configured directory (features, spec, or src), never the project root
  - Asking `generate` for step bodies (or to grow an existing feature) consults the model instead of re-emitting a no-op scaffold
  - A `/generate` chooser note now runs a follow-up round with that note as the instruction
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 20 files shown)


MarcelloDuarte:
Added a second commit: the ask_user chooser (the AI's own yes/no questions) also hands a typed note back to the model with the answer ("no. The human added: ..."), the last chooser with a model listening that was still swallowing notes.

PR #1562, opened 2026-07-27
Journal applied refactorings so refactor never oscillates and next moves to growth

MarcelloDuarte:
STACKED on the refactor-consent PR (merge that first; this PR then shrinks to its own commit).

Dogfood found the oscillation: one session's Extract Method was undone by the next session's Inline Method, each locally defensible, and `next` kept suggesting "Refactor App\TodoList" forever because applying a refactor keeps the file the most recently touched. One-shots had zero cross-invocation memory.

The fix is a small append-only journal (.phpspec/ai/journal.jsonl), one JSON line per APPLIED refactoring (timestamp, target, technique, description). Declined and reverted changes are never recorded. Two consumers:

- `refactor` grounds the model with the recent entries ("do not undo or redo any of these; say no refactoring is worthwhile instead"), and the prompt now legitimises the none outcome explicitly, killing the extract/inline ping-pong.
- `next` applies the timestamp rule deterministically: a refactor suggestion whose target's mtime is at or before the last journalled refactoring becomes a growth steer ("already refactored and has not changed since; grow the story"). A target genuinely changed after the last refactoring keeps its refactor suggestion, which is exactly the changed-after-the-timestamp semantics from review.

Covered by a RefactorJournal spec (record, append, last timestamp, prompt rendering), command examples (applied is journalled, declined is not, the model sees the history), and two next examples pinning both sides of the timestamp rule.

Suite: 1669 examples green on PHP 8.2.30, 8.3.16, 8.4.4, and 8.5.3; 1094 steps; coverage 92.0%; phpstan and php-cs-fixer clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
- - Applied refactorings are journalled (.phpspec/ai/journal.jsonl); the model never undoes or redoes them, and `next` steers to growth when the target has not changed since
+ - Applied refactorings are journalled (.phpspec/ai/journal.jsonl); reversing one demands a stated rationale, and `next` steers to growth when the target has not changed since
 
 ### Changed
  - `refactor` shows the change and asks before applying; non-interactive runs keep auto-apply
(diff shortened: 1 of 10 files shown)


MarcelloDuarte:
Refined per review: reversing a journalled refactoring is permitted when the model genuinely judges the earlier design better, but it must name the change it reverses and state the rationale in its description; a silent taste swap stays forbidden. The reversal then lands in the journal with that rationale, so the history shows both moves and why.

PR #1565, opened 2026-07-27
Offer to act on determined steps instead of printing advice and exiting

MarcelloDuarte:
Dogfood: on a red suite, next printed \"The suite is red: deleting_a_task, deleting_a_task. Make it green.\" and exited. Two defects in one line: the determined steps (write the steps, make it green) resolved to a type with no offer arm, so the deterministic fast path never asked the [Y/n] every other suggestion asks; and the reason printed the failure's subject and example even when they are the same string.

Now the determined steps actuate like the rest:

- A write-steps step offers to run generate \"the steps for <feature>\", which writes the step definitions for that feature deterministically.
- A make-it-green step with a class subject offers generate \"implement <Class>\", whose wording routes to the write-code step with the class and its red spec in context. A subject that is no class (a story slug from a degenerate spec) has no honest one-shot actuator, so it prints the run hint instead of a fake offer.
- The red-suite reason names a subject that matches its example once.

Verification: 1683 examples green on 8.2.30/8.3.16/8.4.4/8.5.3, phpstan clean, cs-fixer clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
  - Accepting a feature suggestion from `next` runs `generate`, so it writes the feature instead of describing a spec at the project root
+ - Determined steps offer too: `next` asks before writing the steps or implementing the failing class, instead of printing advice and exiting
+ - A degenerate failure whose subject matches its example is named once, not twice
  - Auto-verify runs the whole suite when the change touches a feature or steps file
  - A verified change ends the turn; no fresh suggestion rides on its back
  - `refactor` runs its baseline with the installed phpspec, wherever it lives (vendor installs failed to launch it and blamed the specs)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,11 +13,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
  - The navigator offers directly; the chooser is the question, never "shall I offer?"
  - `refactor` shows the change and asks before applying; non-interactive runs keep auto-apply
  - `next` reads the journal: a polished, unchanged class is never re-suggested, and the model proposes the next scenario or feature instead
+ - A generated feature carries real model-written scenarios; the skeleton is only the fallback
+ - Every `next` suggestion offers to act (example, steps, implement, and refactor included), and the model can suggest steps and implement too
+ - A feature whose step bodies are pending reads as the working story: `next` steers to finishing it, never to a new feature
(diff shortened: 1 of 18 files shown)


MarcelloDuarte:
Scope grew: instead of fixing the one reported hole, I audited the whole next loop, then walked it end to end in a live todo project with a real provider, fixing every hole the walk surfaced. The second commit carries:

- Every suggestion type now offers to act: example and refactor had been left as hints (example runs generate, refactor runs the refactor command, which brings its own consent).
- The model's suggest_next vocabulary gains steps and implement, matching the deterministic side; before, the model was forced to say \"example\" when it meant \"implement\", and the accepted offer dead-ended on a spec that was already complete.
- Accepted example and implement offers forward the suggestion's reason into the generate instruction (file tokens stripped), so the one-shot writes the change the suggestion meant instead of guessing.
- Routing precedence: a named class with spec wording wins over a loose \"feature\" later in the sentence; the forwarded reason had been rerouting the whole instruction to a garbage feature file.
- write_feature now carries model-written Gherkin (the skeleton is only a fallback), so an accepted feature offer writes real scenarios instead of Given a starting context placeholders.
- A feature whose defined step bodies are still pending reads as its own suite state (pending, distinct from todo), and next steers deterministically to writing those bodies: against ready code they go green, against missing code they go red and red names the class. Before, the story could sprawl sideways into new features forever.
- When the working story is red, next feeds the model the story's actual files (feature, steps, last-touched spec and source), so the advice names the real gap instead of re-suggesting an example that already exists.
- Drive-by: a failing expectation on a list of objects crashed the failure-message formatter; it now renders Class#id per element.

Walked live: feature with real scenarios -> steps -> spec example matching the suggestion -> implement -> step bodies -> red naming each gap -> green -> refactor offer honouring the journal, plus the decline paths. One separate defect cluster found and left out deliberately (duplicate step phrases across steps files resolve first-wins with mismatched context properties, and generated features invent parallel vocabulary instead of reusing existing steps); filed to the board.

1697 examples green on 8.2.30/8.3.16/8.4.4/8.5.3, coverage 92%, phpstan and fixer clean.

MarcelloDuarte:
CI red was mine: the write_feature rework left the provider-failure path without the skeleton stand-in, and the generate.feature scenarios (real command, bogus api key) had been relying on the old offline scaffold. Fixed in the third commit: when a write-feature ask cannot reach the model (bad key, HTTP error), the derived skeleton stands in with the failure shown beside it; a missing ai section stays an honest config error. My local checklist had skipped the story suite; it ran this time on all four PHPs: 1698 examples plus 27 features / 1100 steps green everywhere, phpstan and fixer clean.

PR #1567, opened 2026-07-28
Error on duplicate step titles and generate one stub per title

MarcelloDuarte:
Dogfood bug: a scenario using the same step twice (\"Given I add a 'pending' task ... / And I add a 'completed' task ...\") got two identical pending stubs appended to its steps file. Deeper, duplicate step titles across files are a silent trap: runtime matching is global and keyword-blind, first definition wins, and a second definition against a different context property crashes confusingly far from the cause (seen live: \"Call to a member function add() on null\").

Two changes:

- StepRegistry rejects a second definition of a title it already holds, whatever the keyword, with a message naming both locations (file:line of each closure, via reflection): Step \"I filter the list\" is already defined at features/steps/a.steps.php:3; remove the duplicate at features/steps/b.steps.php:7. The run reports it red and exits 1; no stack trace.
- StepGenerator emits one definition per title: a step used twice in a scenario produces a single stub, and an existing definition is recognised in either quoting style.

Our own suite carried one duplicate (the file-should-contain assertion, defined in both assertions and extensions steps); the richer of the two bodies now lives once in assertions.steps.php. A new world assertion (file should contain exactly N times) locks the generator behaviour.

Verification: 1704 examples and 27 features / 220 scenarios / 1111 steps green on 8.2.30/8.3.16/8.4.4/8.5.3, coverage 92%, phpstan and cs-fixer clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+ - Two step definitions sharing a title now error at load, naming both locations; matching is keyword-blind, so the duplicate could only shadow silently
+ - A step used twice in a scenario generates one definition, not two
+
 ## [9.0.0-beta.15](https://github.com/phpspec/phpspec/compare/9.0.0-beta.14...9.0.0-beta.15)
 
 ### Added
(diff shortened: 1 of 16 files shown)


MarcelloDuarte:
Second commit closes the question of whether the AI can still write the duplicates the first commit made fatal: it cannot. A new StepVocabulary collaborator reads the step titles defined across all steps files (statically, both quote styles, every registrar), and every writer consults it. The one-shot propose_edit and all pair write tools reject steps content that redefines a title another file owns, with a message steering the model to reuse the existing step (same corrective-tool-result pattern as the ObjectBehavior rejection); content defining a title twice within itself is rejected the same way; and the deterministic scaffold now skips titles defined elsewhere instead of planting stubs that would error at the next load. Editing the file that owns a title still replaces it freely. 1714 examples and 1111 story steps green on all four PHPs, coverage 92.2, phpstan and fixer clean.

PR #1569, opened 2026-07-29
Failure output: sections, matcher-owned phrasing, and sentinel fixes

MarcelloDuarte:
Confirmed bug. An example holding several expectations records one MatchResult per expect() call, and passing ones carry the Detail::Nothing sentinel instead of a message. The pretty formatter rendered every MatchResult of a failing example, so each passing expectation before the failure printed a bogus line:

Failure: @PhpSpec\Result\Detail::Nothing@

The failing-example view now skips non-failures; only the failure is the story. The other formatters (dot, tap, junit, html, agent) were checked with the same multi-expectation repro and were already clean, since they read the example-level message.

Verification: 1722 examples and 27 features / 223 scenarios / 1129 steps green on 8.2.30/8.3.16/8.4.4/8.5.3, phpstan and cs-fixer clean.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
  - End-of-run detail is grouped into Failures, Errors, Warnings, Deprecations, and Skipped sections, and the dot formatter tells the same detailed story as pretty
- - Each matcher declares its own relation phrase, so a toContain failure reads as an aligned pair (expected / to be contained in), never a sentence repeating the values
+ - A failure reads as an aligned pair labeled by its matcher (expected / to contain), long multiline values capped head[...]tail with newlines escaped
 
 ### Fixed
  - The pretty formatter reports only the failed expectation; passing ones before it no longer print a bogus sentinel Failure line
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+ - The pretty formatter reports only the failed expectation; passing ones before it no longer print a bogus sentinel Failure line
+
 ### Added
  - `--format=agent` on every command: `next` emits its suggestion with the exact command that acts on it, `generate` applies and emits receipts, `describe`/`exemplify` accept it as the canonical spelling of `--agent` (now a deprecated alias)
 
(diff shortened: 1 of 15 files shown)


MarcelloDuarte:
Second commit for the second symptom in the same report: a toContain failure on captured multiline output embedded the whole blob in the Failure headline and then printed it again under got:, so one failure read as a repeated block. Expectation::formatMessage now clips string values in the headline to a first-line excerpt (short single-line values stay verbatim); the full value appears exactly once, labeled under expected/got, which every formatter carries structurally. The headline for the reported case is now: Failure: Expected "Analysing project...…" to contain "Would you like me to run it now?". 1724 examples and 1129 story steps green on all four PHPs, phpstan and fixer clean.

MarcelloDuarte:
Third commit turns the report feedback into the design discussed: end-of-run detail is grouped into Failures, Errors, Warnings, Deprecations, and Skipped sections (headers only when non-empty, severity order), rendered by one shared DetailSections collaborator that the dot formatter now uses too, so its bottom output is identical to pretty's instead of the short numbered list. And per the review steer, the relation phrasing is matcher-owned, not a switch in the formatter: each matcher declares its phrase beside its message template through the same trailing-option convention as __fake (toContain and toContainEqual say to be contained in, toHaveKey to be a key of, toStartWith and toEndWith the start and end variants, toMatch to be matched by), the result carries it, and the formatter renders whatever arrives with colons aligned and a not prefix when negated. A failure from a relation-declaring matcher shows the values exactly once, as the pair; matchers without a declared relation keep their message with the generic expected/got beneath. The old per-example error views and dot's private collectors are removed as dead code. 1730 examples and 1129 story steps green on all four PHPs, phpstan and fixer clean.

MarcelloDuarte:
Fourth commit undoes the relation plumbing per review: unnecessary complexity, since the label can be inferred from the matcher name the result already carries. toContain renders as the pair expected / to contain (camelCase split, not prefix when negated), which works for every named matcher with zero declarations; the __relation threading through Expectation, EventfulExpectation, MatchResult, and Failed is reverted entirely. Pair values from long or multiline strings are capped to the first thirty and last thirty characters around a [...] marker with newlines escaped as \n, so the reported example renders exactly as sketched: expected: "\n  Analysing project...\n\n  Ref[...]ke me to refactor that? [Y/n] " / to contain: "Would you like me to run it now?". Anonymous failures (custom matchers without a name) keep their message with the generic pair. 1731 examples and 1129 story steps green on all four PHPs, phpstan and fixer clean.

MarcelloDuarte:
Fifth commit: offered-code diffs render as context windows. Diff::format keeps each change with three unchanged lines either side (line numbers preserved) and elides any stretch between distant changes to a single ... line; a new file still shows in full, since every line is a change. One change in the shared renderer covers generate, refactor, the run generation offers, and pair mode. Verified live on a generate offer against the todo sandbox: the diff opens at the context window, not line 1. 1735 examples and 1129 story steps green on all four PHPs, phpstan and fixer clean.

